PR #11, opened 2015-01-15
files: Label OSTree .origin file as etc_t

cgwalters:
This will allow domains like rhsmcertd_t to write to it.  It is also
intended to be potentially editable configuration.


cgwalters:
(I didn't test this patch)

PR #17, opened 2015-05-10
Initial pdns policy module

rubenk:
Signed-off-by: Ruben Kerkhof <ruben@rubenkerkhof.com>

### DIFF
--- a/pdns.te
+++ b/pdns.te
@@ -1,0 +1,77 @@
+policy_module(pdns, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+type pdns_t;
+type pdns_exec_t;
+init_daemon_domain(pdns_t, pdns_exec_t)
+
+type pdns_unit_file_t;
+systemd_unit_file(pdns_unit_file_t)
+
+type pdns_conf_t;
+files_config_file(pdns_conf_t)
+
+type pdns_var_run_t;
+files_pid_file(pdns_var_run_t)
+
+type pdns_control_t;
+type pdns_control_exec_t;
+init_system_domain(pdns_control_t, pdns_control_exec_t)
+
+########################################
+#
+# pdns_t local policy
+#
+
+allow pdns_t self:capability { setuid setgid chown };
+allow pdns_t self:tcp_socket create_stream_socket_perms;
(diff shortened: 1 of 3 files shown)


doverride:
Looks like it is actually lacking the "pdns_can_network_connect_db" boolean
declaration?

You might not notice it since it is currently optional

rubenk:
Thanks, I've added it.

pdns.te
```diff
@@ -0,0 +1,84 @@
+policy_module(pdns, 1.0.1)
+
+########################################
+#
+# Declarations
+#
+
+## <desc>
+### <p>
```
wrabcak: ## is enough

rubenk: What do you mean?

pdns.te
```diff
@@ -0,0 +1,84 @@
...
+corenet_tcp_bind_dns_port(pdns_t)
+corenet_udp_bind_dns_port(pdns_t)
+
+files_pid_filetrans(pdns_t, pdns_var_run_t, { file sock_file })
+manage_files_pattern(pdns_t, pdns_var_run_t, pdns_var_run_t)
+manage_sock_files_pattern(pdns_t, pdns_var_run_t, pdns_var_run_t)
+
+auth_use_nsswitch(pdns_t)
+
+dev_read_urand(pdns_t)
```
wrabcak: This can be removed, this rule is included in auth_use_nsswitch.

wrabcak:
Hi, sorry for my delay. 
This looks really good! I have just 2 things.
First, If this policy just read from /etc/pdns, we don't need this label, in fedore we allow read etc_t files in default. 
Second thing, If you look on optional block, rules mysql_read_conf and mysql_stream_connect could be in tunable_policy block, due to this is needed when pdns is connecteing to db port? 

Thank you for your policy!

rubenk:
Hi Lukas,

I labelled the config file since it can contain passwords, for example for the mysql backend.

As for your second comment, if I moved the rules to inside the tunable block and the tunable is off, I think that would break if pdns connected locally to mysql via a unix socket.

wrabcak:
Understand. Merging. 
Thank you!

PR #19, opened 2015-06-19
rename several contrib modules according to their filenames

bachradsusi:
These are cosmetic changes needed for fluent transition to
libsemanage-2.4. Since the CIL is used, a module name is not stored in
module but a filename is used instead. With these changes, modules names before update to libsemanage-2.4 will be same as after update.


mgrepl:
Make sure you add required changes to selinux-policy.spec file.

PR #21, opened 2015-07-21
adding selinux policy files for sslh

hogarthj:
I'm the maintainer for sslh in Fedora. 

My goal is to confine the service as it is network facing.

The referred pull request contains the if/te/fc files for this.

This has been tested on F21 - the rc.d reference is to provide a common-ish base for EPEL usage and so that upstream policy covers those choosing to use a sysvinit script on other selinux capable distributions.

The general behaviour of the service as used by most people is to listen on tcp/443 and act as a mutliplexer between https and ssh traffic - although http, openvpn, jabber are also possible targets. 

There are booleans provided to at choice allow sslh to bind to or connect to any port if other services are required.


mgrepl:
It looks really good. It only needs some cosmetic changes. 

What does

$ rpm -qf /etc/sslh.cfg 
$ rpm -qf /etc/sysconfig/sslh

Also you should have sslh_admin() interface available.

wrabcak:
Agree with mgrepl. 
Attach info for mgrepl and then we could merge it.

hogarthj:
/etc/sslh.cfg is owned by sslh-1.17-1

/etc/sysconfig/sslh is not yet owned by the package but when this is sorted I'll be pushing 1.17-2 out which will correct this and own this file.

On the EPEL7 (and 5/6) side of things does RH port policy over or should I include the sslh.pp in the sslh package destined for EPEL?

I'll add the sslh_admin() interface requested and push the update to my fork of the branch.

hogarthj:
That's what happens when you push at the end of a long day ... 

Thanks for the feedback. I've added the gen_require so that that sslh_admin() builds correctly.

On the configuration side of things it seems sensible to be able to create the file if it doesn't exist.

I've been searching the fedora wiki and there's nothing about the expected contents of interface files etc for writing selinux modules for fedora. The closest is a draft from fedora core 5 days which is well out of date!

I've taken a look at the tor selinux policy and amended the sslh one based on that to add the systemd administration support. I was looking at the snmpd one before (hence the accidental paste typo).

sslh.te
```diff
@@ -0,0 +1,102 @@
...
+type sslh_unit_file_t;
+systemd_unit_file(sslh_unit_file_t)
+
+########################################
+#
+# sslh local policy
+#
+
+allow sslh_t sslh_config_t:file read_file_perms;
+allow sslh_t sslh_config_t:dir  search_dir_perms;
```
wrabcak: Here you can use read_files_pattern.

sslh.te
```diff
@@ -0,0 +1,102 @@
...
+
+auth_read_passwd(sslh_t)
+miscfiles_read_localization(sslh_t);
+
+allow sslh_t sslh_var_run_t:dir rw_dir_perms;
+allow sslh_t sslh_var_run_t:file { rw_file_perms create };
```
wrabcak: Here you can use manage_files_pattern.

wrabcak:
After this cosmetic details we could merge this policy.

hogarthj:
Thanks wrabcak ... I've updated with your suggestion for using the pattern. Compiled and installed the policy on my F22 system and ran the service and got the expected selinux behaviour from the sslh_t domain.

PR #23, opened 2015-07-29
Add new policy for "targetd" daemon. #1241415

vmojzis:


mgrepl:
Looks good. Not sure why does it manage lvm locks?

vmojzis:
Because targetd is a storage manager (https://github.com/agrover/targetd/) and it needs to be able to manage volumes in LVM volume group.

mgrepl:
So there are library calls which directly manage LVM? Or is it done using a lvm binary?

vmojzis:
Targetd manages logical volumes (without the use of lvm binary) and therefore needs access to files in /run/lock/lvm/.

mgrepl:
Ok. udev and lvm policies should be called with optional_policy block and also

 +/usr/lib/systemd/system/targetd.service	

should be fixed to 

  +/usr/lib/systemd/system/targetd\.service	

at least.

wrabcak:
or you can use: /usr/lib/systemd/system/targetd.*

wrabcak:
What is state of this issue? @mgrepl  Could you merge it? Then I make changes in our dist-git to make it working.
Thank you.

PR #24, opened 2015-08-05
Add interface "storage_read_fixed_disk" for reading fixed disk device nodes

vmojzis:
Required for "targetd" daemon.


mgrepl:
See my comment in the previous pull request.

PR #25, opened 2015-08-05
"Targetd" policy upgrade.

vmojzis:


mgrepl:
files_create_lock_dirs(targetd_t)

why is this needed? Does it create own lock?

mgrepl:
Also I would use

storage_raw_read_fixed_disk()

vmojzis:
In case /var/lock/lvm directory does not exist (targetd creates lock files there).

vmojzis:
Targetd does not need fixed_disk_raw_read which is described as extremely dangerous as it can bypass the SELinux protections for filesystem objects.

mgrepl:
Ok but in this case there should be called a new interface with filename transition rule for /var/lock/lvm.

Something like

lvm_filetrans_named_lock()

and should have

files_lock_filetrans($1, lvm_lock_t, dir, "lvm")

mgrepl:
Well but you want to allow to access a disk by 

storage_read_fixed_disk()

which is the same.

wrabcak:
In some cases, we forced to allow some "extremely dangerous" actions. It's storage daemon and due to usability it's needs this rule. The challenge of writing policies for SELinux is find balance between security/usability. You could allow it, but after that, you must be careful with domtrans to this domain.

PR #26, opened 2015-08-05
Two fixes to AVC denials found while using staff_u in a graphical environment

dawud:
Two fixes to AVC denials found while using `staff_u` in a graphical environment


mgrepl:
Yes, we have bugs for this issue where I suggested to test it with abrt_helper_exec_t because we already have

abrt_domtrans_helper(domain)

But sure we can think about a new domain.

PR #29, opened 2015-08-06
Targetd policy upgrade.

vmojzis:
- Allow raw fixed disk device read (required for normal daemon operation on server machines).


mgrepl:
Yeap, I agree with Dominick.

vmojzis:
Thanks.

PR #35, opened 2015-09-03
Update the policy for NetworkManager-libreswan-1.0.6

lkundrak:
Pull request #8, reworked.

Tested to work.


wrabcak:
Will be backported also to F23 and F22

PR #36, opened 2015-09-10
Sanlock policy update - New sub-domain for sanlk-reset daemon. #1255307 

vmojzis:


wrabcak:
Good work! I'm fine with this policy.

PR #40, opened 2015-09-29
lircd: Fix new drop-root-permissions denials.

leamas:
Two new denials  due to  lircd dropping root privileges code.

### DIFF
--- a/lircd.te
+++ b/lircd.te
@@ -28,6 +28,8 @@ allow lircd_t self:process signal;
 allow lircd_t self:fifo_file rw_fifo_file_perms;
 allow lircd_t self:tcp_socket { accept listen };
 allow lircd_t self:netlink_kobject_uevent_socket create_socket_perms;
+allow lircd_t passwd_file_t:file { read getattr open };
+allow lircd_t self:capability { setuid setgid dac_override };
 
 read_files_pattern(lircd_t, lircd_etc_t, lircd_etc_t)
 


lircd.te
```diff
@@ -28,6 +28,8 @@ allow lircd_t self:process signal;
 allow lircd_t self:fifo_file rw_fifo_file_perms;
 allow lircd_t self:tcp_socket { accept listen };
 allow lircd_t self:netlink_kobject_uevent_socket create_socket_perms;
+allow lircd_t passwd_file_t:file { read getattr open };
```
mgrepl: There should be 

auth_read_passwd(lircd_t) 

rule instead of this one. Basically we try to find proper interfaces if possible, especially if we try to access private types from another modules.

lircd.te
```diff
@@ -28,6 +28,8 @@ allow lircd_t self:process signal;
...
+allow lircd_t self:capability { setuid setgid dac_override };
```
mgrepl: Should be a part of the currnet rules for "capability" class.

Thank you.

leamas: I don't follow you here... Do you mean that  the last line isn't required, or should be rephrased (remember, this is selinux for dummies)

mgrepl:
In the lircd.te, we already have some rules for "capability" class.

allow lircd_t self:capability { chown kill sys_admin };

There  is no need to add a new line and you can just update this existing.

- allow lircd_t self:capability { chown kill sys_admin };
+allow lircd_t self:capability { dac_override setuid setgid chown kill sys_admin };

doverride:
+allow lircd_t passwd_file_t:file { read getattr open };

is not allowed. You can not reference external types directly in your module
use auth_read_passwd_file() instead (or whatever the macro is called)

leamas:
This one is closed, see #41

PR #41, opened 2015-09-29
lircd: Fix new drop-root-permissions denials (v2)

leamas:
New try: Add some new permission needed for new code dropping root privileges using setuid()

### DIFF
--- a/lircd.te
+++ b/lircd.te
@@ -40,7 +40,6 @@ dev_filetrans(lircd_t, lircd_var_run_t, sock_file)
 
 kernel_request_load_module(lircd_t)
 
-auth_read_passwd(lircd_t)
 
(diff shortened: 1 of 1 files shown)


lircd.te
```diff
@@ -40,6 +40,8 @@ dev_filetrans(lircd_t, lircd_var_run_t, sock_file)
 
 kernel_request_load_module(lircd_t)
 
+auth_read_passwd(lircd_t)
```
mgrepl: Changes are OK. We just need to follow StyleGuide. See 

https://github.com/TresysTechnology/refpolicy/wiki/StyleGuide

and _Local policy rules_ section. It means you need to call

         auth_read_passwd(lircd_t)

after 

        term_use_unallocated_ttys(lircd_t)

Thank you.

PR #42, opened 2015-10-01
Add interface to allow reading files in efivarfs

vmojzis:
Efivarfs contains Linux Kernel configuration options for UEFI systems (UEFI Runtime Variables)


mgrepl:
It is correct. But I would expect another interfaces

write,manage
read symlinks

and so on. See another interfaces in filesystem.if. 

Also for what will be this interface used?

wrabcak:
Hi!
Yes, more interfaces would be fine for future. 

This is needed by systemd-logind daemon which wants to read this fs. I'm waiting for next pull request by @vmojzis. I believe this is part of fix for broken gdm in F23/Rawhide.

PR #43, opened 2015-10-01
Allow systemd-logind read access to efivarfs

vmojzis:
BZ #1244973, #1267207 (partial solution)


mgrepl:
Do you have a bug number for this change?

PR #46, opened 2015-10-07
Allow NetworkManager_t and policykit_t read access to systemd-machined pid files. 

vmojzis:
BZ #1255305


mgrepl:
That looks fine. Should we it as a part of existing interface? It looks more domains could want this access.

PR #62, opened 2015-11-11
Allow systemd-hostnamed to communicate with dhcp via dbus. 

vmojzis:
BZ #1242583


mgrepl:
I would go with systemd_dbus_chat_hostnamed(dhcpc_t) rule instead.

PR #64, opened 2015-11-11
cockpit has grown content in /var/run directory

rhatdan:


cockpit.fc
```diff
@@ -8,3 +8,5 @@
 /usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
 
 /var/lib/cockpit(/.*)?      gen_context(system_u:object_r:cockpit_var_lib_t,s0)
+
+/var/run/cockpit-ws(/.*)?   gen_context(system_u:object_r:cockpit_var_run_t,s0)
```
stefwalter: On Fedora this is ```/run``` and ```/var/run``` is just a symlink.

mgrepl:
Do we need all these transitions? I would think a dir transition could be enough.

mgrepl:
https://bugzilla.redhat.com/show_bug.cgi?id=1279430

doverride:
> I would think a dir transition could be enough.

That is indeed what the accompanying file context specification seems to indicate

cockpit.fc
doverride: yes, but in the policy /var/run is an alias to /run, thus ever spec that applies to /var/run automatically applies to /run as well

stefwalter:
Thanks!

rhatdan:
I usually add all of the transitions rather then be selective.  Don't see a difference between cockpid creating a dir/file/lnk-file in /run being any different from a security point of view.  And this lessens the changes of cockpit eventually creating a file in /run in a different derectory causing problems.

@stefwalter SELinux has an equivalency between /run and /var/run

doverride:
Its not about security it is about efficiency and general common sense. It also makes the policy more readable if things make sense and if fc specs and type transition rules correspond

Your argument "And this lessens the changes of cockpit eventually creating a file in /run in a different derectory causing problems." does not hold with me. Since if cockpit were to for example maintain /riun/cockpit-ws.pid you would still need to add the proper file context spec for it.

rhatdan:
I disagree, over the years being too tight on security has led to an endless succession of bugs for little/no added security.  Actually I believe it has led to less security since we end up with people disabling SELinux because we decided to treat the ability to create a file in /var/run differently then creating a directory.

doverride:
@rhatdan I know your position on this and i respect that but would you agree that this introduces inconsistencies?

e.g. cockpit is allowed to create files in /run with a type transition but there is not accompanying file context spec. So we may end up with files labeled type cockpit_var_run_t that may get reset on a relabel to var_run_t because  the file context spec is missing.

This gives the user just as much of a bad experience as when there would be no type transition at all. It merely moving the problem.

It is inefficient, inconsistent, inaccurate. If you want to make the policy simpler then that is a nobel goal but make sure it is consistent atleast and that it actually solves issues.

PR #66, opened 2015-11-16
Allow list_dir_perms for user_home_type directories

jamielinux:
When starting tmux from inside a particular directory (eg,
`/home/user/projects`), the working directory should remain the same. If
tmux is unable to read the working directory, the working directory is
changed to `/` instead. Fix this by allowing `screen_domain` to read
`user_home_type` directories.


jamielinux:
Would be good to merge this to f23 as well.

jamielinux:
Hm, let me fix the commit title to something more descriptive..

PR #70, opened 2015-11-19
Allow antivirus_t to bind to all unreserved ports. 

vmojzis:
BZ #1248785


mgrepl:
Could you fix this commit message to be more descriptive? Thank you.

PR #84, opened 2015-12-17
Allow uspsmon reading form rand/urand. BZ #1282103, #1282104, #1282105

vmojzis:
Allow uspsmon to use kill within its domain. BZ #1276861


mgrepl:
Merged to rawhide and F23. Thank you.

PR #87, opened 2016-01-05
Allow reading of symlinks in the read_config

lzap:
Looks like in refpolicy this is correct.

### DIFF
--- a/puppet.if
+++ b/puppet.if
@@ -246,8 +246,9 @@ interface(`puppet_read_config',`
     ')
 
     files_search_etc($1)
-	list_dirs_pattern($1, puppet_etc_t, puppet_etc_t)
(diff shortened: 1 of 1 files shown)


puppet.if
```diff
@@ -246,8 +246,9 @@ interface(`puppet_read_config',`
     ')
 
     files_search_etc($1)
-	list_dirs_pattern($1, puppet_etc_t, puppet_etc_t)
+    list_dirs_pattern($1, puppet_etc_t, puppet_etc_t)
```
lzap: This was just a tab vs space.

wrabcak:
@lzap  Do you need back port it to F23 and F22?

lzap:
> @lzap  Do you need back port it to F23 and F22? 

Nope, thanks.

-- 
Later,
 Lukas #lzap Zapletal

PR #103, opened 2016-02-24
virt: add strict policy for virtlogd daemon

berrange:
The virtlogd daemon is currently given the same context as
libvirtd. This is essentially unrestricted host access which
is not at all desirable. The virtlogd daemon is a small single
purpose daemon whose only job is logging. It should have a
dedicated context which strictly controls what it is permitted
todo.

Signed-off-by: Daniel P. Berrange <berrange@redhat.com>


berrange:
New version of patch at https://github.com/fedora-selinux/selinux-policy/pull/104

PR #108, opened 2016-02-29
rkt needs these rules to run locked down containers

rhatdan:
We need to add container_image_t for read/execute content for containers.  I think we should break this code and all of the svirt_lxc_net_t code into a separate container.te module that we can ship independantly like we do with docker-selinux.te.  I would also like to get rid of docker-selinux.te and make it container.te.  But that might be difficult.

### DIFF
--- a/virt.fc
+++ b/virt.fc
@@ -65,6 +65,7 @@ HOME_DIR/\.local/share/libvirt/boot(/.*)?   gen_context(system_u:object_r:svirt_
 /var/lib/oz(/.*)?					gen_context(system_u:object_r:virt_var_lib_t,s0)
 /var/lib/oz/isos(/.*)?				gen_context(system_u:object_r:virt_content_t,s0)
 /var/lib/vdsm(/.*)?				gen_context(system_u:object_r:virt_content_t,s0)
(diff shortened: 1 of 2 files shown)


alban:
For the record, this patch is in both Fedora-24 and Fedora-Rawhide:

```
# # Fedora-Cloud-Base-24_Alpha-7.x86_64-eu-central-1-HVM-standard-0 (ami-748b6d1b)
$ rpm -qa|grep  selinux-policy-targeted
selinux-policy-targeted-3.13.1-179.fc24.noarch
$ rpm -q --changelog selinux-policy-targeted|grep 'Merge pull request #108'
- Merge pull request #108 from rhatdan/rkt
```

```
# # Fedora-Cloud-Base-Rawhide-20160321.0.x86_64-eu-central-1-HVM-standard-0 (ami-69967006)
$ rpm -qa|grep  selinux-policy-targeted
selinux-policy-targeted-3.13.1-179.fc25.noarch
$ rpm -q --changelog selinux-policy-targeted|grep 'Merge pull request #108'
- Merge pull request #108 from rhatdan/rkt
```

Thanks :)

PR #110, opened 2016-03-15
Allow reading resolvconf when it is a symlink.

rbarlow:
/etc/resolv.conf is a symlink on current Fedora Rawhide. This
commit will allow the sysnet_dns_name_resolve() policy to continue
to read the file as a symlink.

https://bugzilla.redhat.com/show_bug.cgi?id=1166071#c7

### DIFF
--- a/policy/modules/system/sysnetwork.if
+++ b/policy/modules/system/sysnetwork.if
@@ -442,6 +442,7 @@ interface(`sysnet_read_config',`
 
 	files_search_etc($1)
 	allow $1 net_conf_t:file read_file_perms;
+	allow $1 net_conf_t:lnk_file read_lnk_file_perms;
 
 	ifdef(`distro_debian',`
 		files_search_pids($1)


policy/modules/system/sysnetwork.if
```diff
@@ -442,6 +442,7 @@ interface(`sysnet_read_config',`
 
 	files_search_etc($1)
 	allow $1 net_conf_t:file read_file_perms;
+	allow $1 net_conf_t:lnk_file read_file_perms;
```
doverride: nit: it is "read_lnk_file_perms"

wrabcak:
Hi,
doverride is right. Could you fix it, please? 

Thank you.

rbarlow:
Hello @doverride and @wrabcak! I made the requested change and one more as well, but I now have two thoughts:

0) I see the same thing I added on line 457. I wonder why that didn't apply on my Rawhide box. Is 'distro_redhat' for RHEL and not Fedora? Also, if we add this on line 445, wouldn't it make sense to remove it from 457?

1) I added another line inside the distro_redhat area: ```allow $1 NetworkManager_var_run_t:dir list_dir_perms;```. Because /etc/resolv.conf now symlinks in strange ways for processes, it seems that my confined celery process in Pulp is trying to "search" on some NetworkManager stuff in /var/run. audit2allow showed me this:

```
$ sudo audit2allow -al


#============= celery_t ==============
allow celery_t NetworkManager_var_run_t:dir search;
allow celery_t net_conf_t:lnk_file read;
```

I assumed that 'search' would be provided by 'list_dir_perms'. I'm very new at SELinux development, so feel free to educate me because I'm sure I am making mistakes all over the place ☺

What do you think about this NetworkManager change?

doverride:
Search is part of list_dir_perms. Can you enclose the AVC denial for the "NetworkManager_var_run_t" requirement?

doverride:
But yes I have the same:

```
	(macro read_config_files ((type ARG1))
		(call file.read_config_lnk_files (ARG1))
		(allow ARG1 config_file read_file_perms)
		(optional dns_read_resolv_config_files_optional_sd_resolved
			(call sd_resolved.read_runtime_files (ARG1)))
		(optional dns_read_resolv_config_files_optional_nm
			(call nm.read_runtime_files (ARG1)))))
```

Basically resolv.conf can be a link to resolv.conf in /var/run/NetworkManager/, and also /var/run/systemd/resolve/.

doverride:
So whoever needs to read /etc/resolv.conf also needs to be able to read NetworkManager and systemd-resolved runtime files.

doverride:
The only difference in my implementation compared to Fedoras' is that I do not treat the /etc/resolv.conf symlink as "net_conf_t". In my policy it is just a "etc_t" symlink.

That is just a personal preference.

Anyhow I would probably create a seperate PR for the NetworkManager stuff, since your current implementation is at the least incomplete.

rbarlow:
Hello again @doverride! I agree that we can and should deal with NetworkManager separately, so I have removed that from this pull request. We can discuss it separately. But since you asked, here is the denial:

```
type=AVC msg=audit(1458145202.837:220): avc:  denied  { search } for  pid=796 comm="celery" name="NetworkManager" dev="tmpfs" ino=20694 scontext=system_u:system_r:celery_t:s0 tcontext=system_u:object_r:NetworkManager_var_run_t:s0 tclass=dir permissive=1
```

What do you think about my first question above? I'm still confused about why the distro_redhat section did not apply to my Fedora Rawhide system. I'm wondering if/why we need this pull request at all, since the code from it is duplicated in the distro_redhat section. It's funny that I say this since I wrote the PR, but maybe we should close it and figure out why that section doesn't apply to my box?

Perhaps the version of selinux-policy on my Rawhide box isn't up to date with the policy here?

doverride:
As far as I am concerned this PR is good now

rbarlow:
@doverride and I had a conversation on IRC, and we came to the conclusion that we should not merge this at this time since the ifdef distro_redhat block has the same line that I'm adding. We need to figure out why Rawhide isn't counting as distro_redhat instead of adding it to the general macro.

PR #111, opened 2016-03-16
Allow resolvconf links into NetworkManager.

rbarlow:
/etc/resolv.conf is a symlink into NetworkManager-labeled paths
on current Fedora Rawhide. This commit will allow the
sysnet_dns_name_resolve() policy to continue to read the file.

Here is an example AVC denial that I experienced when a confined
Celery process from the Pulp project tried to read the
/etc/resolv.conf symlink:

```
type=AVC msg=audit(1458145202.837:220): avc:  denied  { search } for  pid=796 comm="celery" name="NetworkManager" dev="tmpfs" ino=20694 scontext=system_u:system_r:celery_t:s0 tcontext=system_u:object_r:NetworkManager_var_run_t:s0 tclass=dir permissive=1
```

### DIFF
--- a/policy/modules/system/sysnetwork.if
+++ b/policy/modules/system/sysnetwork.if
@@ -454,6 +454,7 @@ interface(`sysnet_read_config',`
         init_search_pid_dirs($1)
 		allow $1 net_conf_t:dir list_dir_perms;
 		allow $1 net_conf_t:lnk_file read_lnk_file_perms;
+		allow $1 NetworkManager_var_run_t:dir list_dir_perms;
 		read_files_pattern($1, net_conf_t, net_conf_t)
 	')
 ')


rbarlow:
This pull request contains code that I originally proposed in https://github.com/fedora-selinux/selinux-policy/pull/111 but it was agreed that it made more sense to split it out into a separate pull request. You can read there for some discussion about the change.

policy/modules/system/sysnetwork.if
```diff
@@ -454,6 +454,7 @@ interface(`sysnet_read_config',`
         init_search_pid_dirs($1)
 		allow $1 net_conf_t:dir list_dir_perms;
 		allow $1 net_conf_t:lnk_file read_lnk_file_perms;
+		allow $1 NetworkManager_var_run_t:dir list_dir_perms;
```
doverride: "search_dir_perms" should be enough and is more appropriate.

I am really curious what that target of the /etc/resolv.conf symlink is labeled.
Can you do a `ls -al /etc/resolv.conf` , then do a `ls -alZ` on the target of the symlink and paste the results here please?

doverride: By the way, this is still wrong because type NetworkManager_var_run_t is external to the sysnetwork module. You would need to call a interface from the NetworkManager module.

In addition to that any reference in sysnetwork module to NetworkManager should be optional. Since one should be able to disable/remote the NetworkManager module (for example if one does not have NetworkManager and instead uses systemd-networkd)

doverride: I suspect we will eventually end up with something like this:

```
optional_policy(\`
    networkmanager_read_pid_files($1)
')
```

rbarlow:
I was able to do what I needed with auth_use_nsswitch(celery_t). I'm not sure how to explain the trouble I had, but I am happy to close this as not an issue. Thanks to @doverride for help!

PR #119, opened 2016-04-14
Allow modemmanager to talk to logind

lkundrak:
ModemManager uses logind's sleep inhibitor capability.

avc:  denied { send_msg } for msgtype=method_call
      in exe="/usr/bin/dbus-daemon" sauid=81
      hostname=? addr=? terminal=?


lkundrak:
This affects the Fedora 24 version of ModemManager as well.

PR #135, opened 2016-06-24
Container processes need to be able to listen on rawip sockets

rhatdan:
"Customer is reporting that any SCTP connections inside a container are failing,
due to an error in recvmsg(). "

https://bugzilla.redhat.com/show_bug.cgi?id=1349956


rhatdan:
Need this in the RHEL7.3 policy.

PR #138, opened 2016-07-20
Add support for user namespace

rhatdan:
Unconfined domains should have the same access to userns capabilties as
they do to real capabilities.

Userdomains that can execute chrome need sys_admin sys_chroot
sys_ptrace capabilities, they probbaly will need more,
but I don't know for sure.


wrabcak:
Make sense to allow. Adding to F24+. 
Thanks Dan.

PR #141, opened 2016-07-27
unconfined_exec_t typebound all exec_type

rhatdan:
Need these rules to allow unconfined domains to transiton to sandbox
domains over all entrypoints


rhatdan:
See email chain with Smalley.  I think eventually we will need to revisit these type bounds once checkpolicy gets merged.

rhatdan:
@wrabcak PTAL

rhatdan:
We probably need to talk about these.

PR #143, opened 2016-07-28
Allow sandbox domains to getattr on all filesytem file types

rhatdan:


wrabcak:
Dan, 
Something here is wrong, you sent over 250 comits, could you please check it and create new pull request?

Thank you.

rhatdan:
Yes I think we are all set on nsfs patches.

PR #147, opened 2016-08-22
Fixes for containers

rhatdan:


wrabcak:
Dan, 
Do we have any BZ for this issue? 
Thanks.

rhatdan:
No these are just things I am seeing from experiments with running systemd inside of a container,  I will create a bugzilla for these.
https://bugzilla.redhat.com/show_bug.cgi?id=1370427
https://bugzilla.redhat.com/show_bug.cgi?id=1370428

PR #151, opened 2016-08-30
Allow sandbox domains to use msg class

rhatdan:


rhatdan:
https://bugzilla.redhat.com/show_bug.cgi?id=1371512

rhatdan:
@wrabcak we need to get this into the RHEL7.3 release if at all possible.

PR #152, opened 2016-08-30
Make unconfined domains create resolv.conf with correct label

rhatdan:


rhatdan:
This might fix
https://bugzilla.redhat.com/show_bug.cgi?id=1370872

PR #155, opened 2016-09-10
Add boolean to allow sandbox domains to mount nfs

rhatdan:
Customers have requested management of NFS Shares from
a confined domain.

Also we should not allow a confined domain to load kernel modules.


rhatdan:
@wrabcak PTAL

rhatdan:
@mgrepl @wrabcak I Have cleaned this up.  PTAL

PR #156, opened 2016-09-25
Make init able to mount on tor_var_log_t and tor_var_lib_t

mscherer:
Should fix #1357395, modeled on fix for #1368621

### DIFF
--- a/tor.te
+++ b/tor.te
@@ -40,6 +40,8 @@ type tor_var_run_t;
 files_pid_file(tor_var_run_t)
 init_daemon_run_dir(tor_var_run_t, "tor")
 files_mountpoint(tor_var_run_t)
+files_mountpoint(tor_var_lib_t)
+files_mountpoint(tor_var_log_t)
 
 type tor_unit_file_t;
 systemd_unit_file(tor_unit_file_t)


doverride:
I think it is a bad idea to treat all ReadWriteDirectories as "mountpoint"

First of all this is systemd specific functionality. Second of all, and i am not sure about this since i have no easy access to Fedora policy, this will probably open up much more access to these ReadWriteDirectories than strictly needed. Last but not least it makes "mountpoint" less meaningful.

Instead i would create a new type attribute, something like "systemd_read_write_directoriy_type" for example. Then associate rules like for example: "allow init_t systemd_read_write_directory:dir mounton;" with the type attribute.

Then conditionally associate the systemd_read_write_directory_type type attribute with the various ReadWriteDirectories

mscherer:
The systemd unit come from upstream tor project [1], so even if that's a systemd specific feature, this would affect most of the Linux distributions already, since systemd is shipped on most modern distribution, and I do not see that tendency to change in a near future.

A similar modification was also already in the fedora package [2], so I just reused the same thing.

Looking at the policy, files_mountpoint seems to mark it as a mount point with a attribute, which is exactly what it should be since systemd mount a directory over it. I agree that this make "mountpoint" being less clear if anything can be a mount point, but the fact is that anything can be a mount point due to the way this feature is implemented (and in theory, anything can be a mount point on the FS, even if that's for 90% of the time silly, and I do not advocate marking anything as such). 

A quick non exhaustive look at the policy show that this give access to init_t to do the mount, and write on files/directories marked with that attribute which is what we want. It also give to process with mount_t (so most likely /bin/mount) the same access. While I agree that's maybe more access than we need (since that's systemd doing the mount, not the mount binary), I also do not see that as opening a gaping hole in the policy, especially since that's bug opened since 2 months, affecting enough Fedora users that I marked around 5 bugs as duplicate.

Your proposal of a specific type is sound, but I also fear that adding specialized type might make the policy more complicated (even if I also recognize this seems more logical). So I will defer to @wrabcak to see what he prefer.



[1] https://gitweb.torproject.org/tor.git/tree/contrib/dist/tor.service.in
[2] https://github.com/fedora-selinux/selinux-policy/commit/6b9712150b9d4bd5a1aa5133ef3590682d252014

doverride:
I look to this from a broader perspective. For example gentoo uses openrc, and so these rules do not apply to gentoo. I would not be surprised if upstream would reject this on these ground alone.

Also even though systemd might be here to stay in the near future, it is probably better to take into account the distant future. We did the same with upstart versus sysv init.

I know similar changes have been make to Fedora, but only now I decided to let my voice be heard on this matter.

Systemd has many more similar functionality that will eventually also need to be dealt with (properly). For example ConditionalFiles (e.g. ConditionPath/FileExists etc), but also PrivateDevices, PidFile, and socket activation.

If you're going to "simplify" all these then I suspect you may as well make systemd unconfined.

mscherer:
So far, no one did report any issue with others features, so we can deal with them later. 

I am in the process of writing a patch to add what you did propose, but that's slightly more complicated that this patch. Not in the principle, that's rather straight forward, more on deciding where the code should live, should it have conditional, etc, etc.

I fear that this will requires a round of discussion before being applied, and my priority is to have it fixed downstream for tor. And given the existing commit, I think we will have to deal with a refactoring in all case, hence I try to do it. I will ping you if I manage to get a patch.

doverride:
To write this in a sustainable way (and in the spirit of open source) one would have to look at how upstream reference policy has decided to deal with systemd.

Upstream reference policy decided to implement a build-time tunable "init_systemd". This allows reference policy to separate the systemd specific rules to a greater extent.

The problem is that Fedora has gone its own way, and has not consulted upstream on these matters. This makes things complicated, and more complicated with every change. ( i do understand the difficult position Fedora is in though, as it often does not have the luxury of deferring matters until some consensus is reached )

wrabcak:
Already fixed.

PR #167, opened 2016-10-14
Add transition rules for sandbox domains

rhatdan:


lsm5:
@mgrepl @wrabcak could you please build a new rpm with these changes for rawhide?

runcom:
> @mgrepl @wrabcak could you please build a new rpm with these changes for rawhide?

for context, we need this into rawhide to be able to build docker again (cause right now it's broken)

mgrepl:
Working on that. Thx.

runcom:
@mgrepl I saw http://koji.fedoraproject.org/koji/buildinfo?buildID=810031, we need that for rawhide as well (since there is where docker is broken).

runcom:
I believe we need this to build container-selinux in F24 as well, @lsm5 @rhatdan? We only build container-selinux for rawhide right now but we need to to that in F24 and F25 as well. @mgrepl we need this patch in an F24 build as well (since F25 build has already been done)

runcom:
sorry to bother again :angel: we will need this commit for F24 as well, otherwise I'm still facing the same issue as in Rawhide and f25 when compiling docker and we can't switch to the new `container-selinux` in F24

runcom:
@mgrepl can you also submit an update for http://koji.fedoraproject.org/koji/buildinfo?buildID=810031 selinux-policy-3.13.1-220.fc25 in F25 so I can build docker...

PR #168, opened 2016-10-27
Lots of new domains like OCID and RKT are user container processes

rhatdan:
This will allow new container daemons to automatically work with
container process types.


pmorie:
lol @ cccccccccc

rhatdan:
@wrabcak PTAL

PR #169, opened 2016-11-05
Add a boolean to enable the use of dac_override

mscherer:
This is needed for tor onion services, as seen on
https://bugzilla.redhat.com/show_bug.cgi?id=1392187


mscherer:
So, I just noticed now, but this PR was closed without much explanation, should I reopen it somewhere else, or was it rejected ?

mscherer:
So I reopened the PR on https://github.com/fedora-selinux/selinux-policy-contrib/pull/10

PR #170, opened 2016-11-10
Allow samba to manage lnk_files with samba_share_fusefs boolean

vinzent:
Right now samba only can manage files and directory but no
symlinks.

This allows smbd_t to also manage lnk_files aka symlinks.


vinzent:
@wrabcak any objections?

PR #172, opened 2016-11-22
Allow puppetagent_t to access timedated dbus

vinzent:
Running a puppet exec resource with timedatectl as
command, auditd logs this error:

```
type=USER_AVC
  msg=audit(11/21/2016 15:04:49.306:59375) :
  pid=741
  uid=dbus auid=unset ses=unset
  subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023
  msg='avc:  denied  { send_msg } for msgtype=method_return dest=:1.1421 spid=31613 tpid=31615
  scontext=system_u:system_r:systemd_timedated_t:s0
  tcontext=system_u:system_r:puppetagent_t:s0
  tclass=dbus
  exe=/usr/bin/dbus-daemon sauid=dbus
  hostname=? addr=? terminal=?'
```

Use the systemd_dbus_chat_timedated interface to allow
puppetagent_t the access.

### DIFF
--- a/puppet.te
+++ b/puppet.te
@@ -350,6 +350,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	systemd_dbus_chat_timedated(puppetagent_t)
 	systemd_dbus_chat_timedated(puppetmaster_t)
 ')
 


vinzent:
fixes a problem detected on RHEL 7.3 upgrade - see also RH support case 01744141

puppet.te
```diff
@@ -350,6 +350,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	systemd_dbus_chat_timedated(puppetagent_t)
 	systemd_dbus_chat_timedated(puppetmaster_t)
```
vinzent: access for puppetmaster_t was already set. but I don't know what business puppetmater_t has with timedated. Maybe sliped in errorrnous?

PR #173, opened 2016-11-28
Allow init system to read container files

rhatdan:


rhatdan:
Need this in order to fix https://github.com/openshift/openshift-ansible/issues/2811

rhatdan:
This is needed for system containers and potentially other use cases going forward.
Needs back port to F24,F25, RHEL7

PR #175, opened 2016-12-08
Allow domains that can list sandbox_dirs to look at container shares

rhatdan:
Overlay stores sandbox data in a type of container_share_t, we want
to allow domains to list these directories also.


giuseppe:
it solves the issue I've observed with journalctl.  Once I reload dbus.pp I am able to see the journal of a runc container with `journalctl -M $CONTAINER`.

rhatdan:
@wrabcak PTAL

PR #176, opened 2016-12-14
Add sssd_secrets labeling

rhatdan:
Also add named_filetrans interface to make sure all labels are correct.

### DIFF
--- a/sssd.fc
+++ b/sssd.fc
@@ -3,6 +3,7 @@
 /etc/sssd(/.*)?			gen_context(system_u:object_r:sssd_conf_t,s0)
 
 /usr/sbin/sssd		--	gen_context(system_u:object_r:sssd_exec_t,s0)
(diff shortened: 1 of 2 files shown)


rhatdan:
@wrabcak PTAL

lslebodn:
Would it be possible to backport this patch also to f24+ (including ded6d59f1a3a99cb080c186df74bd67601c9d835 ?
sssd-secrets is there as well.

BTW I tested latest rawhide branch and it works as expected.

lslebodn:
And if you have  a time could you explain a purpose of `sssd_filetrans_named_content`?
I would like to get more familiar with writing SELinux policy (and sssd is the best starting point for me)

rhatdan:
@lslebodn These interfaces are things we add for administrators.  Named filetrans allow us to specify if the admin (unconfined_t, sysadm_t) creates some content with a certain name, then it will get labeled correctly.

PR #179, opened 2017-01-04
Virt1

rhatdan:


rhatdan:
@wrabcak Could you take a look at these.  We need these for Fedora 25.

PR #180, opened 2017-01-10
Chrome requests sys_ptrace for confined users, but does not need it.

rhatdan:


rhatdan:
This creates a lot of AVC's on Rawhide.

PR #186, opened 2017-02-06
Add fs_associate_cgroupfs interface 

rhatdan:


rhatdan:
@stephensmalley PTAL

stephensmalley:
Hmmm...I guess this would work, but not clear on the tradeoffs of doing it this way versus the kernel patch.  Recommend asking Paul Moore, probably should be discussed on selinux list.  Do you envision defining type transitions for different domains on cgroup/cgroup2?

rhatdan:
@stephensmalley I thought this was the patch required for the kernel patch.  What is the syntax to make this work with the kernel patch?

I think having named file trans might be useful. but for now I expect container runtimes to do something like 

setcon("/sys/fs/cgroup/systemd", "system_u:object_r:container_file_t:s0:c1,c2")

rhatdan:
BTW I will send this to upstream as soon as we are sure of the syntax.

stephensmalley:
No, the kernel patch doesn't require any policy change; it just alters the kernel to support setting of labels from userspace while still defaulting to genfscon as the initial label.  You only need to change the policy if you want to support transitions on file creations by userspace.  If you change the policy to fs_use_trans, you don't need the kernel patch at all; fs_use_trans filesystems already support labeling from userspace.

stephensmalley:
Also, if you want to support genfscon per-file labeling, your kernel patch isn't sufficient; you would need to add cgroup/cgroup2 to the other whitelist as well.  I would like to be able to label the "release_agent" file within each cgroup mount and limit writes to it since it is similar to modprobe or other kernel usermodehelpers.

rhatdan:
@runcom PTAL

runcom:
ack, so we'll need to allow labeling per-file also in kernel, right?

stephensmalley:
Yes, unless Dan actually wants type transitions on cgroup.  However, I will note that we want to get rid of these filesystem type whitelists and replace them with something more general, see
https://github.com/SELinuxProject/selinux-kernel/issues/2

runcom:
@stephensmalley got this patch:
```diff
diff --git a/security/selinux/hooks.c b/security/selinux/hooks.c
index 2789f0a..d982ec8 100644
--- a/security/selinux/hooks.c
+++ b/security/selinux/hooks.c
@@ -808,6 +808,8 @@ static int selinux_set_mnt_opts(struct super_block *sb,
 
 	if (!strcmp(sb->s_type->name, "debugfs") ||
 	    !strcmp(sb->s_type->name, "sysfs") ||
+	    !strcmp(sb->s_type->name, "cgroup") ||
+	    !strcmp(sb->s_type->name, "cgroup2") ||
 	    !strcmp(sb->s_type->name, "pstore"))
 		sbsec->flags |= SE_SBGENFS;
```
@rhatdan could you ack on @stephensmalley comment?

rhatdan:
I can live without transition rules on cgroups. I think it is far more likely that we just need to handle setting xattrs.

rhatdan:
Changed pull request to just add an interface to allow file types to be associated with cgroup_t file systems.  
Need this to be able to assign container_file_t to cgroup_t.

PR #187, opened 2017-02-07
Allow container-selinux to handle all policy for container processes

rhatdan:


rhatdan:
@wrabcak We need this change to fix issues like:

https://bugzilla.redhat.com/show_bug.cgi?id=1420096

PR #192, opened 2017-02-28
Add interface to allow types to associate with cgroup filesystems

rhatdan:
Also fix some cut-and-paste errors on cgroup interfaces
Add interface to allow relabeling of cgroup files.


rhatdan:
@wrabcak PTAL

rhatdan:
@wrabcak PTAL

PR #198, opened 2017-07-05
Add permissions needed by systemd's machinectl shell/login

sigalrm:
This is some work on fixing the machinectl shell and machinectl login commands.

Relevant bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1219729
https://bugzilla.redhat.com/show_bug.cgi?id=1257990
https://bugzilla.redhat.com/show_bug.cgi?id=1416540

Tested on F26 beta, this allows machinectl shell and machinectl login to work for local root login. Have not tested with container logins.

See also pull request to selinux-policy-contrib


policy/support/obj_perm_sets.spt
```diff
@@ -267,7 +267,7 @@ define(`relabel_chr_file_perms',`{ getattr relabelfrom relabelto }')
 # Use (read and write) terminals
 #
 define(`rw_inherited_term_perms', `{ getattr lock read write append ioctl }')
-define(`rw_term_perms', `{ rw_inherited_term_perms open }')
+define(`rw_term_perms', `{ rw_inherited_term_perms open setattr }')
```
wrabcak: setattr should get own permission macro and also interfaces, like dir or files classes have.

wrabcak (changes requested):
PL looks fine, just I would change the last think related to changing rw_term_perms macro. 

Lukas.

wrabcak:
@sigalrm Any plans to improve the PR? 

Lukas.

sigalrm:
I’ll try to get it fixed before F28.

--Eamon W.

sigalrm:
Okay, here is try 2. Changes from previous patch:

- Rebased to current rawhide
- Dropped permission macro change, used existing term_setattr interface instead
- Allow systemd-machined to start getty services

Tested this patch on the the current F27 policy. Verified that all interfaces are still present in rawhide.

sigalrm:
Note: the [companion pull request](https://github.com/fedora-selinux/selinux-policy-contrib/pull/17) in fedora-selinux-contrib is unchanged.

ghost:
On Tue, Feb 20, 2018 at 05:46:21AM +0000, sigalrm wrote:
> Okay, here is try 2. Changes from previous patch:
> 
> - Rebased to current rawhide
> - Dropped permission macro change, used existing term_setattr interface instead
> - Allow systemd-machined to start getty services
> 
> Tested this patch on the the current F27 policy. Verified that all interfaces are still present in rawhide.

The pty should not be generic. It should be a user pty type

because when you do a `machinectl shell .host` you essentially end up with a user shell

any applications that you run from that shell will want to read write the pty

I suggest you make machined create the pty with a private "login pty type"
then make sure that users can "type_change" from the machined login pty type to the user pty type

Systemd will then relabel the machined pty to user pty when the user logs in

and then if you run any apps from that shell, they are able to read/write the pty

> 
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/fedora-selinux/selinux-policy/pull/198#issuecomment-366873185

-- 
Key fingerprint = 5F4D 3CDB D3F8 3652 FBD8  02D5 3B6C 5F1D 2C7B 6B02
https://sks-keyservers.net/pks/lookup?op=get&search=0x3B6C5F1D2C7B6B02
Dominick Grift

ghost:
The above applies to the `machinectl shell .host` part.
The gist is that if you run: ls -alZ `tty` from a `machinectl shell .host` session that the pty type should show up as a user pty type and not a generic pty type

As for `machinectl login container`: This is a different story. You need to target the container manager and container first. I suppose one could try to label systemd-nspawn "container_runtime_exec_t"

nspawn creates the container pty, and relabels it to the  "--selinux-apifs-context="
That means that "--selinux-apifs-context=" should always be specified in selinux enabled systems. I suppose one could try to set that to "system_u:object_r:container_file_t:s0" and set --selinux-context= to "container_t"

Here is a script i use to create simple nspawn containers. I might give some impression of the workarounds needed even to get this simple scenario going:

https://paste.fedoraproject.org/paste/iBWy913RKJmVNvlwroyAUg

ghost:
My last words about this:

When it comes to machinectl/machined there are two different scenarios:

1 .host
2 container

they have different properties. in scenario 1 your pty should end up with user pty type and in scenario 2 your pty should end up with a container pty type

The first scenario just create a local session, the second scenario is just machined using a container pty (a pty created by the container manager (nspawn))

ghost:
Not my last words:

I have looked at the bugzilla enclosed:

https://bugzilla.redhat.com/show_bug.cgi?id=1219729

this is a `machinectl shell .host` and `machinectl login .host` scenario

/dev/pts/ptmx is mislabeled. usually fedora uses /dev/ptmx instead of /dev/pts/ptmx

dbusd_system_t should be allowed to "rw_inherited_ptmx". This file descriptor gets passed to systemd through dbus by machined. The first problem here is though that /dev/pts/ptmx is mislabeled. My question in this bugzilla would first be: why is it using /dev/pts/ptmx instead of /dev/ptmx?

https://bugzilla.redhat.com/show_bug.cgi?id=1257990

this is a `machinectl shell .host` scenario

Here actually the proper thing happens. Instead of /dev/pts/ptmx, the /dev/ptmx fd is passed to systemd though dbus by machined.

And in indeed in this scenario systemd needs to behave like a login program because it is setting up a local session on behalf of an unconfined user using /etc/pam.d/login.

https://bugzilla.redhat.com/show_bug.cgi?id=1416540

This is a `machinectl login container` scenario

The nspawn container manager and container needs to be targeted and nspawn needs to be configured to use SELinux by default on fedora. (and that, and the implications of that should be documented)

In the bugzilla sfix bravely tried what i suggested in one of my replies above. However i suspect that he overlooked a few aspects (like configuring nspawn to use SELinux)

Like a said a few times already there is the scenario where one uses machinectl shell/login to create a local session (.host). In that scenario machined creates the (login) pty. Login users need to be allowed to type_change from this machined login pty to user pty. and in the case of `machinectl shell .host` init needs to be able to relabel machined login pty and user pty (to relabelfrom machined login pty to user pty when a user logs in and to relabelto machined login  pty from user pty when the user logs out) . In the case of `machinectl login .host` the above apllies to local_login_t instead of init (in the former scenario init acts as a login program in the latter scenario login acts as the login program) 

There is also the scenario where one uses machine shell/login to access a shell and to login to containers respectively. In that scenario the pty is created by the container manager. Thus it is important that the container manager it targeted first. nspawn creates the pty then labels it with the context it is instructed to associate with the container file system (it must be configured to do the right thing) Then machined just needs to be allowed to rw to container pty's and machinectl just gets passed the file desciptor.

Pretty complicated but interesting scenarios. The thing here is though that in the nspawn scenario no -one cares about selinux, The selinux people dont care about nspawn and nspawn doesnt care about selinux: and fedora users are stuck in the middle....

nspawn was never meant to be used by the main population. It was just something used to test systemd with internally.

sigalrm:
Reopening

sigalrm:
The pty is not generic, it is a user pty type as suggested by @doverride. The type "systemd_machined_devpts_t" is used as the private login type.

See also reopened pull request to selinux-policy-contrib

sigalrm:
Closing. Will reopen a clean PR after rebasing.

PR #199, opened 2017-07-07
Add a label for port udp 3780, used by conntrackd

mscherer:


mscherer:
This PR is a bit short, I should have added more context. I did deploy conntrackd on Fedora and found out that the daemon is not confined. So I started to write a policy, only to be blocked by the lack of label for the port being used. So this PR is the first step for that.

wrabcak:
Thanks for PR

PR #205, opened 2017-11-15
Fixes for containers

rhatdan:
We now need iptables to work in a user namespace.
We need content labeled correctly when run by unconfined domains.


rhatdan:
https://bugzilla.redhat.com/show_bug.cgi?id=1512783
Is one of the issues here.

wrabcak:
LGTM

PR #207, opened 2017-12-04
Allow systemd to manage container files

rhatdan:


rhatdan:
Lukas now that we can add labels to cgroups, systemd needs to be able to work with them.

wrabcak:
Yep, make sense.
Thanks for PR!

PR #209, opened 2018-01-23
kernel/files.fc: Label /usr/lib/sysimage as usr_t

cgwalters:
Moved from: https://github.com/fedora-selinux/selinux-policy-contrib/pull/43

This ensures that hardlinking works with `/usr/share/rpm` (once the
contrib patch to make it `usr_t` is merged too).

See https://bugzilla.redhat.com/show_bug.cgi?id=1526191
https://github.com/projectatomic/rpm-ostree/pull/959#issuecomment-325780234
https://github.com/projectatomic/rpm-ostree/pull/1142


wrabcak:
LGTM. 

Thanks.

PR #210, opened 2018-01-23
Consistently label usr_t for kernel/initrd in /usr

cgwalters:
A while ago the Fedora kernel packaging changed to add a copy in
`/usr/lib/modules`. Relatedly, on ostree-managed systems for historical reasons
we have the kernel in multiple locations. Unfortunately these locations have
different labels and hence can't be hardlinked. But there's no real reason to
have separate labels, so consolidate them.

For more information, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1526191

Note this will change the label of the kernel in `/usr` even
for "traditional" yum-managed systems, but IMO it's better
anyways as `vmlinuz` isn't really a module.

### DIFF
--- a/policy/modules/kernel/files.fc
+++ b/policy/modules/kernel/files.fc
@@ -219,6 +219,14 @@ ifdef(`distro_redhat',`
 /export(/.*)?			gen_context(system_u:object_r:usr_t,s0)
 /ostree(/.*)?           gen_context(system_u:object_r:usr_t,s0)
 
+# Note the Fedora kernel moved to /usr/lib/modules.  This also relates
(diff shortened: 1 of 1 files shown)


cgwalters:
Moved from https://github.com/fedora-selinux/selinux-policy-contrib/pull/44

wrabcak (approved):
LGTM

wrabcak:
@cgwalters , 

I prepared scratch build[1] with all your fixes, feel free to test it. I'll do builds to koji later this week.

[1] https://copr.fedorainfracloud.org/coprs/build/705237/

PR #218, opened 2018-07-16
Allow unconfined_service_t to transition to container_runtime_t

rhatdan:
I am seeing bugzillas where people are using lxc toold and having
problems with them transitioning properly.


wrabcak (approved):
LGTM

PR #220, opened 2018-07-26
Allow init_t to grab lock on /etc/.pwd.lock

filbranden:
This is required by the `DynamicUser=` feature of systemd 239, which grabs that lock to verify uniqueness of the users it's about to register through its NSS module.

This is maybe too coarse... But it's a bit hard to find a more fine grained solution that still works well with SELinux. See discussion in systemd/systemd#9583 for more details.

Tested: by installing this policy and starting a test service that has `DynamicUser=yes`, confirmed the user was properly created.

/cc @wrabcak @poettering


filbranden:
Already pushed in 5a335fe9177b3461ad5b4c7e06e2c3ab83e53e14, so closing.

PR #224, opened 2018-08-30
Allow to remove duplicates related to defaul nsswitch modules

lslebodn:
Most of explanation is in commit message itself.

And small simplification is also done in netlabel module. But later I tried to refactoring in selinux-policy-contrib and I realized that the macro `auth_use_nsswitch` needn't be the best place maybe `auth_read_passwd` is a better one for allowing *sss* and *systemd* nsswitch modules which are by default in `/etc/nsswitch` on fedora for some time.

@wrabcak please let me know which of these `auth*` related macros make more sense.

BTW `libnss_systemd.so` can be used just for `passwd/group` but `libnss_sss.so` can be also used for `services/netgroup/sudo/autofs`

But systemd-libs also contains `libnss_myhostname.so.2` and  `libnss_resolve.so.2` which can be used for `hosts` and it seems that they also communicate ower dbus

```
sh$ strings /usr/lib64/libnss_systemd.so.2  | grep system_bus_socket
unix:path=/run/dbus/system_bus_socket
,argv5=--machine/var/run/dbus/system_bus_socket

sh$ strings /usr/lib64/libnss_mymachines.so.2  | grep system_bus_socket
unix:path=/run/dbus/system_bus_socket
,argv5=--machine/var/run/dbus/system_bus_socket

sh$ strings /usr/lib64/libnss_resolve.so.2  | grep system_bus_socket
unix:path=/run/dbus/system_bus_socket
,argv5=--machine/var/run/dbus/system_bus_socket
```

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -529,7 +529,6 @@ optional_policy(`
 	init_sigchld(domain)
 	init_signull(domain)
 	init_read_machineid(domain)
-    init_dbus_chat(domain)
 ')
 
 ifdef(`distro_redhat',`
--- a/policy/modules/system/authlogin.if
+++ b/policy/modules/system/authlogin.if
@@ -2205,6 +2205,7 @@ interface(`auth_dontaudit_getattr_passwd',`
 ########################################
 ## <summary>
 ##	Read the passwd passwords file (/etc/passwd)
+##	Allow to use sss nsswitch module for passwd and group.
 ##	Allow to use systemd nsswitch module for passwd and group
 ##	which is used for dynamic users.
 ## </summary>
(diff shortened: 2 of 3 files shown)


lslebodn:
I will continue with simplification of modules for rawhide in selinuc-policy-contrib after decision about `auth_use_nsswitch` vs `auth_read_passwd`.
If we decide to go with `auth_read_passwd` then we can drop last commit. It's already allowed :-)

And I do not think that commit "Limit communication with init_t over dbus" is a regression because such change hasn't been pushed to f28 yet :-)

wrabcak:
I would vote for auth_read_passwd interface. 

Should I leave this PR opened and when you finish, I'll merge it right?

lslebodn:
Updated. Thank you for review.

wrabcak:
LGTM. 

Thanks!

PR #225, opened 2018-09-04
Add .gitignore file

sigalrm:
When following the Compiling instructions on the wiki, there are untracked files present in the working directory after compiling. This is simply an attempt to ignore them.


wrabcak:
Good catch! 

Thanks for help.

PR #227, opened 2018-09-09
Allow getattr as part of files_mounton_kernel_symbol_table.

filbranden:
Allow `getattr` as part of `files_mounton_kernel_symbol_table`.

This makes it possible to resolve symbolic links on the target `/proc/kallsyms` before mounting over it.

systemd tries to follow symbolic links before mounting over proc files, when `ProtectKernelTunables=yes` is set.

This resolves issue systemd/systemd#10032.

Tested on Rawhide with systemd from trunk, by building new selinux-policy RPMs including this change and reproducing the issue with restart of systemd-hostnamed, which fails without this change.

cc @wrabcak -> Please take a look!

Thanks @sourcejedi for helping narrow this down to an SELinux `dontaudit` rule!

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -162,7 +162,7 @@ files_search_base_file_types(domain)
 files_read_inherited_tmp_files(domain)
 files_append_inherited_tmp_files(domain)
 files_read_all_base_ro_files(domain)
-files_dontaduit_getattr_kernel_symbol_table(domain)
+files_dontaudit_getattr_kernel_symbol_table(domain)
 
 # All executables should be able to search the directory they are in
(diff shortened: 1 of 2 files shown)


wrabcak:
LGTM. 
Thanks!

filbranden:
@wrabcak I've been trying to confirm this works by getting a selinux-policy from the nightly COPR, but I don't see any builds after Sept 10... So I'm wondering if my PR somehow broke the build? Or if there's a different issue? Anyways, let me know if there's a known issue and if I can help with addressing it... Thanks!

wrabcak:
Hi, 

No I just didn't create new builds yet. Will do asap. 

Thanks,
Lukas.

PR #229, opened 2018-09-27
Add the port used by nsca (Nagios Service Check Acceptor)

mscherer:


mscherer:
Same as #199, installed that on Centos 7, not confined, gonna write a policy. And one for the client too.

wrabcak:
LGTM. 
Thanks.

PR #230, opened 2018-10-29
kernel/files.fc: Label /run/motd.d(/.*)? as etc_t

rfairley:
This is to allow sshd to read a motd file placed in this
directory.

**Reason**

The reason for this is to let packages that generate a motd
at runtime (e.g. cockpit-ws, greenboot) place the file
in `/run/motd.d/`, rather than in a package-specific
directory in `/run/` containing the motd. The files in
`/run/motd.d/` may then be symlinked from `/etc/motd.d/`
to have the message display on SSH login.

Will also help for
https://github.com/rfairley/fedora-coreos-login-messages
(see https://github.com/coreos/fedora-coreos-tracker/issues/36).

Also, if https://github.com/linux-pam/linux-pam/pull/69
lands, this will allow files to be placed directly into
`/run/motd.d/` (not symlinked) and be displayed on SSH
login. Then, `/run/motd` can be labelled.

**Testing/verification done**

On Fedora 28 Atomic, I have confirmed
```
semanage fcontext -a -t etc_t "/var/run/motd.d(/.*)?"
restorecon -Rv /var/run/motd.d
```
gives the desired behaviour.

**Alternatives**

An alternative label other than `etc_t` that would work
is `pam_var_run_t` - if this would be preferable to `etc_t`.

The other option is to give sshd permission to read
`var_run_t`, but I am not sure how acceptable this is:

```
#============= sshd_t ==============
allow sshd_t var_run_t:file read;
```

### DIFF
--- a/policy/modules/kernel/files.fc
+++ b/policy/modules/kernel/files.fc
@@ -299,6 +299,7 @@ ifndef(`distro_redhat',`
 /var/run/lock/.*		<<none>>
 
 /run/cockpit/motd	--	gen_context(system_u:object_r:etc_t,s0)
+/run/motd.d(/.*)?		gen_context(system_u:object_r:etc_t,s0)
 
 /var/spool(/.*)?		gen_context(system_u:object_r:var_spool_t,s0)
 /var/spool/postfix/etc(/.*)?	gen_context(system_u:object_r:etc_t,s0)


rfairley:
cc @dustymabe @LorbusChris @lucab

thanks :)

policy/modules/kernel/files.fc
```diff
@@ -297,6 +297,7 @@ ifndef(`distro_redhat',`
 /var/run/.*			gen_context(system_u:object_r:var_run_t,s0)
 /var/run/.*\.*pid		<<none>>
 /var/run/lock/.*		<<none>>
+/var/run/motd.d(/.*)?		gen_context(system_u:object_r:etc_t,s0)
```
dustymabe: i'm wondering if the spacing should be changed to align `gen_content` with the `<<none>>` on the line above

rfairley: On GitHub on my end (in Firefox), the spacing of `gen_context` looks to be in line with the `<<none>>` on line 299. Maybe it using tabs is causing it to display differently?

dustymabe: ahh yes. the dreaded varying display width for a tab :). I pulled down the patch and applied it locally and changed my tabwidth to 8 and it looks good.

ignore me @rfairley :smile:

rfairley: Thanks and no problem :) had that hit me on linux-pam as well which looked good only with a width of 8

lucab: It looks like this PR/commit is freely mixing `/var/run` and `/run`. Are the two *always* guaranteed to be symlinked? Given the original usecase, wouldn't it be better to use `/run` here?

rfairley: Yes, I just had a look at the FHS again. Makes sense to use `/run` as it is the newer one, and isn't a symlink - `/var/run` isn't guaranteed to be symlinked, I guess if it is an older system. Changing this, thanks!

rfairley:
`/run` makes sense for our use case, but when changing it (on a Fedora 28 Atomic Host system with targeted policy) with `semanage` the following error is given:

```
# semanage fcontext -a -t etc_t "/run/motd.d(/.*)?"
ValueError: File spec /run/motd.d(/.*)? conflicts with equivalency rule '/run /var/run'; Try adding '/var/run/motd.d(/.*)?' instead
```

But `/run` configured in `kernel/files.fc` seems valid, given there are other lines specifying `/run` there. Could a maintainer confirm that specifying a line beginning with `/run` is valid? Is there a better way than `semanage` to verify that the context gets generated correctly? Thanks.

**Edit** with `chcon` instead of `semanage` the context gets added successfully and behaves as expected, which verifies the context works.

wrabcak:
LGTM. 

Will backport also to F29 and F28.

rfairley:
Thanks @wrabcak !

dustymabe:
woot! thanks @wrabcak

PR #235, opened 2018-12-14
Add new type systemd_timedated_var_lib_t for /var/lib/systemd/timesync/

filbranden:
Add new type `systemd_timedated_var_lib_t` for `/var/lib/systemd/timesync/`.

Set up rules so that systemd-timesyncd can manage the files in that directory.

Before this commit, I was having trouble with `systemd-timesyncd` on Fedora Rawhide, raising AVCs trying to access/modify files tagged with `init_var_lib_t`.

The `lnk` rules are needed because in many cases `/var/lib/systemd/timesync` is a symlink to `../private/systemd/timesync` (left over from the `DynamicUser=yes` days...) In any case, I think they're still sane enough.

cc @wrabcak @keszybz


wrabcak:
LGTM.

wrabcak:
Thanks.

PR #246, opened 2019-01-25
Allow systemd-journald to receive messages including a memfd

filbranden:
I believe these are related to dbus-broker from the bus1 project, which is now able to pass memfd's together with its DBus brokered messages, at least I think this is what changed in my environment around the time I started getting these. (It's possible they're related to systemd-journald code changes, but although I haven't checked, I think that's less likely.)

Without the rule for memfd from this commit, journald logs the following to the console:

```
systemd-journald[621]: Failed to map memfd, ignoring: Permission denied
```

And an AVC such as the following is generated:

```
avc: denied { map } for pid=621 comm="systemd-journal" path="/memfd:sd-test-journal-se (deleted)" dev="tmpfs" ino=51156 \
     scontext=system_u:system_r:syslogd_t:s0 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=0
```

This commit adds a specific rule to allow this access to memfd.

**NOTE**: audit2allow suggests setting a boolean would be possible here:

```
#!!!! This avc can be allowed using the boolean 'domain_can_mmap_files'
allow syslogd_t user_tmp_t:file map;
```

But I think adding the specific rule for journald makes sense here. I'm not sure whether `user_tmp_t` is too broad or not, I'll leave that for you to decide.

/cc @wrabcak

### DIFF
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -457,6 +457,8 @@ allow syslogd_t self:netlink_audit_socket { r_netlink_socket_perms nlmsg_write }
 allow syslogd_t self:netlink_generic_socket create_socket_perms;
 allow syslogd_t syslog_conf_t:file read_file_perms;
 allow syslogd_t syslog_conf_t:dir list_dir_perms;
+# receive messages including a memfd
+allow syslogd_t user_tmp_t:file map;
 
 # Create and bind to /dev/log or /var/run/log.
 allow syslogd_t devlog_t:sock_file manage_sock_file_perms;


WOnder93:
I think this is actually a kernel bug. The kernel shouldn't check for access to internal memfd files, just for the common shm permissions. See e. g. https://github.com/torvalds/linux/commit/e1832f2923ec92d0e590e496c8890675457f8568, which fixed this for internal shmem files.

I'll have to take a closer look. (But in the meantime, I'm afraid this workaround in the policy will be needed...)

WOnder93:
Related to the above shmem patch: https://bugzilla.redhat.com/show_bug.cgi?id=1613056

WOnder93:
Actually, after further investigation I don't think it is really a kernel bug. A memfd is really an anonymous tmpfs file and there is probably no easy/reasonable way to treat/label it in a different way. @wrabcak, the PR looks OK from my side.

wrabcak:
Thanks for help Ondrej.

Merging

PR #249, opened 2019-02-08
Always label /home symlinks as home_root_t

jlebon:
On OSTree systems, `/home` is a symlink to `/var/home`. With the latest
release of RPM-OSTree, `HOME_ROOT` is now `/var/home`. This leaves the
`/home` symlink unmapped and defaulting to `default_t`. This in turn is
preventing systemd from mounting filesystems pointing at `/home` in
`/etc/fstab`.

Let's just unconditionally make sure the link is labeled correctly. On
systems with `HOME_ROOT == /home`, this is just a duplicate rule, which
is harmless.

For more context, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1669982
https://src.fedoraproject.org/rpms/selinux-policy/pull-request/14

### DIFF
--- a/policy/modules/kernel/files.fc
+++ b/policy/modules/kernel/files.fc
@@ -108,6 +108,12 @@ HOME_ROOT/\.journal		<<none>>
 HOME_ROOT/lost\+found	-d	gen_context(system_u:object_r:lost_found_t,mls_systemhigh)
 HOME_ROOT/lost\+found/.*	<<none>>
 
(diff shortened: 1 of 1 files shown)


cgwalters:
This LGTM.  

(We may also need to circle back and do this same thing for `/root` but eh, let's not get tangled up in that right now)

wrabcak:
LGTM.

PR #251, opened 2019-03-13
Grant permissions for onloadfs files of all classes

ghost:
The existing implementation of `fs_rw_onload_sockets()` grants permissions only to the `sock_file` class.  In recent versions of Onload (201606-u1 and later), onloadfs sockets do not set `S_IFSOCK` and so are classed as `file`, not `sock_file`.  Furthermore, onloadfs pipes are `fifo_file`.  This patch grants permissions to all of the relevant classes, including `sock_file` for backwards-compatibility.

### DIFF
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -7014,8 +7014,10 @@ interface(`fs_rw_onload_sockets',`
 		type onload_fs_t;
 	')
 
+	rw_files_pattern($1, onload_fs_t, onload_fs_t)
+	rw_fifo_files_pattern($1, onload_fs_t, onload_fs_t)
 	rw_sock_files_pattern($1, onload_fs_t, onload_fs_t)
-    allow $1 onload_fs_t:sock_file ioctl;
+	allow $1 onload_fs_t:sock_file ioctl;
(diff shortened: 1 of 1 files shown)


policy/modules/kernel/filesystem.if
```diff
@@ -7014,8 +7014,10 @@ interface(`fs_rw_onload_sockets',`
 		type onload_fs_t;
 	')
 
+	allow $1 onload_fs_t:file {ioctl read write getattr};
```
wrabcak: Could you please change line 7017 to:
rw_files_pattern($1, onload_fs_t, onload_fs_t)

wrabcak (changes requested):
Thanks for PR, could you please see my comments?

policy/modules/kernel/filesystem.if
ghost: Thanks for the quick response.  I've made the change in 10a61df.  I'll make an equivalent change in https://github.com/containers/container-selinux/pull/66.

wrabcak (approved):
LGTM.

wrabcak:
Merged, 
Thanks!

PR #253, opened 2019-03-27
Label fontconfig cache and config files and directories BZ(1659905)

zpytela:


ghost (commented):
There are conflicts between your rules and your specs.

Example:
Your `gnome_config_filetrans()` rule implies that there is a `dir` with name `fontconfig` in `~/.config`, however your accompanying file context specification only addresses files (the `--` covers only files)

This means that if the user decides to `restorecon -RvF ~` , he will end up with inconsistencies.

zpytela:
@doverride Thank you for pointing this out. I also managed to find some other issues, so the patch is not ready for review atm.

ghost:
Still inconsistencies (i think):

HOME_DIR/\.config/fontconfig/conf\.d(/.*)?

should probably be:

HOME_DIR/\.config/fontconfig(/.*)?

to make it consistent with the filetrans interface call.

PROTIP: test these things out by using restorecon -RvF to reset the locations in both the file trans as well as the corresponding file spec.

Also not sure if you called the filetrans interface for confined users as well but you should if you did'nt

zpytela:
I accepted the comments and created a new commit which is now ready for reviewing again.

wrabcak:
LGTM.

PR #255, opened 2019-04-01
Add permissions needed by systemd's machinectl shell/login (Attempt 2)

sigalrm:
This is some work on fixing the machinectl shell and machinectl login commands.

Relevant bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1219729
https://bugzilla.redhat.com/show_bug.cgi?id=1257990
https://bugzilla.redhat.com/show_bug.cgi?id=1416540

Tested on F29, this allows machinectl shell and machinectl login to work for local root login. Have not tested with container logins.

See also pull request to selinux-policy
See also pull request #198 (previous attempt)


policy/modules/system/getty.te
```diff
@@ -98,6 +98,8 @@ term_setattr_console(getty_t)
 term_use_console(getty_t)
 term_use_usb_ttys(getty_t)
 term_setattr_usb_ttys(getty_t)
+term_use_all_ptys(getty_t)
+term_setattr_all_ptys(getty_t)
```
ghost, line 102: Makes sense yes, and I would be okay with this but i suspect that you might be able to narrow this down to user pty and machined pty.

policy/modules/system/init.te
```diff
@@ -318,14 +319,16 @@ selinux_compute_access_vector(init_t)
...
 selinux_unmount_fs(init_t)
 
 term_create_pty_dir(init_t)
 term_use_unallocated_ttys(init_t)
 term_use_console(init_t)
 term_use_all_inherited_terms(init_t)
-term_use_generic_ptys(init_t)
 term_use_usb_ttys(init_t)
+term_use_all_ptys(init_t)
+term_setattr_all_ptys(init_t)
```
ghost, line 331: Same here. I suppose this is fine but would be better to narrow this  down to user and macined pty.

You might be wondering why: there can be other pty's as well, for example qemu serial console pty, and neither systemd nor getty have business with those AFAIK

policy/modules/system/locallogin.te
```diff
@@ -124,9 +124,10 @@ term_relabel_unallocated_ttys(local_login_t)
 term_relabel_all_ttys(local_login_t)
 term_setattr_all_ttys(local_login_t)
 term_setattr_unallocated_ttys(local_login_t)
-term_relabel_all_ptys(local_login_t)
-term_setattr_generic_ptys(local_login_t)
 term_use_usb_ttys(local_login_t)
+term_use_all_ptys(local_login_t)
+term_relabel_all_ptys(local_login_t)
+term_setattr_all_ptys(local_login_t)
```
ghost, line 130: Yes, I would say same as above but i don't know why it was already allowed access to all pty's

So these and the ones for getty and init as fine with me this way.

ghost (commented):
Yes this looks nice.

ghost:
> Yes this looks nice.

I suspect that there are a couple of minor things missing related to confined users. For example machined should probably be a domain_interactive_fd subject type since confined applications will probably want to be able to use machined interited pty.

But those are minor things that can always be added later,when someone wants to make it work for confined users.

As Far as i am concerned, this is good to go. thanks

wrabcak:
@doverride, Thanks a lot for review! 

Thanks,
Lukas.

PR #257, opened 2019-04-09
Allow systemd-logind read and write user domain terminals BZ(1696852)

zpytela:
Allow systemd-logind read and write all user domain terminals
instead of just user ttys

Signed-off-by: Zdenek Pytela <zpytela@redhat.com>


wrabcak:
LGTM. 
Thanks.

PR #258, opened 2019-04-11
Fix kernel rw kernel sysctl f30

AdamWill:
This fixes the mistake that led to https://bugzilla.redhat.com/show_bug.cgi?id=1698575 : on rawhide branch this interface was added with the same name as an existing interface and then renamed, but on f30 branch, the interface was added with the same name as an existing interface and then the original interface was removed.

We revert the erroneous removal commit, and cherry-pick the rename from rawhide branch instead.


wrabcak:
LGTM. 

Thanks.

PR #261, opened 2019-04-26
Allow sysadm_t rtkit_scheduled

zpytela:


ghost:
Interesting report and solution. The reporter states two interesting bits of information:

1. SELinux prevents sysadm_u from D-bus talking to rtkit-daemon
2. The scenario works fine for other confined users like user_u or staff_u

So one would probaby start looking at the differences between "sysadm_u" and "user_u or_staff_u" in this regard. This leads to: https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/userdomain.if#L1352

And so one could argue that this PR addresses the issue in that regard.

However,  if you look closer to the bug report then you see that the reporter is just using gdbus to system dbus chat with rtkit. But then he makes the comparison with user_u and staff_u.

So what does he expect?

1. Does he expect you to find and solve the inconsistency between sysadm_u  and user_u or staff_u?
If he does then this is the proper fix.
2. Does he expect you to only allow sysadm_u to system dbus  chat with rtkit, not minding any other consistencies related to rtkit, sysadm_u and user_u or staff_u?
This fix does allow sysadm_u to system dbus chat with rtkit, but it allows more than just that. One could argue that this also fixes an inconsistency with user_u or staff_u, the guestion is whether that is desirable.
3. Does he expect to be able to dbus chat as sysadm_u with any entity using the gdbus command?
This fix does not address that

I might just be reading too much into this. Technically this solution addresses the bugreport, and if the reported later claims that this solution is not what he expected then you can remind him of what he litterally asked, and claim that you addressed what he asked for. Namely:

1. he can now dbus chat with rtkit via gdbus
2. the inconsistency between sysadm_u and user_u or staff_u with regard to this is addressed

But he still wont be able to use gdbus to freely dbus chat with all entities. It is just not clear if that is what he expects (although it certainly looks like it)

zpytela:
@doverride, thank you for the review and deep dive. I also looked at the issue from the other side and as a result I suggest changing the call of `userdom_login_user_template` to `userdom_restricted_xwindows_user_template`, which is its superset, so that all admin-type SELinux users have the same X window privileges as user-type ones. It triggered some duplicate permissions and interface calls. Now the policy package builds, installs, and grants the permission. The patch looks intrusive though, and additional change in dbus.if/dbus_role_template is required.

For anybody interested, this is the templates tracepath from `policy/modules/roles/{staff.te,sysadm.te,unprivuser.te}` and `policy/modules/system/userdomain.if`:
```
userdom_unpriv_user_template(staff)
    userdom_restricted_xwindows_user_template($1)
        userdom_restricted_user_template($1)
            userdom_login_user_template($1)
                userdom_base_user_template($1)
    userdom_common_user_template($1)

userdom_admin_user_template(sysadm)
    userdom_login_user_template($1)
        userdom_base_user_template($1)
    userdom_common_user_template($1)
```
@wrabcak, awaiting review

wrabcak:
Hi All, 
Thanks Guys for fixing this. 

Zdenek,
Could you please resolve conflicts? 

Thanks,
Lukas.

zpytela:
Conflicts resolved, making a scratchbuild to check the changes effect.

wrabcak:
Hi @zpytela, 
Do you have scratch build ? 

Thanks,
Lukas.

wrabcak:
Hi @zpytela , 

What is the status of the PR? 

THanks,
Lukas.

ghost:
I think you already address the issue here:
https://github.com/fedora-selinux/selinux-policy/commit/5a6f251d11084a60e1dcb7b63a5001629592e035

wrabcak:
@doverride ++

Closing the ticket.

PR #263, opened 2019-05-07
Allow transition from udev_t to tlp_t BZ(1705246)

zpytela:


wrabcak:
Hi, 
Thanks for PR, is this solution working? Is it tested?

PR #266, opened 2019-05-27
Fix parallel build of the policy

WOnder93:
This is a cherry-pick from https://github.com/SELinuxProject/refpolicy/commit/c1c11fa2f8ca03d47adaf31bca4b257459448c19.

This will allow parallelizing some parts of the RPM build process so the builds will be a bit faster.


wrabcak:
LGTM! 
Thanks,
Lukas.

PR #268, opened 2019-07-19
Add systemd_private_tmp_type attribute

zpytela:
Add systemd_private_tmp_type attribute to back the systemd PrivateTmp feature.
Add systemd_private_tmp interface to assign the attribute and
systemd_delete_private_tmp interface to allow a type to delete objects
on the filesystem.
Allow systemd_delete_private_tmp(init_t).

Based on an idea of Dominic Grift.


wrabcak:
LGTM. 
THanks.

PR #271, opened 2019-08-07
Run lvmdbusd service as lvm_t

5umm3r15:
Label binary /usr/sbin/lvmdbusd as lvm_exec_t to run lvmdbusd service as lvm_t.
Lvmdbusd -  a service which provides a D-Bus API to the logical volume manager (LVM).


5umm3r15:
Fixed Red Hat Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1726166#

PR #272, opened 2019-08-12
Allow dbus communications with resolved for DNS lookups

polarina:
When one adds `resolve` to nsswitch.conf as recommended by
`man nss-resolve`, SELinux blocks DBus communications with
systemd-resolved.

Anyone who wants to do DNS lookups needs to be able to communicate
with systemd-resolved.


wrabcak:
LGTM. 
Thanks!

PR #274, opened 2019-09-10
Move open, audit_access, and execmod to common file perms

WOnder93:
Note that this had kind of been already attempted before, but it broke
libselinux back then (see 056e503ca21a ("Fix access vectors so they do
not break libselinux")). However, since the userspace headers that used
to be generated from the policy source have since been deprecated [1],
it should be safe to do it now.

[1] https://github.com/SELinuxProject/selinux/commit/76913d8adb61b5afe28fd3b4ce91feab29e284dd

Cc: @stephensmalley, @bachradsusi

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -28,6 +28,9 @@ common file
 	swapon
 	quotaon
 	mounton
+	audit_access
+	open
+	execmod
 }
 
 
@@ -208,9 +211,6 @@ inherits file
 	reparent
 	search
 	rmdir
-	open
-	audit_access
-	execmod
 }
 
 class file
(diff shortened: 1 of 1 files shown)


stephensmalley:
This might break if you have a system with binary modules generated using the older common file definition (e.g. audit2allow-generated ones, third party modules) and you update the base policy with these definitions. Now, it might happen to work for any already installed modules because they will have been already converted to cil and cached as cil and thus have the actual permission strings rather than just bits, but if you were to install a .pp file built against the older /usr/share/selinux/devel/include after installing the update, it might break.

stephensmalley:
I think I'm wrong.  Even with a binary module, they must keep the mapping to strings and resolve at link time since the require ordering may differ from the base module order.  Plus it will get turned into cil upon install and that requires the strings.  So I think this should be fine.  Never mind.

wrabcak:
Hi All, 
I agreed with @WOnder93, that we'll test it before merge this PR.

WOnder93:
OK, so I tested the scenario described by @stephensmalley like this:

1. I patched selinux-testsuite to not rebuild the test_policy.pp once it has already been built (this should IMO be the default behavior - official patch coming soon...).
2. I ran the testsuite under current rawhide policy (which built the test policy package under this policy, loaded it, and then unloaded).
3. I installed selinux-policy with this patch + the new watch* permissions added (watch permissions are a separate thing, but that's the build that I had at hand... if anything they should shift the permission numbers even more).
4. I ran the testsuite again, leaving it to use the old test_policy.pp again.

The testsuite passed fine in both cases, so I take it as a proof that changing the numeric mapping of permissions in the policy will not break loading an older policy package file. @stephensmalley does it make sense the way I tested it or do I need to do it differently?

stephensmalley:
It sounds reasonable, and as I mentioned later, I think my original concern was unfounded.

wrabcak:
LGTM. Merging.

PR #280, opened 2019-09-20
Allow systemd-module-load.service to read /run/modules-load.d/

keszybz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1753885.


wrabcak:
Hi @keszybz , 
Thank you for your PR. I made some changes, because we should use interfaces to allow such a operation because init_var_run_t is defined in init.te files instead of systemd.te file. 
It's fixed in these commits: 438e9b79053f89b54c299e5d09d3633a532b2388 and 31db3dc710352793e122ccc2bab65f1de1021a77

Thanks for your help on SELinux for Fedora!
Lukas.

keszybz:
Thanks!

PR #288, opened 2019-10-14
[RFC] Initial support for watch permissions

WOnder93:
This PR implements support for the new watch permissions introduced ([1]) in the upcoming 5.4 kernel release. To allow applications to work, but at the same time detect their need by applications, the watch permissions are initially globally allowed and audited. As domains are gradually onboarded, they can be selectively excluded using the newly added `domain_noautowatch()` macro. Once sufficient coverage is reached, the "auto allow" rules and macros can be removed/deprecated.

This PR also implements the watch permissions for init_t domain (not necessarily complete, but enough to boot a Rawhide VM without watch-related denials for me) and excludes the `files[ystem]_unconfined_type` domains from auto allow of watch permissions (they are already allowed for these by a pre-existing rule).

[1] https://git.kernel.org/pub/scm/linux/kernel/git/pcmoore/selinux.git/commit/?h=next&id=ac5656d8a4cdd93cd2c74355ed12e5617817e0e7


policy/modules/system/systemd.if
```diff
@@ -1046,6 +1046,7 @@ interface(`systemd_manage_passwd_run',`
 	manage_files_pattern($1, systemd_passwd_var_run_t, systemd_passwd_var_run_t)
 	manage_sock_files_pattern($1, systemd_passwd_var_run_t, systemd_passwd_var_run_t)
 	manage_fifo_files_pattern($1, systemd_passwd_var_run_t, systemd_passwd_var_run_t)
+	watch_dirs_pattern($1, systemd_passwd_var_run_t, systemd_passwd_var_run_t)
```
ghost: Just some additional info about this event:

The watch is for the ask-password path units [1]. systemd watches all path unit objects  [2]. Might be a better way to address this automatically for all potential path units by associatiing path unit specific rules and types with a path unit object type attribute,

[1] ls /usr/lib/systemd/system | grep .path$
[2] man systemd.path

ghost: I actually was reminded about this myself just this morning:

https://defensec.nl/gitweb/dssp2.git/commitdiff/883a71a2e383ba0b3c1c71082e86fe6ca9b0dfcf

WOnder93: Oh, right, I see... I'll try to restructure that part in a better way. Thanks for the review!

WOnder93:
OK, I reworked the stuff around ask-password a bit based on @doverride's comments, hopefully it is now more logical.

bachradsusi:
Have you considered to use https://github.com/SELinuxProject/refpolicy/commit/c656b97a289ce6c2da2871700384f0f9d831be18 instead of https://github.com/fedora-selinux/selinux-policy/pull/288/commits/87b4a7672866fa03a82a08d3d088c42288a8bf26 ?

zpytela:
@WOnder93 if you don't mind, I'd like to continue discuttion in https://github.com/fedora-selinux/selinux-policy/pull/546

WOnder93:
@zpytela Yeah, let's continue there.

PR #289, opened 2019-10-31
Allow cephfs to use xattrs for storing contexts

jtlayton:
cephfs recently gained the ability to store SELinux contexts in an xattr
(like most local filesystems). Change the policy to allow for this.

Signed-off-by: Jeff Layton <jlayton@kernel.org>


wrabcak:
Looks good to me. 

Thanks!

PR #293, opened 2019-11-15
Update authlogin policy to allow using google F2A for polydomain

Koncpa:
Polydomain is attribute which contain also sshd_t domain
and cockpit_session_t domain which use google F2A
Add for auth_filetrans_home interface  filetrans rule to auth_home_t for generic file
Allow polydomain to create auth dir in user home dir with correct label auth_home_t
which is used for google F2A file
Create boolean which allow polydomain to manage google F2A
file with label auth_home_t in user home directory

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1767606


wrabcak:
Please create scratch build on copr.fedoraproject.org and attach the scratch build. 

Thanks,
Lukas.

H20-17:
Any ETA on this?

zpytela:
@Koncpa, what's the current status of this PR?

As it containes the first boolean with uppercase letters ever, I wonder if there is a tool which relies on having only lowercase; I'd rather avoid it.

Also note calling auth_  interface does not need to be in an optional block.

PR #300, opened 2019-12-10
Allow confined users run fwupdmgr

zpytela:
Allow confined users dbus_chat with fwupd_t domain
so that fwupdmgr command can be run

### DIFF
--- a/policy/modules/roles/sysadm.te
+++ b/policy/modules/roles/sysadm.te
@@ -269,6 +269,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	fwupd_dbus_chat(sysadm_t)
+')
+
+optional_policy(`
        	gnome_filetrans_fontconfig_home_content(sysadm_t)
 ')
 optional_policy(`
--- a/policy/modules/roles/unprivuser.te
+++ b/policy/modules/roles/unprivuser.te
@@ -66,6 +66,10 @@ optional_policy(`
 ')
 
 optional_policy(`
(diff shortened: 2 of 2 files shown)


zpytela:
The updated package builds, installs, the permissions are granted.
Before:
```
# rpm -q selinux-policy; sesearch -A -t fwupd_t -c dbus -p send_msg|grep -e staff_t -e sysadm_t -e user_t
selinux-policy-3.14.5-18.fc32.noarch
allow staff_t fwupd_t:dbus send_msg;
```
After:
```
# rpm -q selinux-policy; sesearch -A -t fwupd_t -c dbus -p send_msg|grep -e staff_t -e sysadm_t -e user_t
selinux-policy-3.14.5-18.200.bz1772619.fc32.noarch
allow staff_t fwupd_t:dbus send_msg;
allow sysadm_t fwupd_t:dbus send_msg;
allow user_t fwupd_t:dbus send_msg;
```
For staff_t the rule was already present. As guest_t is not allowed to log in a X session, AVC should not be triggered on login; if there is another way how to trigger it, then it's a dontaudit candidate.

Please review.

wrabcak:
What about xguest?

zpytela:
xguest has all except some dbus messages dontaudited with `dbus_dontaudit_chat_system_bus(xguest_t)`

wrabcak:
Cool great. btw. we should check if we're even able to login via xguest, but that's off topic. 

Merging.

PR #302, opened 2019-12-12
Allow user_t and staff_t manage user home directories

zpytela:


zpytela:
See also:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/176

wrabcak:
Hi, 
Thanks for PR, however it's not needed related to fedora-selinux/selinux-policy-contrib#176 

Closing without merging.

PR #303, opened 2019-12-13
Allow init_t setattr on /var/lib/fprintd

zpytela:
Allow init_t setattr on /var/lib/fprintd
using the fprintd_setattr_var_lib() interface

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -741,6 +741,11 @@ optional_policy(`
 ')
 
 optional_policy(`
+	fprintd_read_var_lib_dir(init_t)
+	fprintd_setattr_var_lib_dir(init_t)
+')
+
+optional_policy(`
     ibacm_create_netlink_rdma_socket(init_t)
     ibacm_read_pid_files(init_t)
 ')


zpytela:
This PR is required for proper building:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/177

wrabcak:
LGTM.

ghost:
On Fri, Dec 13, 2019 at 02:35:26AM -0800, Zdeněk Pytela wrote:
> This PR is required for proper building:
> https://github.com/fedora-selinux/selinux-policy-contrib/pull/177

This does not really address the issue that it is trying to solve:

"This may happen when booting with an empty /var filesystem."

https://gitlab.freedesktop.org/libfprint/fprintd/commit/94a98153627b5ef84d914d0009a96bc9050840c7

ie. Systemd needs to be able to create /var/lib/fprintd as well (when /var is empty), not just set attributes.

The benefit of creating and using proper interfaces that correspond to the systemd directives is that you will end up supporting the scenarios that the directives intent to support.

This PR will only lead to a disconnect between the policy and systemd directives. Not to mention that this is not practical to maintain in the long term.

> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/fedora-selinux/selinux-policy/pull/303#issuecomment-565390996

-- 
Key fingerprint = 5F4D 3CDB D3F8 3652 FBD8 02D5 3B6C 5F1D 2C7B 6B02
https://sks-keyservers.net/pks/lookup?op=get&search=0x3B6C5F1D2C7B6B02
Dominick Grift

zpytela:
@doverride thank you for pointing this out, this is the piece I was missing.
The PR was actually meant as a hotfix until the proper solution is found.

PR #305, opened 2019-12-17
Create files_create_non_security_dirs() interface

zpytela:

PR #306, opened 2019-12-17
Create init_create_dirs boolean to allow init create directories

zpytela:
Create init_create_dirs boolean to allow init to create, mounton,
setattr on non_security_file_type directories

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -37,6 +37,13 @@ gen_tunable(daemons_dump_core, false)
 ## </desc>
 gen_tunable(daemons_enable_cluster_mode, false)
 
+## <desc>
+## <p>
+## Enable init create, setattr, mounton on non_security_file_type
+## </p>
+## </desc>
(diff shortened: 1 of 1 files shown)


zpytela:
Requires also a new interface:
https://github.com/fedora-selinux/selinux-policy/pull/305

With both PRs in place, the package builds, installs and the allow rules seem to be in place.

wrabcak:
Should it be backported also to stable versions of Fedora?

zpytela:
@wrabcak, please backport to F31 as I spotted an F31 bug as well.

wrabcak:
LGTM.

bepvte:
Does this obsolete "systemd_allow_mount_dir"?

zpytela:
@bepvte, this commit is a superset of `systemd_allow_mount_dir()`; as it is so powerfull though it is boolean-enabled so that can be disabled by an administrator, while the first approach utilizing `systemd_mount_directory` attribute will keep working.

PR #311, opened 2020-01-14
userdomain: fix indentation

rubenk:


policy/modules/system/userdomain.if
```diff
@@ -6265,12 +6267,12 @@ interface(`userdom_rw_unpriv_user_semaphores',`
 ##	</summary>
 ## </param>
 #
-interface(`userdom_filetrans_home_content',`
+interface(`userdom_filetrans_home_content',
```
WOnder93, line 6270: You removed a backtick here - that's probably why the Travis build failed.

vmojzis:
I guess we can close this since https://github.com/fedora-selinux/selinux-policy/commit/bc14449a56fa91109cd2889e16ff6e3ffcb0e9dc

PR #313, opened 2020-01-20
Allow init_t set the nice level of all domains BZ(1778088)

zpytela:


wrabcak:
LGTM.

zpytela:
A scratch build has been successfully created and tested.

As this is a systemd v244 feature, no need for backporting as of now.

PR #319, opened 2020-01-30
Add file transition for /dev/nvidia-uvm BZ(1770588)

zpytela:


zpytela:
Reported on F31.

wrabcak:
@zpytela , 
The fix is not complete, there is no record in devices.fc to assign right context to the path. The same issue is with all nvidia devices.

zpytela:
devices.fc already contains:
```
/dev/nvidia.*           -c      gen_context(system_u:object_r:xserver_misc_device_t,s0)
```

wrabcak:
Cool, you're right. 

Thanks.

PR #308, opened 2020-01-10
Allow init_t domain to settatribute to domain event_device_t

Koncpa:
Added macro dev_setattr_input_dev(init_t)
systemd_udevd which have domain init_t can settatr of evdev chr_file
Bugzilla1785804

Build COPR: https://copr.fedorainfracloud.org/coprs/pkoncity/selinux-policy/build/1142807/

PR #320, opened 2020-01-30
Allow init_t to create apache log files.

5umm3r15:
PR for new apache create log dirs macro: fedora-selinux/selinux-policy-contrib#194

Fixed Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1789868

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -437,6 +437,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+    apache_create_log_dirs(init_t)
     apache_delete_tmp(init_t)
     apache_noatsecure(init_t)
 ')


WOnder93:
I don't think this is the right fix - from the BZ report it looks like the lighthttpd service is running as init_t, which it should not. See #308, which I think was the same/similar case.

zpytela:
@WOnder93, in this case it really is systemd that creates the directories if they don't exist,

wrabcak:
@zpytela @WOnder93 , 

This is because of systemd feature described here:
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

The dir really doesn't exists before the service is stared, therefore we see init_t domain type.

zpytela:
@5umm3r15 why did you add more permissions than just create and setattr? Did you came across another use case than in the bugzilla?

wrabcak:
@zpytela , 
 I see that @5umm3r15  is using new macro defined here:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/194/commits/53101c05e8f320dae98cb852a748b3e244e7980b
Macro using patterns for create and setattr . And patterns are defines as:
```
define(`setattr_dir_perms',`{ setattr }')
"define(`create_dir_perms',`{ getattr create }')"
```

WOnder93:
Ah, I see... I was confused by the "comm" field, which is probably modified by systemd for debugging purposes.

zpytela:
So the sesearch command output in c#1 is not valid any longer:
```
# sesearch -A -s init_t -t httpd_log_t -p create -c dir
allow init_t httpd_log_t:dir { add_name create setattr write };
```
This is what confused me.

5umm3r15:
@zpytela thanks, it is fixed

PR #322, opened 2020-02-03
Allow systemd-sleep access swapfiles BZ(1797543)

zpytela:
Add fstools_read_swap_files() interface.
Allow init_t read swapfile_t files.


wrabcak:
Based on the AVC in Red Hat Bugzilla, maybe we need to create new systemd domain for that.

zpytela:
Closing this PR, a new domain seems to be reasonable.

PR #326, opened 2020-02-07
Allow getty_t stream socket connect to plymouthd_t

zpytela:
Resolves: rhbz#1798956


zpytela:
This should go to rawhide only.

wrabcak:
LGTM.

PR #327, opened 2020-02-10
Allow systemd system services read efivarfs files

zpytela:
Allow systemd_hostnamed_t, systemd_hwdb_t, systemd_localed_t, systemd_machined_t,
systemd_rfkill_t, systemd_sysctl_t, systemd_timedated_t read efivarfs files.
 
Systemd since v244 can read configuration options from the EFI variable
SystemdOptions to configure systemd behaviour when modifying the kernel
command line is inconvenient, but configuration on disk is read too
late, e. g. for the options related to cgroup hierarchy setup.

Resolves: rhbz#1800935

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -204,8 +204,6 @@ fs_list_tmpfs(systemd_logind_t)
 fs_list_dos(systemd_logind_t)
 fs_read_dos_files(systemd_logind_t)
 
-fs_read_efivarfs_files(systemd_logind_t)
-
 fs_manage_fusefs_dirs(systemd_logind_t)
 fs_manage_fusefs_files(systemd_logind_t)
 
@@ -264,6 +262,7 @@ fs_getattr_tmpfs(systemd_logind_t)
 fs_read_tmpfs_symlinks(systemd_logind_t)
 fs_mount_tmpfs(systemd_logind_t)
 fs_delete_tmpfs_files(systemd_logind_t)
+fs_read_efivarfs_files(systemd_logind_t)
 userdom_mounton_tmp_dirs(systemd_logind_t)
 
 storage_setattr_removable_dev(systemd_logind_t)
@@ -379,6 +378,7 @@ manage_files_pattern(systemd_machined_t, systemd_machined_var_lib_t, systemd_mac
 manage_lnk_files_pattern(systemd_machined_t, systemd_machined_var_lib_t, systemd_machined_var_lib_t)
 init_var_lib_filetrans(systemd_machined_t, systemd_machined_var_lib_t, dir, "machines")
 
+fs_read_efivarfs_files(systemd_machined_t)
 fs_read_nsfs_files(systemd_machined_t)
 
 kernel_dgram_send(systemd_machined_t)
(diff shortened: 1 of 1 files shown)


wrabcak:
Hi @zpytela , 

Do we know why systemd services requires access to efivarfs files? 

You allowed it for each service explicitly, why you didn't used systemd_domain attribute? 

You also put multiple bugs as resolves. In most of them there is no reference to this github PR and why we don't want to handle it in one BZ? 

Thanks,
Lukas.

ghost:
https://github.com/SELinuxProject/refpolicy/pull/148 explains it rather well

zpytela:
@wrabcak, not all services in systemd_domain require reading efivarfs.
With the update, I added 3 more domains.
I will deal with the bugzillas references once things are sorted out.

wrabcak:
@zpytela 
> @wrabcak, not all services in systemd_domain require reading efivarfs.

Is there some pattern which services require it? Would it make sense to create some new domain for it? 

> With the update, I added 3 more domains.
> I will deal with the bugzillas references once things are sorted out.

If you want to deal with the PR in future, should we consider it as WIP PR?

zpytela:
@wrabcak, I don't think domain can be a solution. I was considering an attribute but declined.
This is not a WIP PR, rather a request for review.

wrabcak:
> @wrabcak, I don't think domain can be a solution. I was considering an attribute but declined.

Yes, I meant attribute. Why it was declined?  

> This is not a WIP PR, rather a request for review.

Attribute could be good candidate for this issue and also it would be great to keep bugzilla communication on one place.

zpytela:
> > @wrabcak, I don't think domain can be a solution. I was considering an attribute but declined.
> 
> Yes, I meant attribute. Why it was declined?
I was not able to see the benefit for one-purpose attribute, so decided this way based on balance between number of rules and overhead of a new attribute. I understand it always is debatable.

> 
> > This is not a WIP PR, rather a request for review.
> > Attribute could be good candidate for this issue and also it would be great to keep bugzilla communication on one place.
As said previously, I will address it when all new bugzillas are sorted out.

wrabcak:
> > > @wrabcak, I don't think domain can be a solution. I was considering an attribute but declined.
> > 
> > 
> > Yes, I meant attribute. Why it was declined?
> > I was not able to see the benefit for one-purpose attribute, so decided this way based on balance between number of rules and overhead of a new attribute. I understand it always is debatable.
> 
Personally I like maybe new attribute for that like we did with "systemd_private_tmp_type", but I'm also fine with solution without attribute. Could you please put comments to the commit so we know why it was added? Thanks. 

> > > This is not a WIP PR, rather a request for review.
> > > Attribute could be good candidate for this issue and also it would be great to keep bugzilla communication on one place.
> > > As said previously, I will address it when all new bugzillas are sorted out.

But what is the plan by sorting it out? After some while when somebody would like to see also bugzilla ticket there will be 5 references and most of them will be closed as duplicate. This could be confusing. All tickets solving same issue for different domains.

zpytela:
Resolved, commit is ready for review again.

wrabcak:
LGTM. 
Thanks,
Lukas.

PR #329, opened 2020-02-14
Add support for systemd-userdbd

zpytela:
The systemd-userdbd.service is a new service in systemd v245,
allowing rich user and group records. It is a lightweight system service
to translate UNIX/glibc NSS records to JSON user records, eventually
used by other systemd services.

This commit adds the following interfaces:
systemd_userdbd_rw_data() to allow access to userdb data,
systemd_userdbd_data_manage_sock_files() and systemd_userdbd_stream_connect()
to allow communication using userdb sockets,
systemd_userdbd_data_pid_filetrans() for its data directory transition.

Resolves: rhbz#1798883

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -556,6 +556,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	systemd_userdbd_stream_connect(nsswitch_domain)
+')
+
+optional_policy(`
     virt_read_lib_files(nsswitch_domain)
 ')
 
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -2256,3 +2256,65 @@ interface(`systemd_delete_private_tmp',`
 	allow $1 systemd_private_tmp_type:lnk_file delete_lnk_file_perms;
 	allow $1 systemd_private_tmp_type:sock_file delete_sock_file_perms;
 ')
+
+#######################################
+## <summary>
+##	Create objects in the pid directory
+##	with a private type with a type transition.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_userdbd_runtime_filetrans',`
+	gen_require(`
+		type init_var_run_t;
+		type systemd_userdbd_runtime_t;
+	')
+
+	filetrans_pattern($1, init_var_run_t, systemd_userdbd_runtime_t, dir, "userdb")
+')
+
+#######################################
+## <summary>
+##	Manage systemd-userdbd data sockets.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_userdbd_runtime_manage_sockets',`
+	gen_require(`
+		type systemd_userdbd_runtime_t;
+	')
+
+	manage_sock_files_pattern($1, systemd_userdbd_runtime_t, systemd_userdbd_runtime_t);
+')
+
+#######################################
+## <summary>
+##	Connect to systemd-userdbd with a unix socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_userdbd_stream_connect',`
+	gen_require(`
+		type systemd_userdbd_t;
+        type systemd_userdbd_runtime_t;
+	')
+
+	files_search_pids($1)
+	list_dirs_pattern($1, systemd_userdbd_runtime_t, systemd_userdbd_runtime_t)
+	read_lnk_files_pattern($1, systemd_userdbd_runtime_t, systemd_userdbd_runtime_t)
+	write_sock_files_pattern($1, systemd_userdbd_runtime_t, systemd_userdbd_runtime_t)
+
+	allow $1 systemd_userdbd_t:unix_stream_socket connectto;
+')
--- a/policy/modules/services/ssh.te
+++ b/policy/modules/services/ssh.te
@@ -415,6 +415,7 @@ optional_policy(`
 
 optional_policy(`
 	systemd_exec_systemctl(sshd_t)
+	systemd_userdbd_stream_connect(sshd_t)
 ')
 
 optional_policy(`
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -682,6 +682,8 @@ systemd_timedated_manage_lib_dirs(init_t)
 systemd_login_mounton_pid_dirs(init_t)
 systemd_mounton_inherited_logind_sessions_dirs(init_t)
 systemd_delete_private_tmp(init_t)
+systemd_userdbd_stream_connect(init_t)
+systemd_userdbd_runtime_filetrans(init_t)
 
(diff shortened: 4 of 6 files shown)


zpytela:
Scratchbuild available here:
https://koji.fedoraproject.org/koji/taskinfo?taskID=41505126

The system boots with new systemd, systemd-userdbd service runs, users can be added, users can log in using ssh.
It will be tested further later, this is a first version sent for review.

policy/modules/system/authlogin.te
```diff
@@ -555,6 +555,10 @@ optional_policy(`
 	samba_dontaudit_write_var_files(nsswitch_domain)
 ')
 
+optional_policy(`
+	systemd_userdbd_rw_data(nsswitch_domain)
```
ghost: Only root can write to /run/systemd/userdb so this is too broad

There are three sock files maintained in there:
* dynamicuser. this is used by nss_systemd (nsswtch_domain) to connect to systemd (init_t) with a unix stream socket.
* multiplexer: this is used by pam clients (programs that has a pam config in /etc/pam.d)  to connect to userdbd with a unix domain socket
* home: this is used by nss_systemd (nsswitch_domain) to connect to homed with a unix stream socket (if home is enabled)

policy/modules/system/systemd.if
```diff
@@ -2256,3 +2256,80 @@ interface(`systemd_delete_private_tmp',`
...
+##	</summary>
+## </param>
+#
+interface(`systemd_userdbd_data_pid_filetrans',`
+	gen_require(`
+		type init_var_run_t;
+		type systemd_userdbd_data_t;
+	')
+
+	filetrans_pattern($1, init_var_run_t, systemd_userdbd_data_t, dir, "userdb")
```
ghost: I would not call this "data", instead "runtime" seems more appropriate

policy/modules/system/systemd.if
```diff
@@ -2256,3 +2256,80 @@ interface(`systemd_delete_private_tmp',`
...
+#
+interface(`systemd_userdbd_rw_data',`
+	gen_require(`
+		type systemd_userdbd_data_t;
+	')
+
+	list_dirs_pattern($1, systemd_userdbd_data_t, systemd_userdbd_data_t)
+	read_lnk_files_pattern($1, systemd_userdbd_data_t, systemd_userdbd_data_t)
+	rw_sock_files_pattern($1, systemd_userdbd_data_t, systemd_userdbd_data_t)
+')
```
ghost, line 2296: o now i see... this interface is not what the name suggests. Still though, i think this too broad

ghost: ... but just a little. I dont think the sock files have to be readable

policy/modules/system/systemd.if
wrabcak: @doverride , yes I also think runtime is more appropriate. Will change it.

policy/modules/system/systemd.if
wrabcak: @doverride  What you mean by last comment? Should nsswitch domains only write to the sockets ?

policy/modules/system/authlogin.te
wrabcak: @doverride  Are you suggesting to label each socket with different label?

policy/modules/system/systemd.if
ghost: Yes just for regular unix stream connect i think. (but seems the entities that write to the dynamicuser and homed sock files to stream oonnect to init_t and homed_t respectively also list the /run/systemd/userdb directory)

ghost: the read_lnk_files is only for entities that connect to userdbd via the multiplexer socket ( there is a symlink in /run/systemd/userdb  that points to the multiplexer sock ffile. (the multiplexer sock file is only used by pam clients (including authlogin_login_pgm_domain) , and `userdbctl`  to unix stream connect to userdbd

policy/modules/system/authlogin.te
ghost: I don't do that myself no but should be possible. You can differentiate using the "connectto" permission:

* init_t:unix_stream_socket connectto (for dynamicuser)
* userdbd_t:unix_stream_socket connectto (for multiplexer)
* homed_t::unix_stream_socket connectto (for home)

ghost: from `man userdbctl`
```
WELL-KNOWN SERVICES
       The userdbctl services command will list all currently running services that provide user or group definitions to the system. The
       following are well-known services are shown among this list.

       io.systemd.DynamicUser
           This service is provided by the system service manager itself (i.e. PID 1) and makes all users (and their groups) synthesized
           through the DynamicUser= setting in service unit files available to the system (see systemd.exec(5) for details about this setting).

       io.systemd.Home
           This service is provided by systemd-homed.service(8) and makes all users (and their groups) belonging to home directories managed by
           that service available to the system.

       io.systemd.Multiplexer
           This service is provided by systemd-userdbd.service(8) and multiplexes user/group look-ups to all other running lookup services.
           This is the primary entry point for user/group record clients, as it simplifies client side implementation substantially since they
           can ask a single service for lookups instead of asking all running services in parallel.  userdbctl uses this service preferably,
           too, unless --with-nss= or --service= are used, in which case finer control over the services to talk to is required.

       io.systemd.NameSeviceSwitch
           This service is (also) provided by systemd-userdbd.service(8) and converts classic NSS/glibc user and group records to JSON
           user/group records, providing full backwards compatibility. Use --with-nss=no to disable this compatibility, see above. Note that
           compatibility is actually provided in both directions: nss-systemd(8) will automatically synthesize classic NSS/glibc user/group
           records from all JSON user/group records provided to the system, thus using both APIs is mostly equivalent and provides access to
           the same data, however the NSS/glibc APIs necessarily expose a more reduced set of fields only.

       Note that userdbctl has internal support for NSS-based lookups too. This means that if neither io.systemd.Multiplexer nor
       io.systemd.NameSeviceSwitch are running look-ups into the the basic user/group databases will still work.
```

ghost: note that the "homed" service can be optional. You dont have to have systemd-homed enabled.

policy/modules/system/systemd.if
ghost: So all in all if the read permission is removed from sock_file then this interface could in theory be used like so:
```
interface(`userdbd_unix_stream_connect_dynamicuser',`
# io.systemd.DynamicUser
allow $1 init_t:unix_stream_socket connectto;
systemd_userdbd_rw_data($1)
')
interface(`userdbd_unix_stream_connect',`
# io.systemd.Multiplexer and io.systemd.NameServiceSwitch
allow $1 userdbd_t:unix_stream_socket connectto;
systemd_userdbd_rw_data($1)
')
interface(`userdbd_unix_stream_connect_home',`
# io.systemd.Home
allow $1 homed_t:unix_stream_socket connectto;
systemd_userdbd_rw_data($1)
')
```
obviously the above example is not proper but its just to give an indication of how it is grouped in my experience.

policy/modules/system/authlogin.te
wrabcak: Okay, I'll keep it as we have it now. Fedora policy should not be so strict.

wrabcak:
Let's push it to the Rawhide. @doverride  thanks again for review!

PR #330, opened 2020-02-28
Allow machinectl pull-tar a container image

zpytela:
Allow systemd_importd_t dbus_manage_session_tmp_dirs().
Allow systemd_machined_t domain_read_all_domains_state().
Update systemd_machined_manage_lib_files() interface to manage also lnk_files.

Resolves: rhbz#1724247

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1171,7 +1171,7 @@ init_stream_connectto(systemd_initctl_t)
 #
 # systemd_importd local policy
 #
-allow systemd_importd_t self:capability { chown fowner fsetid mknod setpcap sys_admin };
+allow systemd_importd_t self:capability { chown dac_override fowner fsetid mknod setfcap setpcap sys_admin };
 allow systemd_importd_t self:process { setcap setfscreate };
 allow systemd_importd_t self:unix_stream_socket create_stream_socket_perms;
 allow systemd_importd_t self:unix_dgram_socket create_socket_perms;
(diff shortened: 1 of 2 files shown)


zpytela:
Applicable to all fedoras.

zpytela:
The dac_override and setfcap capabilities are required in F32.

policy/modules/system/systemd.te
```diff
@@ -396,6 +396,7 @@ kernel_dgram_send(systemd_machined_t)
 # This is a bug, but need for now.
 kernel_read_unlabeled_state(systemd_machined_t)
 
+domain_read_all_domains_state(systemd_machined_t)
```
wrabcak: Is this really needed? If yes, do we know why?

wrabcak:
in commit msg I see this line:
Allow systemd_importd_t dac_override and setfcap capabilities. 
Where it is allowed?

zpytela:
The capabilities were present, then removed, then put back again, I just did not manage to keep changes synchronized with the commit message.

wrabcak:
So, now is PR ready to review?

policy/modules/system/systemd.te
zpytela: It is not required for all domain state. systemd_machined_t reads cgroup of the container systemd process running as unconfined_service_t.
The parent process is systemd-nspawn. It looks like confining nspawn would help to narrow the required permissions. Atm systemd-nspawn runs as unconfined_service_t, too.

mochaaP:
I hope this could raise more awareness. It's been a while.

PR #331, opened 2020-03-05
Allow auditd poweroff or switch to single mode

zpytela:
Allow auditd_t service start and status power_unit_file_t.
Allow auditd_t read initrc_var_run_t.


wrabcak:
LGTM.

PR #332, opened 2020-03-13
Allow login_pgm create and bind on netlink_selinux_socket

zpytela:
Required for the latest pam version which started to use
selinux_check_access() instead of security_compute_av().

Resolves: rhbz#1813023


zpytela:
Going to create a scratchbuild to see if these permissions are sufficient.

PR #333, opened 2020-03-17
Makefile: fix tmp/%.mod.fc target

vmojzis:
%.pp target requires both "tmp/%.mod" and "tmp/%.mod.fc", however only the
former creates the tmp directory when it does not exist. This may lead to
failure since order of satisfying dependencies is not defined.

Fixes:
   #git clone https://github.com/fedora-selinux/selinux-policy-contrib.git -b rawhide
   #cd selinux-policy-contrib
   #make -f /usr/share/selinux/devel/Makefile tmp/ipa.mod.fc
   /bin/sh: tmp/ipa.mod.fc: No such file or directory
   make: *** [/usr/share/selinux/devel/include/Makefile:160: tmp/ipa.mod.fc] Error 1

### DIFF
--- a/support/Makefile.devel
+++ b/support/Makefile.devel
@@ -157,6 +157,7 @@ tmp/%.mod: $(m4support) tmp/all_interfaces.conf %.te
 	$(verbose) $(CHECKMODULE) -m $(@:.mod=.tmp) -o $@
 
 tmp/%.mod.fc: $(m4support) %.fc
+	@test -d $(@D) || mkdir -p $(@D)
 	$(verbose) $(M4) $(M4PARAM) $^ > $@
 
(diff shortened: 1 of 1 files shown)


vmojzis:
This issue is blocking freeipa build in rawhide.

wrabcak:
LGTM. Merging. 
@zpytela, we need Fedora Rawhide build with this change to allow freeipa make build.

WOnder93:
Just a side note: In refpolicy, this exact same fix was merged over 5 years ago:
https://github.com/SELinuxProject/refpolicy/commit/ad2d8287973d236532b6a0a1eae5c23c004a4733

zpytela:
@wrabcak, understand, will prepare all builds soon.

PR #338, opened 2020-03-31
Add file context entry and file transition for /var/run/pam_timestamp

zpytela:
Add file context entry and file transition for /var/run/pam_timestamp
directory in order to allow pam_timestamp module to create a timestamp
file in a similar manner to how /var/run/sudo was previously used.


wrabcak:
LGTM

PR #340, opened 2020-04-01
Change file context for /var/run/pam_ssh to match file transition

zpytela:
File context for /var/run/pam_ssh was added by 153ed8751a. The file
transitions in auth_manage_pam_pid() were updated with c3bb30721e
including pam_ssh, but the fc file was not updated accordingly.


wrabcak:
LGTM.

PR #341, opened 2020-04-03
Modify path for arping in netutils.fc to match both bin and sbin

zpytela:
In iputils newer than iputils-20180629 the arping command moved from
/sbin to /bin, hence the path in the netutils.fc file needs to be
adjusted so that it matches both possible paths for the actual file.
Symlink can have bin_t assigned.

Resolves: rhbz#1820191

### DIFF
--- a/policy/modules/admin/netutils.fc
+++ b/policy/modules/admin/netutils.fc
@@ -2,7 +2,7 @@
 /bin/tracepath.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 /bin/traceroute.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 
-/sbin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
Applicable to F31 and newer.

wrabcak:
I'm confused by commit msg. You want to label /bin/arping as netutils_exec_t and keep /sbin/arping labeled as bin_t ? 
If yes, then it's confusing also in netutils.fc file because you're trying to label /sbin/arping and /bin/arping as netutils_exec_t but specifying only files (not symlink therefore /bin/arping is bin_t). Do I understand it correctly?

zpytela:
The goal was to label plain files regardless of iputils version and therefore regardless of the path. Symlinks are correct to have the default bin_t.

wrabcak:
Good. Just to complete the msg: domain attribute can read symlinks labeled as  bin_t . So there is no need to allow it for netutils_t . 

LGTM. 
Thanks.

PR #345, opened 2020-04-09
Introduce daemons_dontaudit_scheduling boolean

zpytela:
Introduce daemons_dontaudit_scheduling boolean to dontaudit daemons
the setsched process permission and sys_nice capability which are requested
to modify various scheduling information.

With a change introduced to glib2 [1], all daemons using this library start
to require the setsched permission and sys_nice capability. As allowing them
to all daemons is too generic, dontauditing is used instead, but still can be
allowed for particular domain types. With later glib2 updates, library call
fallbacks to different code path if the permission is not allowed so a daemon
does not crash. For more info, see the discussion in rhbz#1795524.

[1] https://gitlab.gnome.org/GNOME/glib/issues/1834

With this update, the previous dontaudit setsched rule was moved to a tunable
block and a dontaudit sys_nice rule was added. The boolean is enabled by default
meaning the denied permissions will be dontaudited. After turning it off,
the denials start to be audited.

Resolves: rhbz#1811407

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -11,6 +11,13 @@ gen_require(`
 
 ## <desc>
 ## <p>
+## Dontaudit all daemons scheduling requests (setsched, sys_nice).
+## </p>
+## </desc>
+gen_tunable(daemons_dontaudit_scheduling, true)
+
+## <desc>
+## <p>
 ## Allow all daemons to use tcp wrappers.
 ## </p>
 ## </desc>
@@ -1271,6 +1278,11 @@ domain_dontaudit_use_interactive_fds(daemon)
 userdom_dontaudit_list_admin_dir(daemon)
 userdom_dontaudit_search_user_tmp(daemon)
 
+tunable_policy(`daemons_dontaudit_scheduling',`
+	dontaudit daemon self:process { setsched };
+	dontaudit daemon self:capability { sys_nice };
+')
+
 tunable_policy(`daemons_use_tcp_wrapper',`
     corenet_tcp_connect_auth_port(daemon)
 ')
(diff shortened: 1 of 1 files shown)


zpytela:
In the meantime, testing a scratchbuild to see the result.

zpytela:
Seems to work as expected. No sys_nice audited, but after turning the boolean off both permissions are:
```
type=PROCTITLE msg=audit(04/09/2020 11:27:19.954:168) : proctitle=/usr/libexec/accounts-daemon
type=SYSCALL msg=audit(04/09/2020 11:27:19.954:168) : arch=x86_64 syscall=sched_setattr success=yes exit=0 a0=0x290 a1=0x55df23318f30 a2=0x0 a3=0x7f18bd103a40 items=0 ppid=1 pid=656 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=accounts-daemon exe=/usr/libexec/accounts-daemon subj=system_u:system_r:accountsd_t:s0 key=(null)
type=AVC msg=audit(04/09/2020 11:27:19.954:168) : avc:  denied  { setsched } for  pid=656 comm=accounts-daemon scontext=system_u:system_r:accountsd_t:s0 tcontext=system_u:system_r:accountsd_t:s0 tclass=process permissive=1
type=AVC msg=audit(04/09/2020 11:27:19.954:168) : avc:  denied  { sys_nice } for  pid=656 comm=accounts-daemon capability=sys_nice  scontext=system_u:system_r:accountsd_t:s0 tcontext=system_u:system_r:accountsd_t:s0 tclass=capability permissive=1
```

zpytela:
This is for F32 and rawhide only.

wrabcak:
LGTM.

PR #346, opened 2020-04-09
Introduce xdm_manage_bootloader booelan

zpytela:
Created xdm_manage_bootloader boolean to create, read, write, and delete files in the /boot director & DOS filesystem.

Resolves: rhbz#1822256


zpytela:
Seems to require this PR first:
https://github.com/fedora-selinux/selinux-policy/pull/347

PR #349, opened 2020-04-20
Allow strongswan use tun/tap devices and keys

zpytela:
Allow ipsec_mgmt_t create tun_socket.
Allow ipsec_mgmt_t read and write tun_tap_device_t devices.
Label directories with keys in /etc/strongswan/swanctl/ as
ipsec_key_file_t.

### DIFF
--- a/policy/modules/system/ipsec.fc
+++ b/policy/modules/system/ipsec.fc
@@ -11,6 +11,13 @@
 /etc/ipsec\.conf		--	gen_context(system_u:object_r:ipsec_conf_file_t,s0)
 /etc/strongswan/ipsec\.secrets.*		--	gen_context(system_u:object_r:ipsec_key_file_t,s0)
 /etc/strongswan/ipsec\.conf		--	gen_context(system_u:object_r:ipsec_conf_file_t,s0)
+/etc/strongswan/swanctl/bliss/(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/ecdsa(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/pkcs.*		gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/private(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/pubkey(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/rsa(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
+/etc/strongswan/swanctl/x509.*		gen_context(system_u:object_r:ipsec_key_file_t,s0)
 /etc/racoon/psk\.txt		--	gen_context(system_u:object_r:ipsec_key_file_t,s0)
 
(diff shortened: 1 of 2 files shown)


zpytela:
Applicable to all supported Fedoras.

policy/modules/system/ipsec.te
```diff
@@ -133,10 +133,12 @@ can_exec(ipsec_t, ipsec_exec_t)
 corecmd_shell_domtrans(ipsec_t, ipsec_mgmt_t)
 allow ipsec_t ipsec_mgmt_t:process2 nnp_transition;
 
+allow ipsec_mgmt_t self:tun_socket create;
```
wrabcak: Do you sure just permissions "create" is enough? No need for more permissions?

policy/modules/system/ipsec.te
```diff
@@ -133,10 +133,12 @@ can_exec(ipsec_t, ipsec_exec_t)
...
 allow ipsec_mgmt_t ipsec_t:fd use;
 allow ipsec_mgmt_t ipsec_t:fifo_file rw_fifo_file_perms;
 allow ipsec_mgmt_t ipsec_t:unix_stream_socket { read write };
 allow ipsec_mgmt_t ipsec_t:process { rlimitinh sigchld signull };
+allow ipsec_mgmt_t tun_tap_device_t:chr_file rw_chr_file_perms;
```
wrabcak: Why macro is not used here?

policy/modules/system/ipsec.fc
```diff
@@ -11,6 +11,13 @@
 /etc/ipsec\.conf		--	gen_context(system_u:object_r:ipsec_conf_file_t,s0)
 /etc/strongswan/ipsec\.secrets.*		--	gen_context(system_u:object_r:ipsec_key_file_t,s0)
 /etc/strongswan/ipsec\.conf		--	gen_context(system_u:object_r:ipsec_conf_file_t,s0)
+/etc/strongswan/swanctl/bliss/(/.*)?	gen_context(system_u:object_r:ipsec_key_file_t,s0)
```
wrabcak, line 14: Would it be possible to label whole directory /etc/strongswan/swanctl as ipsec_key_file_t ? Or are there any other subdirs which requires another label?

wrabcak (changes requested):

zpytela:
In the bugzilla, only create was reported, but you are right it is safer to use common object permissions sets.

wrabcak:
> In the bugzilla, only create was reported, but you are right it is safer to use common object permissions sets.

To which bugzilla ticket are you referring to ? I  see no reference in the commit msg.

zpytela:
https://bugzilla.redhat.com/show_bug.cgi?id=1823173
Commit message updated.

policy/modules/system/ipsec.fc
zpytela: Labeling the whole directory with a few exception would be anothor approach. I chose this one as in swan wiki https://wiki.strongswan.org/projects/strongswan/wiki/SwanctlDirectory
they call it a configuration directory and the credentials dirs predefined. Right after installation it contains, apart from the credential directories, a configuration file and a directory.

> The swanctl configuration directory, usually /etc/swanctl, contains swanctl.conf and a predefined
> set of sub-directories that provide file based credentials, such as private keys and certificates, which are
> read by the swanctl --load-creds command.

wrabcak: I see, it make sense. Different approach would label possible future sub-dirs. Let's follow as it is in commit.

wrabcak (approved):

PR #350, opened 2020-04-27
Remove references to the j variable which is never used in fc_sort.c

zpytela:
Remove declaration and assigning to the j variable in the fc_sort.c
file. This variable is never read from according to the coverity scan test.


wrabcak:
LGTM.

PR #351, opened 2020-05-07
Change arping path regexp to work around how fixfiles handles it

zpytela:
Resolves: rhbz#1832790

### DIFF
--- a/policy/modules/admin/netutils.fc
+++ b/policy/modules/admin/netutils.fc
@@ -2,7 +2,7 @@
 /bin/tracepath.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 /bin/traceroute.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 
-/s?bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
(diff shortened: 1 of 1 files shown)


WOnder93:
This not a "fix", but rather just a workaround. It is `fixfiles` that is broken. I would prefer if the commit title reflected that. Also, please add an explanation to the commit message about what exactly is wrong and why. IIMO, it should include at least this information:
1. fixfiles is run during selinux-policy update, trying to relabel only parts of the filesystem for which the label has changed,
2. due to a bug, fixfiles translates the `/s?bin/arping` regex to `/*`,
3. that leads to the whole system being relabeled, which takes a *very* long time.

(I assume it was late and you just wanted to post a solution quickly, but urgency of the problem is not a good excuse for rushing the commit message ;)

zpytela:
@WOnder93, you are right in all points. However, that's what reviews are for ;).

zpytela:
Note the same regexp is in f31+, so if this pr is accepted, it needs to be merged to f31, f32, and rawhide.

WOnder93 (approved):
Thanks, it looks great now :+1:

wrabcak:
Backported to F31 and F32

zpytela:
I am afraid the fixfiles engine cannot be fooled that easily :( It rather needs to be replaced by 2 lines.

PR #352, opened 2020-05-11
Revert "Change arping path regexp to work around fixfiles incorrect h…

zpytela:
…andling"

This reverts commit de454edb84972514a86a27252ff82cc10c1f5589.
The proposed change did not work around the fixfiles behaviour as expected
since any quantifier early in the file context path specification will be
replaced with the '*' wildcard. Moreover, it would introduce additional,
needless relabeling on the next update.


wrabcak:
LGTM.

PR #353, opened 2020-05-12
Split the path regexp for arping to 2 lines

zpytela:
Split the "/s?bin/arping" file context regexp to 2 lines to prevent
from relabeling the complete filesystem on the package update as regexp
quantifiers are replaced with "*" wildcards in fixfiles.

Resolves: rhbz#1832790

### DIFF
--- a/policy/modules/admin/netutils.fc
+++ b/policy/modules/admin/netutils.fc
@@ -2,7 +2,10 @@
 /bin/tracepath.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 /bin/traceroute.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
 
-/s?bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
+# regexp can't be used as fixfiles converts any quantifier early in the path to the
+# "/*" wildcard, effecting in complete filesystem restorecon on the package update
+/bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
+/sbin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
 
 /usr/bin/lft		--	gen_context(system_u:object_r:traceroute_exec_t,s0)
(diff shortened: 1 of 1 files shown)


bachradsusi:
It would be great to have a comment directly next to these 2 lines with a short explanation why it's split and maybe with a pointer to the userspace bug.

bachradsusi:
https://bugzilla.redhat.com/show_bug.cgi?id=1832327#c5  it looks like it doesn't make sense to do this change and trigger full restorecon again.

zpytela:
@bachradsusi

> It would be great to have a comment directly next to these 2 lines with a short explanation why it's split and maybe with a pointer to the userspace bug.

@bachradsusi, I am not quite getting what you mean. I've changed the commit message to be mroe descriptive.

bachradsusi:
I was referring to the code:

~~~
- /s?bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
+ /bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
+ /sbin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
~~~

But https://github.com/fedora-selinux/selinux-policy/pull/353#issuecomment-627244846

zpytela:
> https://bugzilla.redhat.com/show_bug.cgi?id=1832327#c5 it looks like it doesn't make sense to do this change and trigger full restorecon again.

It makes sense for rawhide an F31 where the previous change hasn't been applied yet.
For F32 I understand it is questionable. However, this latest change will affect only installation with the selinux-policy-3.14.5-38.fc32 package already installed and no other version.

bachradsusi:
I see, the original version was `/sbin/arping`

wrabcak:
From BZ and communication here I understand that this PR is ready to merge right? @zpytela @bachradsusi 
Thanks.

zpytela:
I added the in-line comment as requested and changed the commit message.

wrabcak:
LGTM.

PR #356, opened 2020-05-28
Add file context for /sys/kernel/tracing

zpytela:


zpytela:
This should go to F32+ only, F31 has sysfs_t.

zpytela:
Merging per internal discussion results.

PR #359, opened 2020-06-02
Allow named transition for /run/tlp from a user shell

zpytela:
Resolves: rhbz#1806123


zpytela:
Requires a new interface from:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/256

wrabcak:
LGTM

PR #360, opened 2020-06-03
Introduce systemd_read_efivarfs_type attribute

zpytela:
Introduce systemd_read_efivarfs_type attribute for domains to read efivarfs.
Create a systemd_read_efivarfs() interface to allow a domain be part of
the systemd_read_efivarfs_type attribute.
Replace fs_read_efivarfs_files() calls in systemd.te
with systemd_read_efivarfs().


wrabcak:
I think this PR should ehance also this commit:
ff8b5f9c119a828e92036f86e3d82c898412db59

zpytela:
@wrabcak, this is not possible as `systemd_exec_systemctl` is called for `unconfined_domain_type`.

wrabcak:
Okay, so let's merge it and test it.

PR #372, opened 2020-06-08
Do not audit staff_t and user_t attempts to manage boot_t entries

zpytela:


wrabcak:
@zpytela, 
What about other user domains?

zpytela:
staff_u has it allowed, guest_u is not allowed to execute systemctl.

wrabcak:
@zpytela I don't understand, you said staff_u have it allowed but you're don't auditing it in the PR.?

zpytela:
I meant sysadm_u, sorry for the typo.

wrabcak:
Okay good, now it make sense.

PR #376, opened 2020-06-11
Allow libkrb5 lib read client keytabs

zpytela:
Allow ssh processes with the ssh_t type to search directories in the
/var/kerberos/krb5 directory labeled krb5_keytab_t and read files with
the same type in subdirectories.


wrabcak:
LGTM.

Issue #217, opened 2018-07-13
Systemd-modules-load cannot load kernel modules with 'install' command in /etc/modprobe.d


ghost:
How to reproduce:

```
echo 'install dummy echo nope' > /etc/modprobe.d/test-install-selinux.conf
echo 'dummy' > /etc/modules-load.d/test-install-selinux.conf
systemctl restart systemd-modules-load
rm  /etc/modprobe.d/test-install-selinux.conf
rm /etc/modules-load.d/test-install-selinux.conf
ausearch -i -m avc --just-one
----
type=PROCTITLE msg=audit(13.07.2018 15:22:14.554:212) : proctitle=/usr/lib/systemd/systemd-modules-load 
type=SYSCALL msg=audit(13.07.2018 15:22:14.554:212) : arch=x86_64 syscall=execve success=no exit=EACCES(Permission denied) a0=0x7f3ec8ddd897 a1=0x7ffe0608ff10 a2=0x563ff1c6ffe0 a3=0x8 items=0 ppid=977 pid=978 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=systemd-modules exe=/usr/lib/systemd/systemd-modules-load subj=system_u:system_r:systemd_modules_load_t:s0 key=(null) 
type=AVC msg=audit(13.07.2018 15:22:14.554:212) : avc:  denied  { execute } for  pid=978 comm=systemd-modules name=bash dev="vda3" ino=4194473 scontext=system_u:system_r:systemd_modules_load_t:s0 tcontext=system_u:object_r:shell_exec_t:s0 tclass=file permissive=0 
```

I have found this while trying to load `ip_vs` in Fedora Atomic Host (wanted to try out IPVS in Kubernetes on FAH). `/etc/modprobe.d/firewalld-sysctls.conf` has this:

```sh
install nf_conntrack /sbin/modprobe --ignore-install nf_conntrack $CMDLINE_OPTS && /sbin/sysctl --quiet --pattern 'net[.]netfilter[.]nf_conntrack.*' --system
```

Which results in a whole bunch of denials for `nf_conntrack` in permissive mode:

```
----
type=PROCTITLE msg=audit(13.07.2018 15:35:52.284:80) : proctitle=/sbin/modprobe --ignore-install nf_conntrack 
[...]

PR #378, opened 2020-06-18
Allow systemd-modules to load kernel modules

zpytela:
Allow systemd-modules to load kernel modules at boot time
according to settings in /etc/modules-load.d.

This includes the following permission groups:
- execute generic programs in bin directories in the caller domain,
- execute shells in the caller domain,
- execute insmod in the caller domain.


wrabcak:
systemd_modules_load_t is quite powerful domain, why it needs execute shell and execute binary files ?

zpytela:
Firewalld service needs to reload a kernel module and reset kernel tunables.

Executing shell_exec_t is required for fork; kmod_exec_t for modprobe; bin_t for sysctl.
See also
https://bugzilla.redhat.com/show_bug.cgi?id=1823246#c3
https://github.com/fedora-selinux/selinux-policy/issues/217

PR #380, opened 2020-06-22
Modify systemd_delete_private_tmp() to use delete_*_pattern macros

zpytela:
Modify the systemd_delete_private_tmp() interface to use delete_*_pattern
macros instead of simple allow rules to allow all permissions related to
removing files.


wrabcak:
LGTM

PR #390, opened 2020-07-03
[WIP] Allow userdomain transition to cups named home content

zpytela:
Allow the userdomain attribute file transition to cups named content
in user's home and root's home directories.


zpytela:
Requires https://github.com/fedora-selinux/selinux-policy-contrib/pull/294
be merged first.

PR #244, opened 2019-01-22
kernel/files.fc: Label /var/run/motd\.d(./*)? and /var/run/motd as pam_var_run_t

rfairley:
This fixes changes in #230 and #232 which conflicted due to the
/var/run and /run equivalency, causing the etc_t label not to be
applied.

Fixes: #242

PR #398, opened 2020-07-15
Allow login programs to read MOTD files and symlinks, and label /var/run/zincati/public/motd.d/*

rfairley:
The first commit is a more general fix - login programs should be able to read symlinks out of `/run/motd.d/` e.g. `/run/motd.d/my.motd -> /run/path/to/my.motd`, but currently `sshd` gets an AVC denial (trying on Fedora 32) on dereferencing the symlink. Since this requires extending the allowed rules for the type used for the runtime MOTD file/directory, I think this warrants a new type, `motd_var_run_t` (which had been suggested in https://github.com/fedora-selinux/selinux-policy/pull/244).

The second commit uses the new label, and will enable functionality in https://github.com/coreos/zincati/pull/276.

Please see commit messages for further details.

Will require backporting this to F32, if accepted.

### DIFF
--- a/policy/modules/system/authlogin.fc
+++ b/policy/modules/system/authlogin.fc
@@ -74,8 +74,8 @@ ifdef(`distro_gentoo', `
 
 /var/run/console(/.*)?	 	gen_context(system_u:object_r:pam_var_console_t,s0)
 /var/run/faillock(/.*)?		gen_context(system_u:object_r:faillog_t,s0)
-/var/run/motd		--	gen_context(system_u:object_r:pam_var_run_t,s0)
-/var/run/motd\.d(/.*)?		gen_context(system_u:object_r:pam_var_run_t,s0)
+/var/run/motd		--	gen_context(system_u:object_r:motd_var_run_t,s0)
+/var/run/motd\.d(/.*)?		gen_context(system_u:object_r:motd_var_run_t,s0)
 /var/run/pam_mount(/.*)?	gen_context(system_u:object_r:pam_var_run_t,s0)
 /var/run/pam_ssh(/.*)?		gen_context(system_u:object_r:pam_var_run_t,s0)
 /var/run/pam_timestamp(/.*)?	gen_context(system_u:object_r:pam_var_run_t,s0)
--- a/policy/modules/system/authlogin.fc
+++ b/policy/modules/system/authlogin.fc
@@ -83,3 +83,7 @@ ifdef(`distro_gentoo', `
 /var/run/sudo(/.*)?		gen_context(system_u:object_r:pam_var_run_t,s0)
 /var/(db|adm)/sudo(/.*)?	gen_context(system_u:object_r:pam_var_run_t,s0)
 /var/lib/sudo(/.*)?	gen_context(system_u:object_r:pam_var_run_t,s0)
+
+# Allow services not running as root to write MOTD messages via symlink
+# out of /run/motd.d/. https://github.com/coreos/zincati/pull/276
+/var/run/zincati/public/motd\.d(/.*)?	gen_context(system_u:object_r:motd_var_run_t,s0)
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -77,6 +77,11 @@ files_pid_file(pam_var_console_t)
 type pam_var_run_t;
 files_pid_file(pam_var_run_t)
 
+# For MOTD files read by login programs, e.g. sshd, located in
+# /run/motd.d/, or symlinked out from /run/motd.d/
+type motd_var_run_t;
+files_type(motd_var_run_t);
+
 type shadow_t;
 files_auth_file(shadow_t)
 neverallow ~can_read_shadow_passwords shadow_t:file read;
@@ -640,6 +645,10 @@ selinux_compute_user_contexts(login_pgm)
 
 auth_manage_faillog(login_pgm)
 auth_manage_pam_pid(login_pgm)
+
+list_dirs_pattern(login_pgm, motd_var_run_t, motd_var_run_t);
+read_files_pattern(login_pgm, motd_var_run_t, motd_var_run_t);
+read_lnk_files_pattern(login_pgm, motd_var_run_t, motd_var_run_t);
 
 init_rw_utmp(login_pgm)
 


lucab:
Gentle ping for review/feedback on this.

/cc @zpytela

zpytela:
Hi @rfairley,
Thank you for submitting this commit, I am sorry for the delay in replying. It looks generally good, just a few notes:
- I miss a transition for the `motd.d` directory.
- It kind-of conflicts https://github.com/fedora-selinux/selinux-policy/blob/0b12468476d3e6a96996b81d1d9a5706b5aad90f/policy/modules/kernel/files.fc#L314-L317
I know this is not a result of your commit, but would be nice to have it consistent.
- I am not familiar with coreos/zincati; just checking if the `/run/zincati/public/motd.d` directory is managed properly, including setting the SELinux context.

lucab:
> I am not familiar with coreos/zincati; just checking if the /run/zincati/public/motd.d directory is managed properly, including setting the SELinux context.

I'm not sure if this answer your doubt, but the plan in the referenced PR is to have this directory created at runtime by `systemd-tmpfiles` on each boot, inheriting the proper SELinux context that way: https://github.com/coreos/zincati/pull/276/files#diff-bf95b05744b97a55c6a5b6acd3e392db.

zpytela:
Thanks @lucab,

The problem is that with no explicit setting of the context, the directories just inherit it from the parent. As a result, sshd will not be able to read the link target:

```
# matchpathcon /run/zincati /run/zincati/public
/run/zincati    system_u:object_r:var_run_t:s0
/run/zincati/public     system_u:object_r:var_run_t:s0
# sesearch -A -s login_pgm -t var_run_t -c file
```
I cannot see anything in the PR, so my doubts were not dispelled ;)

lucab:
The link target (shown in the same tmpfiles fragment above) is going to be at `/run/zincati/public/motd.d/deadend.motd`, thus `/run/motd.d/85-zincati-deadend.motd -> /run/zincati/public/motd.d/deadend.motd`.
The expectation is that the policy change in this PR for `/var/run/zincati/public/motd\.d(/.*)?` is going to cover both the parent directory and the target child.

I'm still not super-familiar with SELinux rules though. Is there something that was specifically concerning you regarding `public/` (as opposed to `public/motd.d/`) or am I misunderstanding how this works?

zpytela:
The rule nr 1 is that when a new object is created, it inherits SELinux context from its parent. If this is not desirable, a transition needs to exist in the policy, or it can be changed with a command afterwards. The fc file just applies when restorecon is run.

So what I was asking was if the context is handled manually for `/run/zincati/public/motd.d/`, either using chcon or restorecon. As I understand it, it is not, so the directory will be created with the type inherited from `/run/zincati/public/`.

There are 2 proposed solutions: include restorecon in a script or set up a transition, in a similar way as it is needed for `/run/motd.d` (it may even be covered by the same rule).

rfairley:
Thank you @zpytela for taking a look!

> The rule nr 1 is that when a new object is created, it inherits SELinux context from its parent. If this is not desirable, a transition needs to exist in the policy, or it can be changed with a command afterwards. The fc file just applies when restorecon is run.
> 
> So what I was asking was if the context is handled manually for /run/zincati/public/motd.d/, either using chcon or restorecon. As I understand it, it is not, so the directory will be created with the type inherited from /run/zincati/public/.
> 
> There are 2 proposed solutions: include restorecon in a script or set up a transition, in a similar way as it is needed for /run/motd.d (it may even be covered by the same rule).

Please keep in mind, this was tested and sshd can read the file (the link target) that Zincati drops in `/run/zincati/public/motd.d/` (https://github.com/coreos/zincati/pull/276#issuecomment-658627380), since the `/run/zincati/public/motd.d` directory has the label `pam_motd_var_run_t` applied. The way of setting the context for the Zincati motd.d directory, `/var/run/zincati/public/motd\.d(/.*)?	gen_context(system_u:object_r:pam_motd_var_run_t,s0)` and creating `/run/zincati/public/motd.d` through `tmpfiles.d` is the same method as the current way of setting the context of `/run/motd.d`;  we have `/var/run/motd\.d(/.*)?		gen_context(system_u:object_r:pam_var_run_t,s0)`, and `/run/motd.d` is created through `tmpfiles.d` (https://src.fedoraproject.org/rpms/setup/blob/5a7ff0dc1fdfb24d1ea9afe3be0bd25d03314e1a/f/setup.spec#_55). PAM and sshd do not explicitly call `restorecon`, as far as I know, to set the context of `/run/motd.d`.

I'm not sure when exactly in the startup process the SELinux labels are applied, but it seems it is happening late enough that the directories `/run/motd.d` and `/run/zincati/public/motd.d` are getting the correct contexts with these policy changes (i.e. sometime after `systemd-tmpfiles` has created the directories). Is it required that sshd or Zincati manually restore the labels, or can it be relied on that the correct labels are applied to directories created by `systemd-tmpfiles` (like this PR relies on)?

dustymabe:
Thanks for chiming in @rfairley.

sohankunkerkar:
> Please keep in mind, this was tested and sshd can read the file (the link target) that Zincati drops in `/run/zincati/public/motd.d/` ([coreos/zincati#276 (comment)](https://github.com/coreos/zincati/pull/276#issuecomment-658627380)), since the `/run/zincati/public/motd.d` directory has the label `pam_motd_var_run_t` applied. The way of setting the context for the Zincati motd.d directory, `/var/run/zincati/public/motd\.d(/.*)? gen_context(system_u:object_r:pam_motd_var_run_t,s0)` and creating `/run/zincati/public/motd.d` through `tmpfiles.d` is the same method as the current way of setting the context of `/run/motd.d`; we have `/var/run/motd\.d(/.*)? gen_context(system_u:object_r:pam_var_run_t,s0)`, and `/run/motd.d` is created through `tmpfiles.d` (https://src.fedoraproject.org/rpms/setup/blob/5a7ff0dc1fdfb24d1ea9afe3be0bd25d03314e1a/f/setup.spec#_55). PAM and sshd do not explicitly call `restorecon`, as far as I know, to set the context of `/run/motd.d`.

I just echo what @rfairley said here. I used [these instructions](https://github.com/coreos/zincati/pull/276#issuecomment-658627380) to verify this change and it looks fine to me. Here's what I see in an FCOS VM with the new policy in:

```shell
# ls -laZ /run/zincati/public/motd.d
total 4
drwxrwxr-x. 2 zincati zincati system_u:object_r:pam_motd_var_run_t:s0  60 Sep 16 14:51 .
drwxrwxr-x. 3 zincati zincati system_u:object_r:var_run_t:s0           80 Sep 16 14:46 ..
-rw-rw-r--. 1 zincati zincati system_u:object_r:pam_motd_var_run_t:s0 109 Sep 16 14:51 deadend.motd

# ls -laZ /run/zincati/**
/run/zincati/config.d:
total 0
drwxrwxr-x. 2 zincati zincati system_u:object_r:var_run_t:s0  40 Sep 16 15:14 .
drwxrwxr-x. 5 zincati zincati system_u:object_r:var_run_t:s0 100 Sep 16 15:14 ..
/run/zincati/private:
total 0
drwxrwx---. 2 zincati zincati system_u:object_r:var_run_t:s0  60 Sep 16 15:14 .
drwxrwxr-x. 5 zincati zincati system_u:object_r:var_run_t:s0 100 Sep 16 15:14 ..
lrwxrwxrwx. 1 root    root    system_u:object_r:var_run_t:s0  26 Sep 16 15:14 metrics.promsock -> ../public/metrics.promsock
/run/zincati/public:
total 0
drwxrwxr-x. 3 zincati zincati system_u:object_r:var_run_t:s0           80 Sep 16 15:14 .
drwxrwxr-x. 5 zincati zincati system_u:object_r:var_run_t:s0          100 Sep 16 15:14 ..
srwxr-xr-x. 1 zincati zincati system_u:object_r:var_run_t:s0            0 Sep 16 15:14 metrics.promsock
drwxrwxr-x. 2 zincati zincati system_u:object_r:pam_motd_var_run_t:s0  60 Sep 16 15:25 motd.d

# ls -laZ /run/motd.d/** 
-rw-------. 1 root root system_u:object_r:pam_motd_var_run_t:s0 44 Sep 16 15:14 /run/motd.d/40_console-login-helper-messages.motd
lrwxrwxrwx. 1 root root system_u:object_r:pam_motd_var_run_t:s0 37 Sep 16 15:14 /run/motd.d/85-zincati-deadend.motd -> ../zincati/public/motd.d/deadend.motd

```

zpytela:
Hi,

If it works without a transition for the motd.d directory, I suppose it's the tmpfiles "z/Z" line type.

Have you checked all login methods and also for all selinux users if motd is displayed properly after the change?

I also wonder why is it pam_motd_var_run_t and not just motd_var_run_t - AFAIK only sshd uses the pam module.

zpytela:
To create the most common confined users, use a snippet like this:
```
for user in user staff sysadm; do adduser -Z ${user}_u $user; echo "${user}01" | passwd --stdin $user; done
```
and then log in as user, staff, sysadm.

sohankunkerkar:
> Hi,
> 
> If it works without a transition for the motd.d directory, I suppose it's the tmpfiles "z/Z" line type.
> 
> Have you checked all login methods and also for all selinux users if motd is displayed properly after the change?

AFAIK, motd is printed on SSH and serial console, but not on sudo, so I tried verifying this change for the `SSH` and `serial console` login methods, here's my observation:
1. For SSH, I verified [this](https://github.com/fedora-selinux/selinux-policy/pull/398) change for all selinux users using [this method](https://github.com/fedora-selinux/selinux-policy/pull/398#issuecomment-694906639) and it looks fine to me.
2. For serial console, there's one caveat but that's not going to impact this PR as such. In this case, motd doesn't show up on the serial console login after a successful login; it looks like login invoked by `agetty --autologin` only supports `/etc/motd` by default. After doing some investigation, it seems like [this patch](https://github.com/karelzak/util-linux/commit/5a528e2c6ff9735266fc2607c359e925b074bf2c) will allow `agetty` to read information from `/run/motd.d` (which is available in the latest release of `util-linux`(v2.36) targeted for Fedora-33, and FCOS is currently using `Fedora-32` steam), so that justifies the issue.


> I also wonder why is it pam_motd_var_run_t and not just motd_var_run_t - AFAIK only sshd uses the pam module.

I would point you to [this commit message](https://github.com/fedora-selinux/selinux-policy/pull/398/commits/07cbc7bc99477b0e05dc6f585af213474ffe403f) where @rfairley explained the exact reasoning.

zpytela:
The content looks good. I don't want to nitpick, it's true it is just a literal string, but still think it may become confusing a bit: @rfairley, is there a reason for the "pam" in `pam_motd_var_run_t`?

sshd uses pam_motd, but login reads motd directly, as well as the old r-services, probably no other current service uses the motd concept.

rfairley:
Apologies for the late response. `motd_var_run_t` would be better - any login program is meant to read these directories. I had kept the `pam_` prefix in only due to following naming of the previous type (`pam_var_run_t`) and the suggestion at https://github.com/fedora-selinux/selinux-policy/pull/244#issuecomment-457356237, but there is no reason to have it. Updated to remove the `pam_` prefix.

Thank you @sohankunkerkar for verifying these changes against the tests @zpytela suggested and helping see this PR through.

> If it works without a transition for the motd.d directory, I suppose it's the tmpfiles "z/Z" line type.

I'm not noticing use of `z`/`Z` for `/run/motd.d/` (or `/run`) in `systemd-tmpfiles --cat-config` in my current Fedora system (nor is it done in https://github.com/coreos/zincati/blob/95509522dccad10ef317cb82d5ff73f3198fe16d/dist/tmpfiles.d/zincati.conf#L20-L21), maybe somewhere else is restoring contexts for runtime files after the runtime files are created?

zpytela:
> Apologies for the late response. `motd_var_run_t` would be better - any login program is meant to read these directories. I had kept the `pam_` prefix in only due to following naming of the previous type (`pam_var_run_t`) and the suggestion at [#244 (comment)](https://github.com/fedora-selinux/selinux-policy/pull/244#issuecomment-457356237), but there is no reason to have it. Updated to remove the `pam_` prefix.
> 
> Thank you @sohankunkerkar for verifying these changes against the tests @zpytela suggested and helping see this PR through.
> 
> > If it works without a transition for the motd.d directory, I suppose it's the tmpfiles "z/Z" line type.
> 
> I'm not noticing use of `z`/`Z` for `/run/motd.d/` (or `/run`) in `systemd-tmpfiles --cat-config` in my current Fedora system (nor is it done in https://github.com/coreos/zincati/blob/95509522dccad10ef317cb82d5ff73f3198fe16d/dist/tmpfiles.d/zincati.conf#L20-L21), maybe somewhere else is restoring contexts for runtime files after the runtime files are created?
I am not that familiar with tmpfiles usage, so no need for a change as long as it works.

zpytela:
LGTM now, merging, thank you.

PR #402, opened 2020-08-01
Remove all the kdbus stuff

WOnder93:
KDBus is a dead project that never made it to the mainline kernel (see references in the first commit's log message), so let's stop carrying this utterly useless chunk of code in the policy.


wrabcak:
LGTM,
@zpytela, if we merge this PR we need to remove kdbus module also in disgit.

zpytela:
@wrabcak I can do it right away and create a new build with just this change to isolate it, I just want to go through this commits set first.

wrabcak:
@zpytela so can I merge it?

zpytela:
@wrabcak go ahead.

PR #405, opened 2020-08-11
Allow systemd-machined create userdbd runtime sock files

zpytela:
Create the systemd_create_userdbd_runtime_sock_files() interface.

Resolves: rhbz#1862686


wrabcak:
LGTM

PR #407, opened 2020-08-12
Allow nsswitch_domain to connect to systemd-machined using a unix socket

zpytela:
Create the systemd_machined_stream_connect() interface.

Resolves: rhbz#1865748


wrabcak:
LGTM.

PR #413, opened 2020-08-20
Allow syslogd_t domain to read/write tmpfs systemd-bootchart files

Richard-Filo:
Create the two interfaces to allow mapping and r/w permisions.
Add this two interfaces to the policy for domain syslogd_t.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1838163

The one way how can the systemd-journald get a log data from any services is by socket /run/systemd/journal/socket. But when the message is bigger than max size of datagram, it must be done differently. It is by filedescriptor, which is connected to the datagram and in the file to which the file descriptor refers are the log data that were not sent. The file is created by memfd_create() syscall and in kernel the file is implemented as tmpfs.

That means any service can communicate in this way.

### DIFF
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -720,6 +720,11 @@ optional_policy(`
 ')
 
 optional_policy(`
+	systemd_rw_bootchart_tmpfs_files(syslogd_t)
+	systemd_map_bootchart_tmpfs_files(syslogd_t)
+')
+
+optional_policy(`
     daemontools_search_svc_dir(syslogd_t)
 ')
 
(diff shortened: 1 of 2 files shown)


Richard-Filo:
```
# sesearch -A -s syslogd_t -t systemd_bootchart_tmpfs_t -c file
```
after installation of scratch build:
```
# sesearch -A -s syslogd_t -t systemd_bootchart_tmpfs_t -c file
allow syslogd_t systemd_bootchart_tmpfs_t:file { append getattr ioctl lock map open read write };
```

wrabcak:
LGTM

PR #416, opened 2020-08-25
Change transitions for ~/.config/Yubico

zpytela:
Created the auth_filetrans_auth_home_content() interface which is used
to allow the filename transition in gnome config directory for the
login_pgm and userdomain attributes.

This commit reverts the transitions introduced in
commit 1363710b88904f29915e39335fef0dfb673a0f70.


zpytela:
Requesting a proper review. This commit and the previous need to be backported to all supported Fedoras.

A scratchbuild is available, transition seems to work for unconfined_u.

PR #418, opened 2020-08-27
Add file context for ~/.config/Yubico

zpytela:


zpytela:
CI fails for dev_lock_all_blk_files() interface used, but not defined. This if exists in f33+ only. The system for CI is being built on a f33 system. My scratchbuild seems to work properly. A similar backport for F31 passed CI, so I will merge this PR unless an objection appears soon.

wrabcak:
agree. please merge

PR #421, opened 2020-08-27
Allow unconfined_t to node_bind icmp_sockets in node_t domain

zpytela:
1. 

When uncofined user run ping or traceroute, this process get label unconfined_t.
Allow to ping or traceroute, which run as unconfined_t, to node_bind icmp_sockets in node_t domain.


zpytela:
Fails again in vlock.te due to a mix of f33 and older sources used.

PR #423, opened 2020-08-28
Allow strongswan chown, setgid, setuid capabilities

zpytela:
The setuid/setgid caps are required to change the user the daemon
switches to after startup when charon.user/group key was set.
The chown cap is required to chown /run/strongswan.

### DIFF
--- a/policy/modules/system/ipsec.te
+++ b/policy/modules/system/ipsec.te
@@ -252,7 +252,7 @@ optional_policy(`
 # ipsec_mgmt Local policy
 #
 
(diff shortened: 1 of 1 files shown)


DemiMarie:
@zpytela Turns out that the real problem is with the label on `/usr/sbin/charon-systemd`.  It should be `ipsec_exec_t`.

DemiMarie:
The other problem is that the Fedora package is not built with capability support, so it can’t drop privileges.

zpytela:
Closing in favor of https://github.com/fedora-selinux/selinux-policy/pull/424

PR #424, opened 2020-08-28
Relabel /usr/sbin/charon-systemd as ipsec_exec_t

DemiMarie:
This causes StrongSwan to be run with the `ipsec_exec_t` context, which allows it to bind to its sockets.


zpytela:
The commit is fine; could you just run some scenarios to verify everything works as expected? There were probably many permissions granted to ipsec_mgmt_t based on the previous setting.

DemiMarie:
@zpytela I use this every day at home.

zpytela:
Merging to rawhide. @DemiMarie please let me know if a backport is required, too.

DemiMarie:
@zpytela a backport would be greatly appreciated, thanks!

zpytela:
F33 PR: https://github.com/fedora-selinux/selinux-policy/pull/446

PR #425, opened 2020-09-02
kernel/filesystem: Add exfat support (no extended attributes)

travier:
According to [1], the exfat filesystem does not support extended
attributes so it should be handled like fat/vfat/ntfs.

[1] https://en.wikipedia.org/wiki/Comparison_of_file_systems


travier:
From https://discussion.fedoraproject.org/t/selinux-denials-for-systemd-processes/22921

wrabcak:
@WOnder93 PTAL

WOnder93 (approved):
LGTM

PR #426, opened 2020-09-03
 Confine systemd-sleep service

wrabcak:
Create new SELinux security policy for systemd-sleep service to allow hibarnate system to swapfile in enforcing mode.

@zpytela  we want to introduce this change only in fedora-rawhide (not F33.)

### DIFF
--- a/policy/modules/system/fstools.if
+++ b/policy/modules/system/fstools.if
@@ -157,6 +157,24 @@ interface(`fstools_getattr_swap_files',`
 
 ########################################
 ## <summary>
+##	Read/Write swapfile
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fstools_rw_swap_files',`
+	gen_require(`
+		type swapfile_t;
+	')
+
+	allow $1 swapfile_t:file rw_file_perms;
+')
+
+########################################
+## <summary>
 ##	Create, read, write, and delete the FSADM pid files.
 ## </summary>
 ## <param name="domain">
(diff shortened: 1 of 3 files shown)


zpytela:
I think this can work as an initial policy for the sleep-related services which can be further tested.

What I am not comfortable with, although not related to this PR, is that swapfile_t is a part of 3 generic attributes (file_type, non_auth_file_type, non_security_file_type) meaning that 32 domains have read access to it and 11 domains write access.

wrabcak:
> I think this can work as an initial policy for the sleep-related services which can be further tested.
> 

It was tested by sytemd developers and there are no more systemd-sleep related SELinux denials. 

> What I am not comfortable with, although not related to this PR, is that swapfile_t is a part of 3 generic attributes (file_type, non_auth_file_type, non_security_file_type) meaning that 32 domains have read access to it and 11 domains write access.

I don't understand what you mean by this? We have multiple similar issues in the policy.  This PR is not granting access to any of these attributes, so maybe you can create BZ for the issue but as you mentioned, it's unrelated.

MartinNowak:
Could this have broken suspend-then-hibernate in F34?
```
type=AVC msg=audit(1617712931.069:1172): avc:  denied  { wake_alarm } for  pid=72095 comm="systemd-sleep" capability=35  scontext=system_u:system_r:systemd_sleep_t:s0 tcontext=system_u:system_r:systemd_sleep_t:s0 tclass=capability2 permissive=0
```

zpytela:
@MartinNowak could you open a new github issue?

MartinNowak:
> @MartinNowak could you open a new github issue?

https://github.com/fedora-selinux/selinux-policy/pull/680

PR #429, opened 2020-09-04
Allow gnome-initial-setup execute in a xdm sandbox

zpytela:
Add permissions required when /usr/libexec/gnome-initial-setup is
executed during initial setup, that is still from xdm, in a sandbox
in order to access an existing online account from the linux user's
environment.

Resolves: rhbz#1870476


wrabcak:
LGTM

PR #430, opened 2020-09-10
Add to map non_security_files to the userdom_admin_user_template temp…

Richard-Filo:
…late

If domain can manage non_security_files, from perspective of security it is correct to allow mapping of the files.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1873964

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -1685,10 +1685,12 @@ template(`userdom_admin_user_template',`
 	# Manage almost all files
 	files_manage_non_security_dirs($1_t)
 	files_manage_non_security_files($1_t)
+	# Map almost all files
+	files_map_non_security_files($1_t)
 	# Relabel almost all files
 	files_relabel_non_security_files($1_t)
 
-    files_mounton_rootfs($1_t)
+	files_mounton_rootfs($1_t)
 
 	init_telinit($1_t)
 


Richard-Filo:
```
# sesearch -A  -t init_var_lib_t -c file -s sysadm_t 
allow sysadm_t file_type:file { getattr relabelfrom relabelto };
allow sysadm_t non_security_file_type:file { append create getattr ioctl link lock open read relabelfrom relabelto rename setattr unlink write };
```
after installation of scratch build:
```
# sesearch -A  -t init_var_lib_t -c file -s sysadm_t 
allow sysadm_t file_type:file { getattr relabelfrom relabelto };
allow sysadm_t non_security_file_type:file { append create ioctl link lock map open read rename setattr unlink write };
```

zpytela:
It's interesting that "getattr relabelfrom relabelto" are no longer a part of the output. Have you made some investigation why this had happened?
I suppose it is harmless because of the more generic ruleset for file_type, but still.

WOnder93:
Interesting, that could be a bug (or rather an imperfection) in libsepol's policy optimization (which removes redundant rules / permissions from rules in the final policy). I'm taking a note to look at this tomorrow.

WOnder93:
I can't see the same reference output on F3 nor Rawhide, only on F31, where the policy optimization is not available yet. @Richard-Filo Did you get the first output from a different system than where you tested the scratch build?

Richard-Filo:
I tested on Fedora 33.

Richard-Filo:
Sorry, that was my bad. Correct output before the scratch build:
```
# sesearch -A  -t init_var_lib_t -c file -s sysadm_t
allow sysadm_t file_type:file { getattr relabelfrom relabelto };
allow sysadm_t non_security_file_type:file { append create ioctl link lock open read rename setattr unlink write };

```

zpytela:
@Richard-Filo in other words the only difference in the output is the map permission added?

@WOnder93 really looks like some optimization as namely getattr is almost always present in macros and patterns:
```
policy/support/obj_perm_sets.spt
define(`manage_file_perms',`{ create open getattr setattr read write append rename link unlink ioctl lock }')
```

Richard-Filo:
Yes, it is.

zpytela:
OK, I'll merge it to rawhide once f33 branch is created.

zpytela:
Merging; this commit goes to rawhide only.

PR #431, opened 2020-09-16
Add 2 addl. interfaces to lvm

tasleson:
We need interfaces to allow applications the ability to:
* Update the lvm hints file
* Send dbus messages

This is needed for targetd storage daemon.

Signed-off-by: Tony Asleson <tasleson@redhat.com>

### DIFF
--- a/policy/modules/system/lvm.if
+++ b/policy/modules/system/lvm.if
@@ -452,4 +452,40 @@ interface(`lvm_manage_lock',`
 ')
 
 
+########################################
+## <summary>
+##	Allow dbus send for lvm dbus API (only send needed)
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`lvm_dbus_send_msg',`
+	gen_require(`
+		type lvm_t;
+		class dbus send_msg;
+	')
+    allow $1 lvm_t:dbus send_msg;
 
+')
+
+########################################
+## <summary>
+##	Allow lvm hints file access
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`lvm_rw_var_run',`
+	gen_require(`
+		type lvm_t;
+		type lvm_var_run_t;
+	')
+    allow $1 lvm_var_run_t:file { rw_file_perms };
+
+')


policy/modules/system/lvm.if
```diff
@@ -452,4 +452,41 @@ interface(`lvm_manage_lock',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`lvm_dbus_send_msg',`
+	gen_require(`
+		type lvm_t;
+		class dbus send_msg;
+	')
+    allow $1 lvm_t:dbus send_msg;
```
zpytela, line 470: The dbus communication usually needs to be allowed for both directions, it's not this case?

policy/modules/system/lvm.if
```diff
@@ -452,4 +452,41 @@ interface(`lvm_manage_lock',`
...
+## <summary>
+##	Allow lvm hints file access
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`lvm_hints_file_access',`
```
zpytela: To have it aligned with other interfaces names, it should rather be lvm_read_var_run

policy/modules/system/lvm.if
```diff
@@ -452,4 +452,41 @@ interface(`lvm_manage_lock',`
...
+	gen_require(`
+		type lvm_t;
+		type lvm_var_run_t;
+		class file { getattr read write open lock };
```
zpytela: Class and permissions do not need to be declared in policy.

policy/modules/system/lvm.if
```diff
@@ -452,4 +452,41 @@ interface(`lvm_manage_lock',`
...
+	')
+    allow $1 lvm_var_run_t:file { getattr read write open lock };
```
zpytela: it's better to use permissions sets, or even file patterns.

policy/modules/system/lvm.if
zpytela: As there also is the write permission, it should actually rather be rw than read.

zpytela (changes requested):

policy/modules/system/lvm.if
tasleson: In this specific case, we only need one direction as there is no reply.

policy/modules/system/lvm.if
tasleson: Done: converted to use permission sets.

policy/modules/system/lvm.if
tasleson: Done, removed.

policy/modules/system/lvm.if
tasleson: Done

policy/modules/system/lvm.if
zpytela: Note there are `rw_file_perms`; unfortunately, there is no `rw_files_pattern` which ensures the domain can access the parent directory.

zpytela:
Please adjust the commit message in this PR, too. The interfaces names will fit into the first line, and optionally the content of comment#1 can be in the body.

policy/modules/system/lvm.if
tasleson: Changed to `rw_file_perms`

tasleson:
> Please adjust the commit message in this PR, too. The interfaces names will fit into the first line, and optionally the content of comment#1 can be in the body.

Done

wrabcak:
LGTM.

zpytela:
Thanks, merging to rawhide, please let me know if backport to any other fedora is required.

tasleson:
> Thanks, merging to rawhide, please let me know if backport to any other fedora is required.

Please, merge these changes into F33 & F32, thanks!

zpytela:
Done, will merge them when ci finishes.
https://github.com/fedora-selinux/selinux-policy/pull/442
https://github.com/fedora-selinux/selinux-policy/pull/443

PR #433, opened 2020-09-17
kernel/filesystem: Add exfat support (no extended attributes)

zpytela:
According to [1], the exfat filesystem does not support extended
attributes so it should be handled like fat/vfat/ntfs.

[1] https://en.wikipedia.org/wiki/Comparison_of_file_systems

Resolves: rhbz#1874836

### DIFF
--- a/policy/modules/kernel/filesystem.te
+++ b/policy/modules/kernel/filesystem.te
@@ -268,7 +268,7 @@ genfscon cifs / gen_context(system_u:object_r:cifs_t,s0)
 genfscon smbfs / gen_context(system_u:object_r:cifs_t,s0)
 
 #
-# dosfs_t is the type for fat and vfat
+# dosfs_t is the type for fat, vfat and exfat
 # filesystems and their files.
 #
 type dosfs_t;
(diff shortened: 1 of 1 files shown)


zpytela:
It looks like the CI build uses f33/f34 packages. The error refers to the `dev_lock_all_blk_files` interface which neither exists in base/f32 nor is refered to in contrib/f32, unlike f33.

WOnder93:
We could fix that by checking out the matching contrib branch in Travis - but then we'd have to hardcode the branch into `.travis.yml` and remember to update it (`rawhide` -> `f<N>`) when branching off a new stable release. However, we could guard this by a check like (pseudocode):
```
if this-is-the-fedora-selinux-repo && [ "$TRAVIS_BRANCH" != "the-hardcoded-branch-name" ]; then FAIL; fi
```
@zpytela If you'd like that, I can send pull requests :)

zpytela:
@WOnder93, thank you for the proposal, if the only downside is updating travis.yml with each new branch, I like it.
Frankly what surprised me most was the inconsistency - reference to a non existing interface, while it should be either both (f33) or none (f32).

PR #436, opened 2020-09-18
[RFC] Check out the right -contrib branch in Travis

WOnder93:
Add instrumentation to allow checking out the right branch in Travis for
the -contrib repo. The branch name needs to be correctly set on all
branches when they are created (this is verified using a Travis check)
so that people get the right result when runnng CI on their own forks.

--
This should fix the issue seen in PR #433.
PRs with cherry-pick + TARGET_BRANCH update for active f* branches will follow after this one's acked/merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,25 @@ language: c
 services:
     - docker
 
+env:
+    # Expected name of target branch - this branch is fetched also from -contrib
+    # Remember to update it when creating a new release branch (CI will warn you
+    # in case you forget :)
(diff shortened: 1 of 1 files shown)


zpytela:
Merging to rawhide, thank you.

WOnder93:
I opened PRs for `f32 and `f33` (see above). I discovered that on the `f31` branch there was already `-b f31` added manually, so we don't need to do anything there.

zpytela:
Thanks, actually the f32 branch is the main one to resolve the problem in.

PR #437, opened 2020-09-21
Label 4460/tcp as ntske_port_t

mlichvar:
This is needed for chrony NTS-KE support.

https://bugzilla.redhat.com/show_bug.cgi?id=1872624


zpytela:
@mlichvar, thank you for your PRs. This one seems to be in a mid-air collision with https://github.com/fedora-selinux/selinux-policy/pull/438, so please rather review its commit message.

PR #438, opened 2020-09-21
Label 2460/tcp port as nts_ke_t

zpytela:
The NTS protocol is a security extension for time protocols like NTP or PTP.
It provides strong cryptographic protection against packet manipulation,
prevents tracking, scales, is robust against packet loss, and minimizes
the loss of accuracy due to the securing process.

The tcp port 2460 is used for NTS key establishment.


wrabcak:
LGTM. Please merge.

mlichvar:
Looks good to me and it has a better commit message than the #437.

zpytela:
Merging and backporting to F33.

PR #444, opened 2020-09-24
Allow domain write to an automount unnamed pipe

zpytela:
With the kernel commit 13c164b1a186 ("autofs: switch to kernel_write"),
an additional LSM permission check is done when a process tries to
access a directory on an autofs volume, which has not been mounted yet,
and it results in a write operation to the automount pipe.

This commit allows any domain write to the unnamed pipe kernel uses to
communicate with automount to service the directory access request and
should be considered a temporary workaround until a different
implementation in kernel is found.

Resolves: rhbz#1874338

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -570,6 +570,12 @@ optional_policy(`
 ')
 
 optional_policy(`
+	# A workaround to handle additional permissions check
+	# introduced as an involuntary result of a kernel change
+	automount_write_pipes(domain)
+')
+
+optional_policy(`
 	sosreport_append_tmp_files(domain)
 ')
 


zpytela:
@WOnder93, please review the commit message for accuracy.

WOnder93:
I would change: `check is needed` -> `check is done` (it's not necessarily needed, but done anyway...) and `in a write syscall` -> `in a write operation` (it is not a syscall per se, just a usual function call within the same syscall, which happens to do also the LSM check). And I would recommend also adding a short comment in the code, so it can be discovered easily during code inspection that it is only a workaround. (Otherwise it may easily fall through the cracks unless someone also looks at the commit history/git blame.)

Otherwise LGTM.

zpytela:
Fixed, thanks.

WOnder93:
I still can't see the comment in the code. Is it an oversight or you just didn't like the suggestion? :)

zpytela:
```
perl -e 'for (1..100) { print "I will not forget to git-add after changing a file.\n"; }'
```

WOnder93 (approved):
(I'd also add a link to the BZ into the comment for convenience, but it's possible to dig it out of git history, so I don't insist on it.)

raven-au:
> I would change: `check is needed` -> `check is done` (it's not necessarily needed, but done anyway...) and `in a write syscall` -> `in a write operation` (it is not a syscall per se, just a usual function call within the same syscall, which happens to do also the LSM check). And I would recommend also adding a short comment in the code, so it can be discovered easily during code inspection that it is only a workaround. (Otherwise it may easily fall through the cracks unless someone also looks at the commit history/git blame.)

@WOnder93 is correct, the callback to the automount daemon is done during kernel path resolution.
Mostly due to a "open like" system calls but "stat like" system calls can also result in a call back.
@WOnder93's description of it as kernel internal is most accurate.

zpytela:
Backporting:
F33: https://github.com/fedora-selinux/selinux-policy/pull/447
F32: https://github.com/fedora-selinux/selinux-policy/pull/448
F31: https://github.com/fedora-selinux/selinux-policy/pull/449

PR #450, opened 2020-09-27
Convert fc_sort.c to Python to get rid of dependency on gcc

WOnder93:
This backports commits from refpolicy that rewrite `fc_sort` to Pythin, which will allow us to drop the build-time dependency on gcc. That means slighly faster Travis CI and builds (esp. local mockbuilds).


WOnder93:
I realized that `git` can also be dropped, so I spliced it in as a separate commit.

zpytela:
LGTM provided the policy builds after the improvements.

I wonder if there is a noticeable performance penalty for the fc_sort change to python.

And good catch for the kernel/devices.fc entry.

WOnder93:
> LGTM provided the policy builds after the improvements.

I find your lack of faith disturbing GIF

It does: https://koji.fedoraproject.org/koji/taskinfo?taskID=52823625
(Notice in root.log that gcc is not installed.)

> I wonder if there is a noticeable performance penalty for the fc_sort change to python.

I measured it and in Travis it now takes ~60 ms instead of <10 ms. It's only run once during the whole build, so hardly a big deal.

> And good catch for the kernel/devices.fc entry.

It was actually caught by the new fc_sort :) (And also SELint, but I'm holding off on my SELint patches until we merge contrib into base.)

zpytela:
Thank you,

I will merge it later to have a dedicated build.

(And if faith means to think something is true without having proof, then I'll probably lack it again in the future.)

WOnder93:
> (And if faith means to think something is true without having proof, then I'll probably lack it again in the future.)

Yeah, I meant it only as a joke :)

PR #452, opened 2020-10-02
Allow systemd-logind read swapfiles

zpytela:


zpytela:
I will merge it later today if there are no further comments.

PR #455, opened 2020-10-13
Allow init dbus chat with kernel

zpytela:
With systemd as user and group names resolver, bidirectional
dbus communication between systemd and kernel is required.
    
This commit reverts previously introduced commit
f4b4b99846308b75fd9bb2f50d6e9897459f87a5 to dontaudit one direction
of the communication, resulting in dbus timeouts.
    
Resolves: rhbz#1694681

### DIFF
--- a/policy/modules/kernel/kernel.te
+++ b/policy/modules/kernel/kernel.te
@@ -392,9 +392,9 @@ optional_policy(`
 ')
 
 optional_policy(`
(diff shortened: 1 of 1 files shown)


WOnder93:
Why?

WOnder93:
That looks better, thanks! :+1:

PR #456, opened 2020-10-14
Allow initrc_t create /run/chronyd-dhcp directory with a transition

zpytela:
Chronyd is required to read preferred sources files stored in
/run/chronyd-dhcp to be able to get correct time settings
from the dhcp server and have them applied.

Resolves: rhbz#1880948

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -1210,6 +1210,10 @@ ifdef(`distro_redhat',`
 	')
 
 	optional_policy(`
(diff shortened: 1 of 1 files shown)


zpytela:
Requires a new if from https://github.com/fedora-selinux/selinux-policy-contrib/pull/344

zpytela:
Based on positive feedback with the scratchbuild, merging and backporting to F33.

PR #459, opened 2020-10-19
Revert "Removed adding to attribute unpriv_userdomain from userdom_un…

zpytela:
…priv_type template"

This reverts commit 6b048173767701ec7f37dbe302316bdfbb2f9b4d as the
change prevents unconfined users from logging in in GUI, being a result
of missing transition from xdm_t to unconfined_t.


zpytela:
@Koncpa, FYI reverting the commit for F33, just this one.

PR #461, opened 2020-10-21
Revert "Label /var/run/zincati/public/motd.d/* as motd_var_run_t"

travier:
This reverts commit 540ea957bd85105213929f442dd6df5f9255c1b7.

This change [1] was initially required to enable a future Zincati feature [2] that was eventually redesigned to work differently, with a lesser reliance on arbitrary symlinks [3].

The first change [4] from [1] is still valid and can be kept as is.

[1] https://github.com/fedora-selinux/selinux-policy/pull/398
[2] https://github.com/coreos/zincati/pull/276
[3] https://github.com/coreos/zincati/pull/367
[4] https://github.com/fedora-selinux/selinux-policy/commit/8b9c6a9df672d84d4f717a044da2314ca84eb3e9


travier:
This should be applied to all branches that received the original commit mentioned in #398.

CC @lucab, @sohankunkerkar

lucab:
LGTM

zpytela:
OK, merging. It looks like the PR made it only to rawhide, so no backport is required.

PR #462, opened 2020-10-22
Merge contrib

zpytela:


zpytela:
Do not merge, for testing purposes only.
This PR is to verify selinux-policy-contrib can be merged with selinux-policy.

PR #469, opened 2020-10-29
allow google authenticator tmpfile

duritong:
Google Authenticator started writing to a new tmpfile, which should
also be allowed.

See https://bugzilla.redhat.com/show_bug.cgi?id=1840113 &
https://github.com/google/google-authenticator-libpam/issues/101

### DIFF
--- a/policy/modules/system/authlogin.if
+++ b/policy/modules/system/authlogin.if
@@ -2311,7 +2311,7 @@ interface(`auth_filetrans_admin_home_content',`
 	')
 
 	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator")
-	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~")
+	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~.*")
 	userdom_admin_home_dir_filetrans($1, auth_home_t, dir, ".yubico")
 ')
 
@@ -2375,7 +2375,7 @@ interface(`auth_filetrans_home_content',`
 	')
 
 	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator")
-	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~")
+	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~.*")
 	userdom_user_home_dir_filetrans($1, auth_home_t, dir, ".yubico")
 ')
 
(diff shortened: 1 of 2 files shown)


duritong:
This must be backported to the various releases

policy/modules/system/authlogin.if
```diff
@@ -2311,7 +2311,7 @@ interface(`auth_filetrans_admin_home_content',`
 	')
 
 	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator")
-	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~")
+	userdom_admin_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~.*")
```
zpytela, line 2314: While regexps can be used in file context specifications, they are not allowed in file transitions. It would require to change the location to a directory with a fixed name.

policy/modules/system/authlogin.if
```diff
@@ -2375,7 +2375,7 @@ interface(`auth_filetrans_home_content',`
 	')
 
 	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator")
-	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~")
+	userdom_user_home_dir_filetrans($1, auth_home_t, file, ".google_authenticator~.*")
```
zpytela, line 2378: Same here.

zpytela (changes requested):

zpytela:
@Koncpa FYI.

policy/modules/system/authlogin.if
duritong: Does this mean, that it would be better that google-authenticator would use a directory, like ~/.ga/ for all configs and then we would add this one?

I can also change the package to move to that location (if not configured differently), BUT I guess we still need to orchestrate that with the selinux policy packages.

WOnder93: Yes, if you can actually influence the location, then putting all of the files in a common directory would be the best solution.

And if you're going to change the location, then I'd suggest to move it to `~/.config/google-authenticator` so that there is one less application cluttering the root of the home directory :) (But that's just a general recommendation, not related to SELinux policy.)

See also:
https://unix.stackexchange.com/q/24347
https://unix.stackexchange.com/a/581687

q2dg:
Please...

Koncpa:
Hi @q2dg ,

we still wait for resolving this issue: https://github.com/google/google-authenticator-libpam/issues/101

EarlRamirez:
Updating .google_authenticator label from auth_home_t to ssh_home_t resolved this issue for me

m0canu1:
I'm getting into this because the permissions set to `.google_authenticator` file and its temporary "twins" should also be allowed to **radiusd**.

PR #471, opened 2020-11-03
Allow user domains transition to rpmdb_t

zpytela:
Allow unconfined_t and sysadm_t user domains transition to rpmdb_t
on executing the rpmdb command.


zpytela:
Requires an interface from https://github.com/fedora-selinux/selinux-policy-contrib/pull/358

zpytela:
Merging as the commit was proved to work expectedly when the interface exists.

PR #472, opened 2020-11-06
Set default file context for /var/lib/ipsec/nss

zpytela:
The NSS database used by libreswan has moved from /etc/ipsec.d
to /var/lib/ipsec/nss since libreswan 4.0. The default file context
specification needs to be adjusted accordingly in selinux-policy.

Resolves: rhbz#1883666


zpytela:
Merging per feedback in the bz.

PR #476, opened 2020-11-11
Label /etc/resolv.conf as net_conf_t only if it is a plain file

zpytela:
With systemd-resolved in place, /etc/resolv.conf is a symlink to
../run/systemd/resolve/stub-resolv.conf. Some domains allowed access to
the net_conf_t type class file do not have access to the same type class
lnk_file, while all domains have access to etc_t type class lnk_file.


zpytela:
A build created, no apparent problem noticed.

PR #477, opened 2020-11-11
Allow syslog append to public_content_rw_t

zpytela:


zpytela:
Verified using a policy scratch build.

PR #478, opened 2020-11-19
virtiofs supports Xattrs and SELinux

rhatdan:

### DIFF
--- a/policy/modules/kernel/filesystem.te
+++ b/policy/modules/kernel/filesystem.te
@@ -45,6 +45,7 @@ fs_use_xattr vxfs gen_context(system_u:object_r:fs_t,s0);
 fs_use_xattr odms gen_context(system_u:object_r:fs_t,s0);
 fs_use_xattr vxclonefs gen_context(system_u:object_r:fs_t,s0);
 fs_use_xattr ceph gen_context(system_u:object_r:fs_t,s0);
+fs_use_xattr virtiofs gen_context(system_u:object_r:fs_t,s0);
 
 # Use the allocating task SID to label inodes in the following filesystem
 # types, and label the filesystem itself with the specified context.
@@ -324,6 +325,15 @@ genfscon reiserfs / gen_context(system_u:object_r:nfs_t,s0)
 genfscon panfs / gen_context(system_u:object_r:nfs_t,s0)
 genfscon gadgetfs / gen_context(system_u:object_r:nfs_t,s0)
 genfscon 9p / gen_context(system_u:object_r:nfs_t,s0)
+
+#
+# virtiofs_t is the default type for virtio file systems
+# and their files.
+#
+type virtiofs_t;
+fs_noxattr_type(virtiofs_t)
+files_mountpoint(virtiofs_t)
+genfscon virtiofs / gen_context(system_u:object_r:virtiofs_t,s0)
 
 ########################################
 #


rhatdan:
@wrabcak PTAL

wrabcak:
Hi Dan, 

@WOnder93  looked into kernel and it looks like we need only change in policy. 

Do we have easy reproducer ? It would be great to test the use-case with scratch build before merge. 

Thanks,
Lukas.

rhvgoyal:
@wrabcak I have a setup where I can easily reproduce it. All, I had to do was mount virtiofs filesystem in a vm and do a "chcon unconfined_u:object_r:admin_home_t:s0 /mnt/virtiofs/bar.txt" and I get -EOPNOTSUPP. Same command succeeds on /tmp/bar.txt.

So if a scratch build is available, I can give it a try.

rhvgoyal:
For more context, a user of virtiofs brought forward this issue on virtio-fs mailing list.

https://www.redhat.com/archives/virtio-fs/2020-November/msg00088.html

WOnder93:
FWIW, you can apply the fix locally without rebuilding the selinux-policy RPM as follows:
```
# echo '(fsuse xattr virtiofs (system_u object_r fs_t ((s0) (s0))))' >virtiofs.cil
# semodule -i virtiofs.cil
```
And to check that the change is applied:
```
# seinfo --fs_use | grep virtiofs
    fs_use_xattr virtiofs system_u:object_r:fs_t:s0;
```
To revert the local workaround:
```
# semodule -r virtiofs
```

rhatdan:
Cool, I did not know that now works.  /me needs to keep up a little better on SELinux features.

wrabcak:
@rhatdan, Is there any urgent issue related to this? Do we need to do some quick backport to already released fedora?

zpytela:
An F33 copr-build available here:
https://download.copr.fedorainfracloud.org/results/zpytela/selinux-policy/fedora-33-x86_64/01779444-selinux-policy/
or use
```
f33# dnf copr enable zpytela/selinux-policy
```

rhatdan:
We need this in F33 as well as RHEL8.4.  F32 when you are next releasing.

rhvgoyal:
@WOnder93 Thanks for the instructions. I tried it and it works for me. I am now successfully able to do a chcon on a virtiofs file. I am assuming this will also resolve the restorecon issue.

zpytela:
@rhvgoyal thank for confirming. It would be good to see it working in scenarios described in the discussions, but not sure if you have resources for that.

@rhatdan as soon as the pr is merged, I will backport it to f33/f32/rhel84.

rhvgoyal:
@zpytela I will request the reporter and see if he is willing to test it using above procedure.

hcoin:
Testing results follow.  Short version:  Commands above applied without
error,  failure remains until vm is rebooted, then success.  Good enough
for today!

Thanks

Harry Coin

---

  Detail:

In this case, the VM host is running a debian/ubuntu os, not running
selinux, the underlying filesystem is btrfs. 

root@noc1:~# uname -a
Linux noc1.1.quietfountain.com 5.8.0-29-generic #31-Ubuntu SMP Fri Nov 6
12:37:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

ps axu

...

root       84306  0.0  0.0  80216   992 ?        Sl   14:45   0:00
/usr/lib/qemu/virtiofsd --fd=44 -o
source=/vmsystems/fedora_generic,xattr,flock,no_posix_lock -o writeback

root       84356  4.1  0.0 4165836 30088 ?       Sl   14:45   1:23
/usr/lib/qemu/virtiofsd --fd=44 -o
source=/vmsystems/fedora_generic,xattr,flock,no_posix_lock -o writeback

...

On the otherwise default fedora workstation guest we have:

[root@fedora ~]# uname -a
Linux fedora.1.quietfountain.com 5.9.10-200.fc33.x86_64 #1 SMP Mon Nov
23 18:12:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
[root@fedora ~]# mount

...

myfs on / type virtiofs (rw,relatime)

...

[root@fedora ~]# touch foo
[root@fedora ~]# restorecon foo
restorecon: Could not set context for /root/foo:  Operation not supported
[root@fedora ~]# echo '(fsuse xattr virtiofs (system_u object_r fs_t ((s0) (s0))))' >virtiofs.cil
[root@fedora ~]# semodule -i virtiofs.cil
[root@fedora ~]# seinfo --fs_use | grep virtiofs

fs_use_xattr virtiofs system_u:object_r:fs_t:s0;

[root@fedora ~]# restorecon foo
restorecon: Could not set context for /root/foo:  Operation not supported
[root@fedora ~]# touch foo2
[root@fedora ~]# restorecon foo2
restorecon: Could not set context for /root/foo2:  Operation not supported
[root@fedora ~]# reboot

...

[root@fedora ~]# restorecon foo2
[root@fedora ~]# touch foo3
[root@fedora ~]# restorecon foo3
[root@fedora ~]#

rhvgoyal:
I too had to unmount and mount virtiofs again for this new rule to take affect. Not sure why new rule takes affect only after remount or a fresh mount. May be some caching of rule somewhere.

WOnder93:
The `fs_use` directives are applied at mount time and cannot be changed for the given superblock (it would be quite non-trivial to implement changing it dynamically, AFAIK).

I guess my only concern with this change is what happens if the host FS that is being mapped to guest doesn't support xattrs and SELinux is not enabled on the host... I guess both before and after this change all files would get some default label (without a possibility to change it), although it may be a different one after this patch. Would you guys be able to test that scenario as well? (Just try passing through some vfat FS or something like that.) Anyway, in such case the whole mount can be given a fixed label using the `context=` mount option, so it's not a problem if you can configure the mount options.

hcoin:
I don't have a way to cook up some other FS on the host, sorry.
Can anyone mention the date after which public default downloads will incorporate this fix?

rhvgoyal:
> The `fs_use` directives are applied at mount time and cannot be changed for the given superblock (it would be quite non-trivial to implement changing it dynamically, AFAIK).
> 
> I guess my only concern with this change is what happens if the host FS that is being mapped to guest doesn't support xattrs and SELinux is not enabled on the host... I guess both before and after this change all files would get some default label (without a possibility to change it), although it may be a different one after this patch. Would you guys be able to test that scenario as well? (Just try passing through some vfat FS or something like that.) Anyway, in such case the whole mount can be given a fixed label using the `context=` mount option, so it's not a problem if you can configure the mount options.

@WOnder93  

"virtiofsd" has an option to enable/disable xattr support. With virtiofs policy module loaded, if I run virtiofsd without xattr enabled, mount fails. Previously it used to work.

virtiofsd returns -ENOSYS when it receives GETXATTR request. And mount fails. 

If I remove above said virtiofs selinux policy module, then mount without xattr is successful.

So if once choses to enable SELinux in guest, they will have to run virtiofsd by default with xattr enabled. Otherwise mount will fail.

Does SELinux have any notion of fallback to something else if xattr are not supported. Or one needs to use "context=" option and then it will work.

rhvgoyal:
@rhatdan Dan can we come up with a fallback policy so that SELinux defaults to something if xattr are not supported. Instead of failing the mount

rhatdan:
@zpytela @wrabcak Looks like we also need:

```
#
# virtiofs_t is the default type for Virtio file systems
# and their files.
#
type virtiofs_t;
fs_noxattr_type(virtios_t)
files_mountpoint(virtios_t)
genfscon virtiofs / gen_context(system_u:object_r:virtios_t,s0)
```

WOnder93:
I'm afraid that won't work. It's not possible to have both genfscon and fs_use_xattr applied from the policy to a single FS. At the kernel level it is possible to have all files without label get the same "fallback" label, while files with a security.selinux xattr set would get the label from the xattr. This is being used for cgroupfs and we also used in RHEL-8 for cephfs (which added xattr support, but wanted unlabeled files to still get the same label as before for backwards compatibility). However, I'm not sure if such slightly ugly solution would be accepted upstream.

rhatdan:
This is the behavior of NFS, which is where I copied the policy from?  Does NFS also have special code in the kernel?  The kernel engineers are requesting this feature, so if the kernel needs to be modified, now is the time.

WOnder93:
Yes, NFS is a very special case. It uses so called "native" labeling, which is applied instead of genfscon when the filesystem signals that it supports it. It doesn't work directly through xattrs, but rather the filesystem informs SELinux what label to use via an LSM hook (I don't know the details, but that's roughly how it works). You can't use that for virtiofs without modifying it to use this interface (but you don't want to do that, because that thing is IMO badly designed anyway...). I'll kick off a discussion on the SELinux mailing list to get the community's opinion on how to resolve this situation. Unfortunately there doesn't seem to be any quick and easy solution.

WOnder93:
@rhatdan Could you please fix the typo in this PR (s/gevirtiofscon/genfscon/) so we can merge it? Along with the recently merged kernel patch it should be enough to make this work the way we want.

rhatdan:
Fixed.

WOnder93 (approved):

zpytela:
Thank you, Dan and Ondrej, merging.

PR #479, opened 2020-11-20
Label /run/agetty.reload as getty_var_run_t

thkukuk:
The file is created as /run/agetty.reload, so the rule for
/var/run/agetty.reload does not match and the file gets the wrong label.

### DIFF
--- a/policy/modules/system/getty.fc
+++ b/policy/modules/system/getty.fc
@@ -12,6 +12,7 @@
 
 /var/run/mgetty\.pid.*	--	gen_context(system_u:object_r:getty_var_run_t,s0)
 /var/run/agetty\.reload.*	--	gen_context(system_u:object_r:getty_var_run_t,s0)
+/run/agetty\.reload.*	--	gen_context(system_u:object_r:getty_var_run_t,s0)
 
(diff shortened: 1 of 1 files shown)


zpytela:
@thkukuk there are equivalency rules so the current settings should be sufficient:
```
# semanage fcontext -l |tail
/var/lib/xguest/home = /home
/var/named/chroot/usr/lib64 = /usr/lib
/var/named/chroot/lib64 = /usr/lib
/var/named/chroot/var = /var
/home-inst = /home
/home/home-inst = /home
/var/roothome = /root
/sbin = /usr/sbin
/sysroot/tmp = /tmp
/var/usrlocal = /usr/local
# matchpathcon /run/agetty.reload
/run/agetty.reload      system_u:object_r:getty_var_run_t:s0
```
Can you describe how you got into your issue?

zpytela:
Just spotted the equivalency is not in the output.
```
# semanage fcontext -l |grep "^/run.*="
/run = /var/run
/run/lock = /var/lock
/run/systemd/system = /usr/lib/systemd/system
/run/systemd/generator = /usr/lib/systemd/system
/run/systemd/generator.late = /usr/lib/systemd/system
```

thkukuk:
Seems more like a broken machine...

PR #484, opened 2020-11-30
Add support scripts for COPR automation and RPM build to CI

WOnder93:
This will allow building COPR builds directly from source git, without the need for uploading SRPMs or having a dedicated dist-git repo. As an added bonus, this COPR script can also be used in the CI to do a full RPM build using the current rawhide dist-git.


WOnder93:
Here is an example COPR build against an earlier version of this branch:
https://copr.fedorainfracloud.org/coprs/omos/selinux-policy-experimental/build/1799442/
```
Source Type: Build from an SCM repository
SCM type: git
Clone URL: https://github.com/WOnder93/selinux-policy
Committish: srpm
Build SRPM with: make_srpm
```

zpytela:
Looks interesting.

PR #485, opened 2020-11-30
Fix a subset of SELint issues and add partial SELint checking to CI

WOnder93:
This PR is based on top of #484.

Fix "low-hanging fruit" subset of SELint issues and run it in CI with all currently passing checks enabled.

Cc: @vmojzis

### DIFF
--- a/policy/modules/contrib/daemontools.if
+++ b/policy/modules/contrib/daemontools.if
@@ -43,7 +43,7 @@ interface(`daemontools_service_domain',`
 		type svc_run_t;
 	')
 
-	domain_auto_trans(svc_run_t, $2, $1)
+	domain_auto_transition_pattern(svc_run_t, $2, $1)
 	daemontools_ipc_domain($1)
 
 	allow svc_run_t $1:process signal;
--- a/policy/modules/contrib/sandbox.te
+++ b/policy/modules/contrib/sandbox.te
@@ -49,7 +49,7 @@ corecmd_exec_all_executables(sandbox_domain)
 
 dev_dontaudit_getattr_all(sandbox_domain)
 
-files_rw_all_inherited_files(sandbox_domain, -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t )
+files_rw_all_inherited_files(sandbox_domain, { -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t })
 corecmd_entrypoint_all_executables(sandbox_domain)
 files_entrypoint_all_mountpoint(sandbox_domain)
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,3 +35,29 @@ jobs:
         with:
           name: rpms
           path: rpmbuild/RPMS
+  selint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y uthash-dev libconfuse-dev autoconf-archive m4
+      - name: Build templated .if and .te files
+        run: make generate
+      - name: Clone the SELint repo
+        run: git clone https://github.com/TresysTechnology/selint /tmp/selint
+      - name: Build SELint
+        run: |
+          cd /tmp/selint
+          git checkout d939a02e7528e623c8f8a091c392ed021047dba4
+          autoreconf -i
+          ./configure --without-check
+          make -j $(nproc)
+          sudo make -j $(nproc) install
+      - name: Run SELint
+        run: |
+          selint -s -r . -F \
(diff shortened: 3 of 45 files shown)


WOnder93:
Looks like SELint added some new checks... I'll rather switch to using a fixed commit and then we can bump it regularly (along with fixing/silencing any newly found issues).

policy/modules/contrib/sandbox.te
```diff
@@ -49,7 +49,7 @@ corecmd_exec_all_executables(sandbox_domain)
 
 dev_dontaudit_getattr_all(sandbox_domain)
 
-files_rw_all_inherited_files(sandbox_domain, -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t )
+files_rw_all_inherited_files(sandbox_domain, `-exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t')
 corecmd_entrypoint_all_executables(sandbox_domain)
```
ghost, line 53: I think it might have been complaining about the trailing whitespace and that the `' is probably not needed (it might even cause unexpected results)

```
- files_rw_all_inherited_files(sandbox_domain, -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t )
+ files_rw_all_inherited_files(sandbox_domain, -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t)
```

policy/modules/contrib/daemontools.if
```diff
@@ -43,7 +43,7 @@ interface(`daemontools_service_domain',`
 		type svc_run_t;
 	')
 
-	domain_auto_trans(svc_run_t, $2, $1)
+	domain_auto_transition_pattern(svc_run_t, $2, $1)
```
ghost, line 46: For all of these: have you considered using `domtrans_pattern()` instead? Seems more appropriate.

WOnder93: Hm... the extra permissions seem logical, but I don't know the history of these interfaces and refpolicy also replaced it with `domain_auto_transition_pattern()`, so I just went with the safe (equivalent) choice for this PR.

ghost:
Ondrej Mosnáček <notifications@github.com> writes:

> @WOnder93 commented on this pull request.
>
> ------------------------------------------------------------------------------------------------------
>
> In policy/modules/contrib/daemontools.if:
>
>> @@ -43,7 +43,7 @@ interface(`daemontools_service_domain',`
>  		type svc_run_t;
>  	')
>  
> -	domain_auto_trans(svc_run_t, $2, $1)
> +	domain_auto_transition_pattern(svc_run_t, $2, $1)
>
> Hm... the extra permissions seem logical, but I don't know the history of these interfaces and
> refpolicy also replaced it with domain_auto_transition_pattern(), so I just went with the safe
> (equivalent) choice for this PR.

I see, I agree that if refpolicy uses domain_auto_transition_pattern()
that it is probably a good idea to use that (at least for now)

Eventually though I suspect that in many cases domtrans_pattern() is
more appropriate as it addresses pipe permissions which are often needed
to some extend or another.

>
> —
>

-- 
gpg --locate-keys dominick.grift@defensec.nl
Dominick Grift

policy/modules/contrib/sandbox.te
WOnder93: No, removing the trailing space doesn't silence the warning... But maybe the check is simply wrong here and it's OK without the quotes. IIUC, the only difference quotes make is that they suppresses macro expansion and special characters, plus allow having a newline character in the argument. I guess I'll just remove the extra space and suppress W-007 for this line...

ghost: Yes, this interface call is a bit of a hack no matter how you look at it. There is no easy way to address this cleanly given the current limitations of m4.

I guess the thing is that the interface expects 2 param's and that the easiest way out here was to trick it to accept more than 2 param's. A type attribute could have been created and then be referenced as the second parameter but that would also present difficulties.

WOnder93:
One of the stub interfaces was still being used by container-selinux. I opened a PR to replace all stub calls in container-selinux with direct gen_require()s (which are no worse, really): https://github.com/containers/container-selinux/pull/123

policy/modules/contrib/sandbox.te
cgzones: The warning can be avoid by grouping the arguments:

```
files_rw_all_inherited_files(sandbox_domain, { -exec_type -configfile -usr_t -lib_t -locale_t -var_t -var_run_t -device_t -rpm_log_t })
```

WOnder93: Indeed, that seems to generate the same rules. Although depending on the interpretation of the `{}` grouping one might expect `{ A { -B -C } }` to yield `A | (~B | ~C)`, it seems the parser just discards the `{}` and subtracts the union of the "negative" sets from the union of the "positive" sets (or whatever the logic is).

Thanks for the suggestion!

PR #486, opened 2020-11-30
Add groupadd_t fowner capability

zpytela:
The sssd daemon is configured to run as a non-root user. Its database
files need to have ownership changed accordingly, as a result some
capabilities may be required to change metadata of the files.

When sss_cache is called as a child of groupadd executed from a confined
domain, e. g. rpm_script_t, it needs the fowner capability for groupadd_t
to complete the groupadd command and run utime() syscall.

Resolves: rhbz#1884179


zpytela:
Pre-tested, merging.

PR #488, opened 2020-12-01
FreeRADIUS policy updates to enable support for DHCP server, DHCP client, VPMS and BFD

terryburton:
net_raw is needed for directly binding to interfaces (SO_BINDTODEVICE) as required to support DHCP.

The daemon is a multi-protocol server that commonly provides network services other than RADIUS.


terryburton:
Is there anything further that I can do to help progress this?

zpytela:
@terryburton I am sorry for the delay. I have only a few minor comments.

According to IANA, port 1589 is assigned for both tcp and udp, but I don't know much about the protocol or its prospects, seems to be a private one. Does it make sense define it in the policy for tcp, too?

I'd also like to see some kind of order in the rules for ports. There are hardly any good examples in existing policies and definitely the approach is not consistent, so it is not possible to conform to existing state, but keeping alphabetical order whenever possibe is good when somebody else wants to know the ropes. Blocks of related interfaces calls for the same port is fine, just reordering the blocks would help in reading.

Do you have this policy tested and the service working?

terryburton:
@zpytela No problems.

I've added the IANA assignment of TCP/1589 to VQP. It's more of a reservation that of practical use. FreeRADIUS won't use TCP so I've only included UDP in the policy.

I've done a lexical sort on the protocol-specific corenet rules and left generic rules as is. That will at least keep things tidy as new protocol support is added to FreeRADIUS.

The policy has been tested successfully with packaged FreeRADIUS v3 (distro and upstream) and with upstream packages for the FreeRADIUS v4 development branch.

I've posted additional commits for your review. Let me know if you would prefer me to squash and rebase the related commits prior to merge.

zpytela:
@terryburton thank you for your time and effort, no additional change required.

PR #489, opened 2020-12-01
New perf event class

zpytela:

PR #490, opened 2020-12-01
Allow unconfined_u and sysadm_u users manage perf_events

zpytela:
Resolves: rhbz#1901957


WOnder93:
Could you please put all three patches into a single PR? Since the latter two depend on the first one, it doesn't really make sense to separate them. We don't want to merge the first patch adding the definitions before we identify all necessary rules, so the patches will have to be applied together anyway.

zpytela:
Frankly, I split it intentionally into 3 parts, but I can un-split them if it helps.

zpytela:
Closing in favor of https://github.com/fedora-selinux/selinux-policy/pull/489

PR #491, opened 2020-12-01
Allow pcp-pmcd open perf events

zpytela:


zpytela:
Closing in favor of https://github.com/fedora-selinux/selinux-policy/pull/489

PR #492, opened 2020-12-02
Libvirt SELinux

5umm3r15:
New SELinux policy for libvirt drivers:
Hypervisor drivers:
 - virtqemud  (QEMU/KVM)
 - virtlxcd   (LXC)
 - virtvboxd  (VirtualBox)

Secondary drivers:
 - virtstoraged   (host storage mgmt)
 - virtnetworkd   (virtual network mgmt)
 - virtinterface  (network interface mgmt)
 - virtnodedevd   (physical device mgmt)
 - virtsecretd    (security credential mgmt)
 - virtnwfilterd  (ip[6]tables/ebtables mgmt)
 - virtproxyd     (proxy daemon)

SELinux policy for virtvxz and virtxend has not been created yet.

In the first commit are services separated from the original virt files that are not libvirt related and virt_supplementary policy module is created.

In the second commit is designed policy for libvirt drivers, including modification of old virt interfaces and also creating new interfaces.

Copr build of SELinux policy with new libvirt policy:
https://copr.fedorainfracloud.org/coprs/nknazeko/selinux-policy/

Libvirt repository with SELinux subpackage is available on github:
https://github.com/5umm3r15/libvirt-selinux/tree/selinux

Copr build of Libvirt with SELinux subpackage: https://copr.fedorainfracloud.org/coprs/nknazeko/libvirt-selinux/

SELinux policy was also created for Libvirt Testsuite and used for testing virt policy. 
The repository is available on github: https://github.com/5umm3r15/perl-Sys-Virt-TCK-selinux/tree/selinux
Copr build of Libvirt-TCK with SELinux subpackage: https://copr.fedorainfracloud.org/coprs/nknazeko/libvirt-tck-selinux/

Libvirt test suite gives AVC messages related to this bug: 
[BZ1858260 - SELinux prevents svirt_t read|write on lvm_control_t during VM creation](https://bugzilla.redhat.com/show_bug.cgi?id=1858260)


5umm3r15:
@zpytela @vmojzis @wrabcak Can you please review it?

wrabcak:
Thank you @5umm3r15 , what I don't understand is why is this PR against selinux-policy ?

5umm3r15:
To be more transparent with changes against old policy.

5umm3r15:
Hi! This is the final version of virt and virt_supplementary policies, which will be part of our policy.
Virt.* policy files will be also included in SELinux policy subpackage for Libvirt component.
@vmojzis @zpytela @wrabcak can you please review it? 
Thank you

policy/modules/contrib/virt.te
```diff
@@ -14,13 +14,19 @@ attribute virsh_transition_domain;
...
+attribute virt_driver_var_run;
 attribute virt_image_type;
 attribute virt_tmpfs_type;
 attribute svirt_file_type;
 attribute virt_file_type;
 attribute sandbox_net_domain;
 attribute sandbox_caps_domain;
 
+permissive virtxend_t;
+permissive virtvzd_t;
```
WOnder93: There could also be a code comment explaining why these are permissive for now.

WOnder93 (commented):
I took a brief look at the changes and didn't find anything that would stick out as a problem. It looks very nice, good job!

Apart from the minor comment about documenting the permissive domains, I have one more style request/suggestion: Please try to use consistent whitespace. When modifying existing code, try to match the style of the surrounding code; when adding new code (new files) I'd suggest to use tabs, since those are used more often in the codebase:
```
$ git grep '    ' policy/modules/*/*.{te,if} | wc -l
6028
$ git grep '\t' policy/modules/*/*.{te,if} | wc -l
102854
```
The current state of whitespace is quite bad, but at least we could try to make it not worse / slightly better with new changes, IMHO :)

@zpytela What's your opinion on whitespace consistency?

zpytela:
Whitespace consistency is really poor in the whole policy context; I also prefer tabs to spaces in new policy, and only use spaces when the rest of the file uses them.

policy/modules/contrib/virt.te
wrabcak: There is also a possibility to add these domains for Fedora here: https://src.fedoraproject.org/rpms/selinux-policy/blob/master/f/permissivedomains.cil 

and we can keep policy sources without permissive domain.

WOnder93: I don't think that would make much sense. If the policy is absent, it simply wouldn't work without the `permissive`s. And it's not good to have this information in a different repo - it's *here* that the policy needs to be added in the future, so the permissive declarations should also be here. Then they can also be removed atomically in the same commit that adds the missing policy.

wrabcak: No problem, as you say. :) In past, we kept permissive domains there. Also, if there is no SELinux policy for these types, what about unconfined_domain() macro? Using this macro we avoid lots of SELinux policy denials and confused users. WDYT?

zpytela:
@5umm3r15 I see a lot of work has been done here, thank you for that. In this context I only have some minor comments.

1. Please convert all leading spaces to the right amount of tabs. Similarly do so in the comments inside an interface.
2. Block with booleans should be the first one in the te file.
3. Interfaces calls after kernel_* should be in alphabetical order in the te file.

I would also like to ask: Is this split meant temporarily or for good? It may be a bit confusing where to look for any virt_ interface or type in the future. In that case, we can at least have some brief explanation/description.

policy/modules/contrib/virt.te
WOnder93: That might indeed be a better approach, but I don't know enough about either approach to judge :)

zpytela: I think the main question is: do we want to have users receive AVCs and report them or not?

IMHO in this stage of development should rather be permissive.

5umm3r15: I would prefer to have AVC's reported from Libvirt team, not users. For each driver I had many AVC messages and it might be confusing for users. Maybe it will be better, to use unconfined_domain macro until the AVC messages are reproduced internally and policy for these drivers is made.

5umm3r15:
@zpytela @WOnder93 @wrabcak 
Thank you for the review. I replaced spaces by tabs, alphabetically ordered interfaces in kernel layer modules,
then in system layer modules, boolean blocks and optional policy blocks according to [style guide](https://github.com/TresysTechnology/refpolicy/wiki/StyleGuide).

The main virt policy, types and interfaces, is in the virt module, that will be handled by Libvirt. A lot of modules are using virt interfaces, so both virt and virt_supplementary modules will be in our policy for now. I am not sure if we decide to handle this problem another way in the future.

policy/modules/contrib/virt.te
vmojzis: Agreed. I'd use the unconfined_domain here and then permissive domains in the policy libvirt-selinux will ship (this version will be tested by libvirt team). This way users can start using the new policy (split-daemon configuration will be used by default soon), we have time to polish the policy based on libvirt testing, and any issues still present in the final policy can be reported by users.

vmojzis:
Rebased and added virt_supplementary to policy/modules.conf

PR #494, opened 2020-12-03
Update systemd-sleep policy

zpytela:
Allow systemd-sleep write to the /sys/power directory.
Dontaudit sys_ptrace capability.


zpytela:
My rawhide scratchbuild passes the updated tests with no denial. I will merge this PR soon if there are no objections.

PR #495, opened 2020-12-04
Updates to varnishlog_t

dridi:
This is a small series of changes to polish the `varnishlog_t` type. The mmap change has been running in production for a while now, ever since it was discussed a year and a half ago:

https://github.com/fedora-selinux/selinux-policy-contrib/pull/110#issuecomment-493926534

### DIFF
--- a/policy/modules/contrib/varnishd.if
+++ b/policy/modules/contrib/varnishd.if
@@ -73,7 +73,8 @@ interface(`varnishd_read_lib_files',`
 	')
 
 	files_search_var_lib($1)
-	read_files_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
+	list_dirs_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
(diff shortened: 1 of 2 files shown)


zpytela:
@Dridi I would like to double check: The execmem permission is required any time the varnishlog service is running?

dridi:
Not necessarily any time, but whenever a regular expression is used with `varnishlog` or `varnishncsa`, for example:

```
varnishncsa -q 'ReqHeader:Host ~ "\.?example\.com$"'
```

`libvarnishapi` will unconditionally enable JIT compilation if it is available on the system.

See the detection code here:

https://github.com/varnishcache/varnish-cache/blob/6.5/lib/libvarnish/vre.c#L45-L49

And the JIT compilation is systematically enabled here:

https://github.com/varnishcache/varnish-cache/blob/6.5/lib/libvarnish/vre.c#L90

Similarly to how `varnishd_t` unconditionally gets the `execmem` permission, I decided against adding a boolean for `varnishlog_t`.

zpytela:
Merging, thank you for your contribution.

PR #496, opened 2020-12-07
Adds rule to allow glusterd to access RDMA socket

lveyde:
Resolves: rhbz#1779052

Signed-off-by: Lev Veyde <lveyde@redhat.com>

### DIFF
--- a/policy/modules/contrib/glusterd.te
+++ b/policy/modules/contrib/glusterd.te
@@ -79,6 +79,7 @@ allow glusterd_t self:tcp_socket { accept listen };
 allow glusterd_t self:unix_stream_socket { accept listen connectto };
 allow glusterd_t self:rawip_socket create_socket_perms;
 allow glusterd_t self:unix_stream_socket create_stream_socket_perms;
+allow glusterd_t self:netlink_rdma_socket create_socket_perms;
 
 manage_dirs_pattern(glusterd_t, glusterd_conf_t, glusterd_conf_t)
 manage_files_pattern(glusterd_t, glusterd_conf_t, glusterd_conf_t)


lveyde:
@wrabcak @rhatdan

policy/modules/contrib/glusterd.te
```diff
@@ -79,6 +79,7 @@ allow glusterd_t self:tcp_socket { accept listen };
 allow glusterd_t self:unix_stream_socket { accept listen connectto };
 allow glusterd_t self:rawip_socket create_socket_perms;
 allow glusterd_t self:unix_stream_socket create_stream_socket_perms;
+allow glusterd_t self:netlink_rdma_socket { bind create setopt };
```
zpytela: Please use one of the macros available for socket permissions in https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt

zpytela (changes requested):
@lveyde, do you have any test using this socket type?

zpytela:
@lveyde please also ensure the same commit goes to the glusterfs-selinux package as the gluster team manages their own policy. I actually expect all future development happens only there.

wrabcak:
@zpytela, Can you provide the link for the gluster-selinux repo, so all parties are on same page where the commit should go?

zpytela:
https://github.com/gluster/glusterfs-selinux/

WOnder93 (approved):

zpytela (approved):

zpytela:
Thank you for the PR; consider cherry-picking to https://github.com/gluster/glusterfs-selinux/

PR #497, opened 2020-12-07
Allow systemd-logind manage pid files

zpytela:
Added init_manage_pid_files() interface.


WOnder93:
Maybe I would reword it to "manage init's pid files", but otherwise looks good to me.

PR #499, opened 2020-12-08
Add systemd-resolved manage its private runtime symlinks

zpytela:
Resolves: rhbz#1896796


WOnder93:
s/Add/Allow/, otherwse LGTM.

zpytela:
Changed to Allow and an interface updated, the access was requested by sssd and may be more common.

WOnder93 (approved):
Makes sense :+1:

zpytela:
Thanks @WOnder93 .

PR #503, opened 2020-12-09
Fb nsswitchdomain resolve

zpytela:


WOnder93 (approved):
LGTM.

BTW, I noticed that [this line in domain.te](https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/domain.te#L682) generates the following allow rule when the container module is enabled:
```
allow domain pidfile:sock_file { append getattr open write };
```
That seems overly permissive, although I'm not sure how to address it...

EDIT: I opened an issue for it so it is tracked: #504

keszybz:
Late review, but LGTM.

zpytela:
Thanks anyway @keszybz

PR #509, opened 2020-12-14
Allow stub-resolv.conf to be a symlink

WOnder93:
(And a small cleanup to remove a reference to `systemd_resolved_var_run_t` from `sysnetwork.if`.)

Tested on a Rawhide VM image by replacing `/var/run/systemd/resolve/stub-resolv.conf` with a symlink to `resolv.conf` and restarting `chronyd.service`.

### DIFF
--- a/policy/modules/system/sysnetwork.fc
+++ b/policy/modules/system/sysnetwork.fc
@@ -38,7 +38,7 @@ ifdef(`distro_redhat',`
 /etc/sysconfig/network-scripts(/.*)? gen_context(system_u:object_r:net_conf_t,s0)
 /var/run/systemd/network(/.*)?  gen_context(system_u:object_r:net_conf_t,s0)
 /var/run/systemd/resolve/resolv\.conf   --  gen_context(system_u:object_r:net_conf_t,s0)
-/var/run/systemd/resolve/stub-resolv\.conf   --  gen_context(system_u:object_r:net_conf_t,s0)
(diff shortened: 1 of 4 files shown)


dustymabe:
> Tested on a Rawhide VM image by replacing `/var/run/systemd/resolve/stub-resolv.conf` with a symlink to `resolv.conf` and restarting `chronyd.service`.

You don't need to create the symlink yourself. It should suffice to:

```
sudo mkdir /etc/systemd/resolved.conf.d                                                                                                     
echo -e '[Resolve]\nDNSStubListener=no' | sudo tee /etc/systemd/resolved.conf.d/no-stub.conf
sudo systemctl restart systemd-resolved
```

That will create the symlink (and all other files in /run/systemd/resolve/) for you.

Throw in a reboot for good measure if you like.

WOnder93:
After further investigation we discovered that an additional fix in systemd is needed for the symlink to get the correct label:
https://github.com/systemd/systemd/pull/17976

dustymabe:
> [systemd/systemd#17976](https://github.com/systemd/systemd/pull/17976)

Merged upstream now. Can we get this policy update into a fedora 33 build today? We can workaround the need for that systemd PR for now until it works its way downstream.

zpytela:
Merging, thanks.

zpytela:
Backporting to F33;
https://github.com/fedora-selinux/selinux-policy/pull/513

PR #511, opened 2020-12-15
Allow sssd, unix_chkpwd, groupadd stat /proc filesystem

zpytela:
With the f2d77890bfcbe5b514c6205f288eeb73fe2225af commit, the getattr
permission for the proc filesystem was allowed to the login_pgm
attribute. It is required by all services or commands working with pam
though.

Resolves: rhbz#1907365

### DIFF
--- a/policy/modules/admin/usermanage.te
+++ b/policy/modules/admin/usermanage.te
@@ -227,6 +227,8 @@ allow groupadd_t self:unix_stream_socket create_stream_socket_perms;
 allow groupadd_t self:unix_dgram_socket sendto;
 allow groupadd_t self:unix_stream_socket connectto;
 
+kernel_getattr_proc(groupadd_t)
+
 fs_getattr_xattr_fs(groupadd_t)
 fs_search_auto_mountpoints(groupadd_t)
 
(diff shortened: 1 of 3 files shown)


WOnder93:
This is most likely related to a [change in labcap-ng](https://src.fedoraproject.org/tests/selinux/pull-request/151#comment-62553) - it would be nice to have that information in the commit message.

I'm OK with the patch, although I wonder if at this point we wouldn't be better off by allowing statfs on procfs or the whole `domain`... Who knows what other programs will use libacp-ng or what other code decides to use the same trick? It isn't really a security-sensitive operation, as the returned data contains only:
```
fstatfs(3, {f_type=PROC_SUPER_MAGIC, f_bsize=4096, f_blocks=0, f_bfree=0, f_bavail=0, f_files=0, f_ffree=0, f_fsid={val=[0, 0]}, f_namelen=255, f_frsize=4096, f_flags=ST_VALID|ST_NOSUID|ST_NODEV|ST_NOEXEC|ST_RELATIME}) = 0
```
But I'm just thinking out loud here, feel free to just merge this PR :)

zpytela:
@WOnder93, my understanding is pam was involved in the requirement: https://bugzilla.redhat.com/show_bug.cgi?id=1853730#c6

I also wondered if allowing for domain would be safe before addresing that bz where it clearly was related to logging in only.

WOnder93:
I see, so there are actually multiple uses of such pattern... That's even more reason to consider allowing it universally :) But I guess we can wait if any more domains pop up that need it. Right now it seems to be allowed to only about half domains so maybe it's not really a good candidate to allow universally.

WOnder93 (approved):

zpytela:
@WOnder93 so I have changed the commit message a bit.

WOnder93:
Thanks, looks good to me!

PR #514, opened 2020-12-15
Allow sssd read /run/systemd directory

zpytela:
The nsswitch_domain is already allowed search /run/systemd, sssd however
requires the read permission, granted by the list_dir_perms pattern.

Resolves: rhbz#1903335

### DIFF
--- a/policy/modules/contrib/sssd.te
+++ b/policy/modules/contrib/sssd.te
@@ -144,6 +144,8 @@ auth_manage_cache(sssd_t)
 auth_login_manage_key(sssd_t)
 dontaudit sssd_t init_t:dbus send_msg;
 
+# sssd uses inotify to watch for changes in /run/systemd/resolve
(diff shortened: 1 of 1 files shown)


WOnder93:
Maybe we should also add a code comment that it is needed to allow watching for changes in /run/systemd/resolve (IIUC). This is one of the places that will need also appropriate watch permissions once we add them, so such comment could be helpful.

zpytela:
Good point, updated to understand the background.

WOnder93:
I meant adding a comment in the *code*, not just the commit message :) It can be very short, just should make it clear that the call is there because sssd wants to watch `/run/systemd/resolve/resolv.conf` (which usually entails adding watches to all path components, IIRC). But the commit message is nice, so it wasn't a wasted effort :+1:

zpytela:
Comment added.

PR #515, opened 2020-12-16
Label /dev/wmi/dell-smbios as acpi_device_t

zpytela:
Resolves: rhbz#1899352


WOnder93:
Hm... maybe we should have a different type for this one? From the info I could find, APM BIOS and Dell SMBIOS don't seem to be quite the same thing.

zpytela:
GIven that APM is dead, would it make a difference to call it acpi? It is just a string.

WOnder93:
If it's dead, then it's IMO all the more reason to use a separate label. If we later allow some domains access to `apm_bios_t` based on the need to use `/dev/wmi/dell-smbios` and at some point later decide to remove all APM-related stuff, how do we tell apart which rules are needed for `/dev/wmi/dell-smbios` vs. `/dev/apm_bios`? (`/dev/snapshot` should also be something else, but that's another thing...)

Or from another POV, it's easier to later merge related labels than to split an overloaded one.

WOnder93 (approved):
Would be nice to also update PR title for posterity, but the content looks good to me.

PR #516, opened 2020-12-16
Allow ipsec set the context of a SPD entry to the default context

zpytela:
Resolves: rhbz#1880474


WOnder93 (approved):
Seems pretty straightforward.

PR #517, opened 2020-12-16
Label /dev/vhost-vdpa-[0-7] as vhost_device_t

zpytela:
To support the change, a set of appropriate file transitions
were added to the dev_filetrans_all_named_dev() interface.

Resolves: rhbz#1907485


WOnder93 (approved):
Makes sense :+1:

zpytela:
I'd like to have clarified first how many interfaces will really be needed, that's why I made this PR a draft.

WOnder93:
I see... According to the kernel code it seems there can be up to `VHOST_VDPA_DEV_MAX` such devices, which is `1 << 20`... That's way too many to cover them all, and we can't have too many filename transitions so I'd go with just 8 or 10. Though maybe we could make at least the fc regex match all of them (`/dev/vhost-vdpa-[0-9]+`).

I need to seriously look at finally adding some prefix matching to filename transitions...

zpytela:
Regexp and commit message updated.

PR #518, opened 2020-12-16
Label /dev/isst_interface as cpu_device_t

zpytela:
To support the change, an appropriate file transition was added
to the dev_filetrans_all_named_dev() interface.

Resolves: rhbz#1902227


WOnder93 (approved):
Seems reasonable.

PR #521, opened 2020-12-17
.copr: improve timestamp format

WOnder93:
Add also time (up to seconds) to minimize the likelihood of bad build
ordering and also make sure the timestamp is normalized to the UTC
timezone.


zpytela:
Can there be a delimiter between %F and %T?

WOnder93:
Yes, that'll probably look nicer. I'll add add an underscore (`_`) between them.

zpytela:
LGTM. Particularly I appreciate the emoticon at the end of line.

PR #522, opened 2020-12-17
Allow openvswitch_t perf_event open permission

zpytela:
Resolves: rhbz#1908817


WOnder93:
Did you try the scenario also in permissive mode? I bet more than just `open` is really needed.

zpytela:
It is suspicious as open without further action hardly makes sense, but no other denial popped up. We need functional testing to test further.
BTW it is the reason why this PR is marked draft.

zpytela:
No further testing revealed other than open permission. I am for merging this PR for now.

WOnder93 (approved):
OK, I had a look at the kernel code and it seems there are some code paths where `open` is needed without `cpu`, `kernel`, or `tracepoint`. So I guess it makes sense to grant just `open` for now.

PR #523, opened 2020-12-20
Fix snapper on silverblue.

hansborr:
On Fedora Silverblue, /mnt is a symlink to /var/mnt.
The regex for snapper contexts does not work for this use case.

Adjust the regex to work for filesystems under both /mnt/ and /var/mnt/

### DIFF
--- a/policy/modules/contrib/snapper.fc
+++ b/policy/modules/contrib/snapper.fc
@@ -7,7 +7,7 @@
 
 /var/log/snapper\.log.* --  gen_context(system_u:object_r:snapperd_log_t,s0)
 
-/mnt/(.*/)?\.snapshots(/.*)?   gen_context(system_u:object_r:snapperd_data_t,s0)
+/(var/)?mnt/(.*/)?\.snapshots(/.*)?   gen_context(system_u:object_r:snapperd_data_t,s0)
 /\.snapshots(/.*)?   gen_context(system_u:object_r:snapperd_data_t,s0)
 /usr/\.snapshots(/.*)?   gen_context(system_u:object_r:snapperd_data_t,s0)
 /var/\.snapshots(/.*)?   gen_context(system_u:object_r:snapperd_data_t,s0)


zpytela:
@hansborr thank you for the PR. There is no problem with the content; however, regexps at the beginning can have unexpected consequences, so although for /mnt it should not be the case, I'd rather avoid it and have 2 entries instead if you don't mind.

For the record, it is %postin in selinux-policy-* and fixfiles:
```
165 for i in /sys /proc /dev /run /mnt /var/tmp /var/lib/BackupPC /home /tmp /dev; do
166     [ -e $i ]  && EXCLUDEDIRS="${EXCLUDEDIRS} -e $i";
167 done
168 LogExcluded
169
170 if [ -f ${PREFC} -a -x /usr/bin/diff ]; then
171         TEMPFILE=`mktemp ${FC}.XXXXXXXXXX`
172         test -z "$TEMPFILE" && exit
173         PREFCTEMPFILE=`mktemp ${PREFC}.XXXXXXXXXX`
174         sed -r -e 's,:s0, ,g' $PREFC | sort -u > ${PREFCTEMPFILE}
175         sed -r -e 's,:s0, ,g' $FC | sort -u | \
176         /usr/bin/diff -b ${PREFCTEMPFILE} - | \
177             grep '^[<>]'|cut -c3-| grep ^/ | \
178             egrep -v '(^/home|^/root|^/tmp|^/dev)' |\
179         sed -r -e 's,[[:blank:]].*,,g' \
180                -e 's|\(([/[:alnum:]]+)\)\?|{\1,}|g' \
```
etc., replacing all regexps with a wildcard.

zpytela:
@hansborr hold on, we are discussing should this rather be fixed with equivalency; this is to be made in the package dist-git instead of github.

hansborr:
> @hansborr hold on, we are discussing should this rather be fixed with equivalency; this is to be made in the package dist-git instead of github.

Could you provide me with a link to this discussion?

WOnder93:
Sorry, it was a private conversation. I basically suggested adding `/var/mnt /mnt` to `file_contexts.sub` [1], which would basically make any fc rule for `/mnt/...` to also apply to `/var/mnt/...`. There are only a few rules for `/mnt/...` currently, but it would be more future-proof, since as you say on Silverblue `/mnt` is symlinked to `/var/mnt`, and so the same labeling rules should apply to both.

[1] https://src.fedoraproject.org/rpms/selinux-policy/blob/master/f/file_contexts.subs_dist

zpytela:
@hansborr, the PR has been merged:
https://src.fedoraproject.org/rpms/selinux-policy/pull-request/144

It should be fixed in the next selinux-policy package build.

hansborr:
Thanks. I'll close the PR.

Issue #525, opened 2020-12-28
iscsiadm command usage in systemd unit file results in AVC denial


patsoffice:
The issue for this problem was originally raised in https://github.com/coreos/fedora-coreos-tracker/issues/705.

I downloaded the current (as of 2020/12/27) Fedora Cloud image and installed it as a VM.

An iscsi target was created on a local FreeNAS server and the `iscsi-initiator-tools` were installed in the VM. I then created a systemd unit file in `/etc/systemd/system/iscsi-setup-testing.service` with the contents:

```
[Unit]
Description=Add iSCSI node for Fedora Cloud testing volume
Before=iscsid.service
ConditionPathExists=!/var/lib/iscsi/nodes/iqn.2005-10.org.freenas.ctl:fc-testing

[Service]
Type=oneshot
ExecStart=/usr/sbin/iscsiadm -m node -T iqn.2005-10.org.freenas.ctl:fc-testing -o new -p 192.168.1.0:3260
ExecStart=/usr/sbin/iscsiadm -m node -T iqn.2005-10.org.freenas.ctl:fc-testing -o update -n node.startup -v automatic
RemainAfterExit=yes

[Install]
WantedBy=multi-user.target
```

Running `sudo systemctl start iscsi-setup-testing.service` results in a failure and the status is:

```
● iscsi-setup-testing.service - Add iSCSI node for apps volume
     Loaded: loaded (/etc/systemd/system/iscsi-setup-testing.service; disabled; vendor preset: disabled)
     Active: failed (Result: exit-code) since Sun 2020-12-27 17:28:30 PST; 35min ago
    Process: 6999 ExecStart=/usr/sbin/iscsiadm -m node -T iqn.2005-10.org.freenas.ctl:fc-testing -o new -p 192.168.1.0:3260 (code=exited, status=6)
   Main PID: 6999 (code=exited, status=6)
        CPU: 24ms

Dec 27 17:28:28 cloud-00 systemd[1]: Starting Add iSCSI node for apps volume...
Dec 27 17:28:30 cloud-00 iscsiadm[6999]: iscsiadm: Could not open /var/lib/iscsi/nodes/iqn.2005-10.org.freenas.ctl:fc-testing/192.168.1.0,3260: Permission deniedd
Dec 27 17:28:30 cloud-00 iscsiadm[6999]: iscsiadm: Error while adding record: encountered iSCSI database failure
[...]

PR #526, opened 2020-12-28
Add missing dac_override capability for iscsi mod

patsoffice:
This PR should fix the issue raised in https://github.com/fedora-selinux/selinux-policy/issues/525 and https://github.com/coreos/fedora-coreos-tracker/issues/705.

### DIFF
--- a/policy/modules/contrib/iscsi.te
+++ b/policy/modules/contrib/iscsi.te
@@ -35,7 +35,7 @@ files_pid_file(iscsi_var_run_t)
 # Local policy
 #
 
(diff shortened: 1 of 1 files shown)


zpytela:
Thank you for the PR; we need to figure out first though why the `dac_override` capability was requested.

patsoffice:
Closing this PR as the problem should be fixed upstream in open-iscsi. Specifically, the directory creation permissions used are 0660 (vs. 0770) and that's what's the root cause of the dac_override denial.

PR #528, opened 2021-01-04
Allow domain read corepattern

zpytela:


WOnder93 (approved):
I suspect this might actually be a kernel bug. The usermode helper should probably be read under kernel credentials (if my assumption that it is internally read by the kernel when handling a crash is correct). But that will likely take some time to fix, so in the meantime let's allow it to the whole domain.

zpytela:
Rebased to resolve conflicts.

zpytela:
Rebased, fixed the first PR commit message.

PR #529, opened 2021-01-05
Allow subscription manager run kpatch

zpytela:


zpytela:
The scratchbuild seems to allow the permissions as requested.

WOnder93:
The first commit adds an empty `CHECKING` file. I guess that's not intentional?

Also, I think it would make more sense to do a transition here (there is one even from unconfined_t). Even if it doesn't make much difference in this case, it would be more consistent.

zpytela:
A transition was the first thing I considered, but in this case kpatch is used only to gather information for subscription manager.
Thanks for spotting the empty file.

WOnder93:
The CHECKING file is still there :)

I thought a bit more about the exec vs transition and I conclude that it's probably OK like this.

zpytela:
So trying again, hope the file is not immortal.

WOnder93 (approved):

zpytela:
Having double-checked no file creeped in in the meantime, merging.

PR #530, opened 2021-01-05
Add rsync_sys_admin tunable to allow rsync sys_admin capability

zpytela:
The rsync_sys_admin tunable was added to allow rsync the sys_admin
capability conditionally. This permission is required for rsync to
be able to restore files with extended attributes in all attributes
namespaces. Glusterfs makes use of the "trusted" namespace for which
the sys_admin capability is required.
The boolean is off by default.

Resolves: rhbz#1889673

### DIFF
--- a/policy/modules/contrib/rsync.te
+++ b/policy/modules/contrib/rsync.te
@@ -35,6 +35,14 @@ gen_tunable(rsync_anon_write, false)
 ## </desc>
 gen_tunable(rsync_full_access, false)
 
+## <desc>
+##	<p>
+##	Allow rsync sys_admin capability.
+##	This capability is required to restore files
+##	with extended attributes in the "trusted" namespace.
+##	</p>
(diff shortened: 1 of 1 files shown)


policy/modules/contrib/rsync.te
```diff
@@ -35,6 +35,12 @@ gen_tunable(rsync_anon_write, false)
 ## </desc>
 gen_tunable(rsync_full_access, false)
 
+## <desc>
+##	<p>
+##	Allow rsync sys_admin capability
```
WOnder93: Can you please add the information here that this is needed for setting the `trusted.*` extended attributes? I think it would be helpful to have it there, not just in the git history.

zpytela: Description enhanced.

WOnder93 (approved):

PR #531, opened 2021-01-06
Allow staff_u run pam_console_apply

zpytela:
Until now, the staff_u user had the permission to execute
pam_console_apply, but the staff_r role was not associated
appropriately, so the transition did not finish successfully.

### DIFF
--- a/policy/modules/system/authlogin.if
+++ b/policy/modules/system/authlogin.if
@@ -1307,6 +1307,30 @@ interface(`auth_domtrans_pam_console',`
 
 ########################################
 ## <summary>
+##	Execute pam_console in the pam timestamp domain
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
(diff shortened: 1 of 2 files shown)


policy/modules/system/authlogin.if
```diff
@@ -1305,6 +1305,30 @@ interface(`auth_domtrans_pam_console',`
...
+##	Execute pam_console in the pam timestamp domain
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	The role to allow the PAM domain.
```
WOnder93: Nit: this should probably read "allow transitioning into the PAM domain". Otherwise LGTM.

zpytela:
You're absolutely right, I copy-pasted it for the sake of consistency. Now it is the others which need improvement.

WOnder93 (approved):

PR #535, opened 2021-01-08
Allow sysadm read and write /dev/rfkill

zpytela:
This permission is required to run rfkill which enables or disables
wireless devices.

Resolves: rhbz#1831630


zpytela:
Works with a copr-build containing this patch.
https://copr.fedorainfracloud.org/coprs/zpytela/selinux-policy/build/1869723/

WOnder93 (approved):

PR #537, opened 2021-01-12
access_vectors: Add new capabilities to cap2

zpytela:
Updated location of capability definitions to point to current location within kernel source code.

CAP_BPF and CAP_PERFMON mainlined in: cb8e59cc87201af93dfbb6c3dccc8fcad72a09c2, original commit: a17b53c4a4b55ec322c132b6670743612229ee9c
CAP_CHECKPOINT_RESTORE mainlined in: 74858abbb1032222f922487fd1a24513bbed80f9, original commit: 124ea650d3072b005457faed69909221c2905a1f

The missing capabilities were noticed on archlinux with kernel 5.8.14-arch1-1.

Signed-off-by: Dannick Pomerleau <dannickp@hotmail.com>

Resolves: rhbz#1915264

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -130,7 +130,7 @@ common x_device
 #
 common cap
 {
-	# The capabilities are defined in include/linux/capability.h
+	# The capabilities are defined in include/uapi/linux/capability.h
 	# Capabilities >= 32 are defined in the cap2 common.
 	# Care should be taken to ensure that these are consistent with
 	# those definitions. (Order matters)
@@ -177,6 +177,9 @@ common cap2
 	wake_alarm
 	block_suspend
 	audit_read
+	perfmon
+	bpf
+	checkpoint_restore
 }
 
 #
--- a/policy/modules/contrib/linuxptp.te
+++ b/policy/modules/contrib/linuxptp.te
@@ -154,7 +154,7 @@ allow ptp4l_t self:packet_socket create_socket_perms;
 allow ptp4l_t self:unix_stream_socket create_stream_socket_perms;
 allow ptp4l_t self:shm create_shm_perms;
 allow ptp4l_t self:udp_socket create_socket_perms;
-allow ptp4l_t self:capability { net_admin net_raw sys_admin sys_time };
+allow ptp4l_t self:capability { net_admin net_raw sys_time };
 allow ptp4l_t self:capability2 { wake_alarm };
 allow ptp4l_t self:netlink_route_socket rw_netlink_socket_perms;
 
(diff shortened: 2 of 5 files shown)


WOnder93:
We'll also need to add the `perfmon` capability to all domains that use `perf_event`. There is currently a fallback to CAP_SYS_ADMIN, but we shouldn't rely on it long-term. It may be possible to remove `sys_admin` afterwards from some of these domains, but that would need careful testing (or at least verifying that the capability was added specifically to allow something that is now gated by CAP_PERFMON).

WOnder93:
And related to #538, according to the selinux-testsuite policy, `lockdown { confidentiality }` is also needed in case of some `perf_event` related operations (I think only with `cpu` and `tracepoint`, but it will need double-checking).

zpytela:
The perfmon cap has been added.

Checking the changed files, no one seems to be a good candidate for removing sys_admin: it is only present in pcp, added with
5bfcae9780e5176d74d61231d1b2af1650ead09e, so unlikely related to perf actions.

zpytela:
Lockdown confidentiality added, too.

zpytela:
Wait, it should rather go to #538.

policy/modules/kernel/domain.te
```diff
@@ -245,6 +245,7 @@ optional_policy(`
 
 # Use bpf tools
 allow unconfined_domain_type self:bpf { map_create map_read map_write prog_load prog_run };
+allow unconfined_domain_type self:capability2 perfmon;
```
WOnder93: This should be unnecessary thanks to:
https://github.com/fedora-selinux/selinux-policy/blob/46d834991b9d07400b9feb8aee2d85e40cfe95d7/policy/modules/system/unconfined.if#L23

zpytela: Fixed.

WOnder93:
I think we should consider creating some semantic interfaces in `kernel.if` or `misc_patterns.spt` for BPF and perf_event. For BPF it would be something like:
```
interface(`kernel_use_bpf',`
    allow $1 self:capability2 bpf;
    allow $1 self:bpf { map_create map_read map_write prog_load prog_run };
')
```

Note that for legacy socket filters only the capability is needed - we could have another interface/macro for that. Such a single-rule interface might seem a bit awkward, but it would clearly document the facility that was the reason to add the permission. I believe we should generally move more and more in the direction of semantic macros, otherwise we'll never get the policy to a maintainable state.

And for perf_event:
```
interface(`kernel_use_perf_event',`
    allow $1 self:capability2 perfmon;
    allow $1 self:perf_event open;
')
```
(The additional `perf_event` permissions could be added explicitly as needed or via further interfaces, e.g. `kernel_use_perf_event_all()`.)

zpytela:
I agree semantic macros and attributes make the policy better maintainable. Thank you for your feedback, I will get back to the suggestion later.
I actually consider kernel_write_perf_event and kernel_manage_perf_event, based on the current state in the policy.
The bpf class possibly can also make use of some bpf_all_perms.

PR #538, opened 2021-01-12
define lockdown class and access

zpytela:
This was introduced in the merge b1dba2473114588be3df916bf629a61bdcc83737 in the linux kernel.

Signed-off-by: bauen1 <j2468h@gmail.com>

Resolves: rhbz#1915184


zpytela:
Lockdown confidentiality entrie added for domains using perf_event operations.

PR #539, opened 2021-01-12
Add knot policy

mscherer:
Resubmission of https://github.com/fedora-selinux/selinux-policy-contrib/pull/365

### DIFF
--- a/policy/modules/contrib/knot.te
+++ b/policy/modules/contrib/knot.te
@@ -1,0 +1,89 @@
+policy_module(knot, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+type knot_t;
+type knot_exec_t;
+init_daemon_domain(knot_t, knot_exec_t)
+
+permissive knot_t;
+
+type knot_conf_t;
+files_type(knot_conf_t)
+
+type knot_var_lib_t;
+files_type(knot_var_lib_t)
+
+type knot_var_run_t;
+files_pid_file(knot_var_run_t)
+
+type knot_tmp_t;
+files_tmp_file(knot_tmp_t)
+
+########################################
+#
+# knot local policy
+#
+allow knot_t self:fifo_file rw_fifo_file_perms;
+allow knot_t self:unix_stream_socket create_stream_socket_perms;
+
+allow knot_t self:tcp_socket create_stream_socket_perms;
+allow knot_t self:udp_socket create_socket_perms;
+allow knot_t self:unix_dgram_socket create_socket_perms;
+
+allow knot_t self:process { setcap };
+allow knot_t self:capability { setpcap };
+
+domain_use_interactive_fds(knot_t)
+
+files_read_etc_files(knot_t)
+fs_getattr_all_fs(knot_t)
+
+miscfiles_read_localization(knot_t)
+allow knot_t knot_var_lib_t:file { map } ;
+
+corenet_all_recvfrom_netlabel(knot_t)
+corenet_tcp_sendrecv_generic_if(knot_t)
+corenet_udp_sendrecv_generic_if(knot_t)
+corenet_tcp_sendrecv_generic_node(knot_t)
+corenet_udp_sendrecv_generic_node(knot_t)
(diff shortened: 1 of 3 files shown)


WOnder93:
Hi, note that in this repo the contrib files are located in `policy/modules/contrib`, so you need to add them to that subdirectory instead of the root.

mscherer:
Oups, didn't realized :/

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+allow knot_t self:tcp_socket create_stream_socket_perms;
+allow knot_t self:udp_socket create_socket_perms;
+allow knot_t self:unix_dgram_socket create_socket_perms;
+
+allow knot_t self:process { setcap };
+allow knot_t self:capability { setpcap };
+
+domain_use_interactive_fds(knot_t)
+
+files_read_etc_files(knot_t)
```
zpytela, line 42: Was this permission requested? Domains are already allowed common access to objects with the base_type attribute.

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+fs_getattr_all_fs(knot_t)
```
zpytela, line 43: Is access to all filesystems required or just to some like tmpfs?

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+
+miscfiles_read_localization(knot_t)
+allow knot_t knot_var_lib_t:file { map } ;
```
zpytela, line 46: Move this line right after the managing files part, like it is on 72.

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+corenet_udp_bind_dns_port(knot_t)
+corenet_sendrecv_dns_server_packets(knot_t)
+corenet_tcp_connect_all_ports(knot_t)
+
+
+sysnet_read_config(knot_t)
+
+auth_use_nsswitch(knot_t)
+
+logging_send_syslog_msg(knot_t)
```
zpytela, line 67: logging is not a part of base module, so should be in the optional block in the last section.

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+files_var_lib_filetrans(knot_t, knot_var_lib_t, dir)
+
+manage_files_pattern(knot_t, knot_var_run_t, knot_var_run_t)
+manage_sock_files_pattern(knot_t, knot_var_run_t, knot_var_run_t)
+files_pid_filetrans(knot_t, knot_var_run_t, file)
+
+manage_files_pattern(knot_t, knot_tmp_t, knot_tmp_t)
+manage_dirs_pattern(knot_t, knot_tmp_t, knot_tmp_t)
+files_tmp_filetrans(knot_t, knot_tmp_t, file)
+files_tmp_filetrans(knot_t, knot_tmp_t, dir)
```
zpytela, line 86: These 2 lines can be joined together, using `{ dir file }`

policy/modules/contrib/knot.te
```diff
@@ -0,0 +1,89 @@
...
+corenet_udp_bind_generic_node(knot_t)
+corenet_tcp_bind_dns_port(knot_t)
+corenet_udp_bind_dns_port(knot_t)
+corenet_sendrecv_dns_server_packets(knot_t)
+corenet_tcp_connect_all_ports(knot_t)
+
+
+sysnet_read_config(knot_t)
+
+auth_use_nsswitch(knot_t)
```
zpytela, line 65: Rather sort base interfaces calls alphabetically.

zpytela (changes requested):

zpytela:
@mscherer I only have 2 questions and a few minor comments.

Additionally, please remove all leading and trailing blank lines as well as doubled ones.

If you are happy with testing, you can conssider removing the permissive statement.

WOnder93:
Also, please add an entry for this module to `policy/modules.conf`, so that it is built by CI.

PR #540, opened 2021-01-12
Allow sd machine userdb

zpytela:


policy/modules/system/systemd.te
```diff
@@ -423,7 +423,7 @@ init_manage_config_transient_files(systemd_machined_t)
 logging_dgram_send(systemd_machined_t)
 
 systemd_read_efivarfs(systemd_machined_t)
-systemd_create_userdbd_runtime_sock_files(systemd_machined_t)
```
WOnder93, line 426: How about also removing `systemd_create_userdbd_runtime_sock_files()`, which is unused after this change? (You can put everything into one commit so it doesn't end up looking awkward. It is really just one logical change.)

zpytela: I was pondering it and eventually decided not to remove it. The only use of this interface is for systemd-machined, but we have already backported it to f33 so there is a small chance we would break compatibility. Still, I will reconsider it.

zpytela:
Unused interface removed and commits squashed.

policy/modules/system/systemd.te
zpytela: Interface removed.

WOnder93 (approved):

PR #545, opened 2021-01-21
Label SDC(scini) Dell Driver

wrabcak:
Label custom Dell EMC PowerFlex v3.5.x modules with default label
modules_object_t, this will allo use scini driver on SELinux enabled RHEL-based SDC nodes.

More information:
https://cpsdocs.dellemc.com/bundle/VXF_DEPLOY/page/GUID-DC80A2E9-6B92-46F8-9276-1704B41E148E.html

### DIFF
--- a/policy/modules/kernel/files.fc
+++ b/policy/modules/kernel/files.fc
@@ -259,6 +259,10 @@ ifndef(`distro_redhat',`
 /usr/src/kernels/.+/lib(/.*)?	gen_context(system_u:object_r:usr_t,s0)
 ')
 
+# Deploy Dell EMC PowerFlex - SDC(scini) driver configuration
+# https://cpsdocs.dellemc.com/bundle/VXF_DEPLOY/page/GUID-DC80A2E9-6B92-46F8-9276-1704B41E148E.html
+/usr/bin/emc/scaleio/(.*)\.ko  --		gen_context(system_u:object_r:modules_object_t,s0)
+
 #
 # /var
 #


policy/modules/kernel/files.fc
```diff
@@ -259,6 +259,10 @@ ifndef(`distro_redhat',`
 /usr/src/kernels/.+/lib(/.*)?	gen_context(system_u:object_r:usr_t,s0)
 ')
 
+# Deploy Dell EMC PowerFlex - SDC(scini) driver configuration
+# https://cpsdocs.dellemc.com/bundle/VXF_DEPLOY/page/GUID-DC80A2E9-6B92-46F8-9276-1704B41E148E.html
+/usr/bin/emc/scaleio(/.*)?		gen_context(system_u:object_r:modules_object_t,s0)
```
WOnder93: Could there be also other kinds of files in this directory? The linked documentation actually suggests to only relabel `*.ko` files, which should be safer.

wrabcak: This is generic custom path used by emc, for this reason I would keep it more "open".

WOnder93: I don't understand what you mean by "open"... If you label everything as `modules_object_t`, then you are actually making it more "closed" in that you expect only kernel modules to be there. If there is e.g. some binary, some processes might be denied from executing it. And kernel modules should always have the '.ko' extension, so you don't really risk anything by only overriding the label for .ko files (especially when the vendor itself recommends it that way).

policy/modules/kernel/files.fc
```diff
@@ -259,6 +259,10 @@ ifndef(`distro_redhat',`
 /usr/src/kernels/.+/lib(/.*)?	gen_context(system_u:object_r:usr_t,s0)
 ')
 
+# Deploy Dell EMC PowerFlex - SDC(scini) driver configuration
+# https://cpsdocs.dellemc.com/bundle/VXF_DEPLOY/page/GUID-DC80A2E9-6B92-46F8-9276-1704B41E148E.html
+/usr/bin/emc/scaleio/[^/]*.ko  --		gen_context(system_u:object_r:modules_object_t,s0)
```
WOnder93: This won't match ko files that are inside subdirectories... which could be a bit risky, since the documentation uses a regex that matches also those. I think we should just use the same regex as in the document (`.../scaleio/(.*)\.ko`). Also, the dot in `.ko` needs to be escaped ;)

wrabcak: Thank you for quick feedback. I hope my latest commit should fix it. :)

WOnder93 (approved):
Looks all good now.

zpytela:
Merging, thanks.

PR #547, opened 2021-01-22
Run frr service as frr_t and add new bfd_multi port

Koncpa:
Confined frr service as frr_t and create policy files and rules for frr service.
FRRouting (FRR) is an IP routing protocol which
includes protocol daemons for BGP, IS-IS, LDP, OSPF, PIM, and RIP.
Add new port called bfd_multi for bfdd daemon, which is use in frr service.

### DIFF
--- a/policy/modules.conf
+++ b/policy/modules.conf
@@ -3092,3 +3092,10 @@ rhcd = module
 # wireguard
 #
 wireguard = module
+
+# Module: frr
+#
+# FRRouting (FRR) is an IP routing protocol suite.
+#
+frr = module
+
--- a/policy/modules/contrib/frr.fc
+++ b/policy/modules/contrib/frr.fc
@@ -1,0 +1,25 @@
+/usr/libexec/frr(/.*)?              gen_context(system_u:object_r:frr_exec_t,s0)
+
+/usr/lib/systemd/system/frr.*   gen_context(system_u:object_r:frr_unit_file_t,s0)
+
+/etc/frr(/.*)?                  gen_context(system_u:object_r:frr_conf_t,s0)
+
+/var/log/frr(/.*)?              gen_context(system_u:object_r:frr_log_t,s0)
+/var/tmp/frr(/.*)?              gen_context(system_u:object_r:frr_tmp_t,s0)
+
+/var/lock/subsys/bfdd       --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/bgpd       --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/eigrpd     --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/fabricd    --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/isisd      --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/nhrpd      --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ospf6d     --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ospfd      --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/pbrd       --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/pimd       --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ripd       --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ripngd     --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/staticd    --  gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/zebra      --  gen_context(system_u:object_r:frr_lock_t,s0)
+
+/var/run/frr(/.*)?              gen_context(system_u:object_r:frr_var_run_t,s0)
--- a/policy/modules/contrib/frr.if
+++ b/policy/modules/contrib/frr.if
@@ -1,0 +1,162 @@
+## <summary>policy for frr</summary>
+
+########################################
+## <summary>
+##	Execute frr_exec_t in the frr domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed to transition.
+## </summary>
+## </param>
+#
+interface(`frr_domtrans',`
+	gen_require(`
+		type frr_t, frr_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, frr_exec_t, frr_t)
+')
+
+######################################
+## <summary>
+##	Execute frr in the caller domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`frr_exec',`
+	gen_require(`
+		type frr_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	can_exec($1, frr_exec_t)
+')
+
+########################################
+## <summary>
+##	Read frr's log files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`frr_read_log',`
+	gen_require(`
+		type frr_log_t;
+	')
+
+	read_files_pattern($1, frr_log_t, frr_log_t)
+	optional_policy(`
+		logging_search_logs($1)
+	')
+')
+
+########################################
+## <summary>
+##	Append to frr log files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`frr_append_log',`
+	gen_require(`
+		type frr_log_t;
+	')
+
+	append_files_pattern($1, frr_log_t, frr_log_t)
+	optional_policy(`
+		logging_search_logs($1)
+	')
+')
+
+########################################
+## <summary>
+##	Manage frr log files
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`frr_manage_log',`
+	gen_require(`
+		type frr_log_t;
+	')
+
+	manage_dirs_pattern($1, frr_log_t, frr_log_t)
+	manage_files_pattern($1, frr_log_t, frr_log_t)
+	manage_lnk_files_pattern($1, frr_log_t, frr_log_t)
+	optional_policy(`
+		logging_search_logs($1)
+	')
+')
+
+########################################
+## <summary>
+##	Read frr PID files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`frr_read_pid_files',`
+	gen_require(`
+		type frr_var_run_t;
+	')
+
+	files_search_pids($1)
+	read_files_pattern($1, frr_var_run_t, frr_var_run_t)
+')
+
+########################################
+## <summary>
+##	All of the rules required to administrate
+##	an frr environment
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`frr_admin',`
+	gen_require(`
+		type frr_t;
+		type frr_log_t;
+		type frr_var_run_t;
+	')
+
+	allow $1 frr_t:process { signal_perms };
+	ps_process_pattern($1, frr_t)
+
+	tunable_policy(`deny_ptrace',`',`
+		allow $1 frr_t:process ptrace;
+	')
+
+	admin_pattern($1, frr_log_t)
+
+	files_search_pids($1)
+	admin_pattern($1, frr_var_run_t)
+	optional_policy(`
+		logging_search_logs($1)
+	')
+	optional_policy(`
+		systemd_passwd_agent_exec($1)
+		systemd_read_fifo_file_passwd_run($1)
+	')
+')
--- a/policy/modules/contrib/frr.te
+++ b/policy/modules/contrib/frr.te
@@ -1,0 +1,114 @@
+policy_module(frr, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+type frr_t;
+type frr_exec_t;
+init_daemon_domain(frr_t, frr_exec_t)
+
+type frr_log_t;
+logging_log_file(frr_log_t)
+
+type frr_tmp_t;
+files_tmp_file(frr_tmp_t)
+
+type frr_lock_t;
+files_lock_file(frr_lock_t)
+
+type frr_conf_t;
+files_config_file(frr_conf_t)
+
+type frr_unit_file_t;
+systemd_unit_file(frr_unit_file_t)
+
+type frr_var_run_t;
+files_pid_file(frr_var_run_t)
+
+########################################
+#
+# frr local policy
+#
+allow frr_t self:capability { chown dac_override dac_read_search kill net_bind_service net_raw setgid setuid };
+allow frr_t self:netlink_route_socket rw_netlink_socket_perms;
+allow frr_t self:packet_socket create;
+allow frr_t self:process { setcap setpgid };
+allow frr_t self:rawip_socket create_socket_perms;
+allow frr_t self:tcp_socket { connect connected_stream_socket_perms };
+allow frr_t self:udp_socket create_socket_perms;
+allow frr_t self:unix_stream_socket connectto;
+
+allow frr_t frr_conf_t:dir list_dir_perms;
+read_files_pattern(frr_t, frr_conf_t, frr_conf_t)
+read_lnk_files_pattern(frr_t, frr_conf_t, frr_conf_t)
+
+manage_dirs_pattern(frr_t, frr_log_t, frr_log_t)
+manage_files_pattern(frr_t, frr_log_t, frr_log_t)
+manage_lnk_files_pattern(frr_t, frr_log_t, frr_log_t)
+logging_log_filetrans(frr_t, frr_log_t, { dir file lnk_file })
+
+allow frr_t frr_tmp_t:file map;
+manage_dirs_pattern(frr_t, frr_tmp_t, frr_tmp_t)
+manage_files_pattern(frr_t, frr_tmp_t, frr_tmp_t)
+files_tmp_filetrans(frr_t, frr_tmp_t, file)
+
+manage_files_pattern(frr_t, frr_lock_t, frr_lock_t)
+manage_lnk_files_pattern(frr_t, frr_lock_t, frr_lock_t)
+files_lock_filetrans(frr_t, frr_lock_t, { file lnk_file })
+
+manage_dirs_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_lnk_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_sock_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+files_pid_filetrans(frr_t, frr_var_run_t, { dir file lnk_file })
+
+allow frr_t frr_exec_t:dir search_dir_perms;
+can_exec(frr_t, frr_exec_t)
+
+kernel_read_network_state(frr_t)
+kernel_read_net_sysctls(frr_t)
+kernel_read_system_state(frr_t)
+
+auth_use_nsswitch(frr_t)
+
+corecmd_exec_bin(frr_t)
+
+corenet_tcp_bind_appswitch_emp_port(frr_t)
+corenet_udp_bind_bfd_control_port(frr_t)
+corenet_udp_bind_bfd_echo_port(frr_t)
+corenet_udp_bind_bfd_multi_port(frr_t)
+corenet_tcp_bind_bgp_port(frr_t)
+corenet_tcp_bind_cmadmin_port(frr_t)
+corenet_udp_bind_cmadmin_port(frr_t)
+corenet_tcp_bind_firepower_port(frr_t)
+corenet_tcp_bind_priority_e_com_port(frr_t)
+corenet_udp_bind_router_port(frr_t)
+corenet_tcp_bind_qpasa_agent_port(frr_t)
+corenet_tcp_bind_smntubootstrap_port(frr_t)
+corenet_tcp_bind_versa_tek_port(frr_t)
+corenet_tcp_bind_zebra_port(frr_t)
+
+domain_use_interactive_fds(frr_t)
+
+fs_read_nsfs_files(frr_t)
+
+sysnet_exec_ifconfig(frr_t)
+
+userdom_read_admin_home_files(frr_t)
+
+optional_policy(`
+	logging_send_syslog_msg(frr_t)
+')
+
+optional_policy(`
+	modutils_exec_kmod(frr_t)
+	modutils_getattr_module_deps(frr_t)
+	modutils_read_module_config(frr_t)
+	modutils_read_module_deps_files(frr_t)
+')
+
+optional_policy(`
+	networkmanager_read_state(frr_t)
+')


policy/modules/contrib/frr.if
```diff
@@ -0,0 +1,160 @@
...
+		type frr_log_t;
+		type frr_var_run_t;
+	')
+
+	allow $1 frr_t:process { signal_perms };
+	ps_process_pattern($1, frr_t)
+
+    tunable_policy(`deny_ptrace',`',`
+        allow $1 frr_t:process ptrace;
+    ')
```
WOnder93: There is some whitespace inconsistency here.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,81 @@
...
+systemd_unit_file(frr_unit_file_t)
+
+type frr_var_run_t;
+files_pid_file(frr_var_run_t)
+
+########################################
+#
+# frr local policy
+#
+allow frr_t self:capability { dac_override kill net_bind_service net_raw setgid setuid };
```
WOnder93: Did you investigate why frr needs the `dac_override` capability? Sometimes it's caused by wrong permissions being set on files.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,81 @@
...
+
+corenet_tcp_bind_bgp_port(frr_t)
+corenet_tcp_bind_zebra_port(frr_t)
+corenet_udp_bind_router_port(frr_t)
+
+auth_use_nsswitch(frr_t)
+
+domain_use_interactive_fds(frr_t)
+
+files_manage_generic_locks(frr_t)
```
WOnder93: What exact lock file is frr trying to access? If it's its own lock file, then perhaps it should be labeled as `frr_lock_t`, avoiding the need for this generic rule?

policy/modules/contrib/frr.if
Koncpa: This part of policy I generate via sepolicy generate, so It's possible that here have appeared some whitespace inconsistency. Thanks, I fix this.

policy/modules/contrib/frr.te
Koncpa: Yes, I investigate this issue, the frr need delete pid file **/var/run/frr/bgpd.pid** so need bypass ownership of the **rm** to delete this file. If you want I can send you full audit message.

policy/modules/contrib/frr.te
Koncpa: Frr trying access to lock files like **/var/lock/susbys/bgpd**. This lock files are related with protocols, which frr use. I think, that we are not able to label dir /varl/lock/subsys as frr_lock_t, because dir is not use only for frr lock files, but It's possible to label each lock file which use frr or what do you think?

Koncpa: I tried to label each lock file and test case with these changes was succesfull without any fail and AVC's, so I update this PR.

policy/modules/contrib/frr.te
WOnder93: Looking at the daemons the service starts, it might indeed be unavoidable if the file is deleted by `watchfrr` (which runs as root) instead of `bgpd` itself. Please paste here the full audit record so that my curiosity is satisfied ;)

policy/modules/contrib/frr.te
WOnder93: The `files_manage_generic_locks(frr_t)` is still there in the last version - can you try with it removed?

policy/modules/contrib/frr.te
Koncpa: Here you have :)
`type=PROCTITLE msg=audit(01/21/2021 09:15:43.987:358) : proctitle=rm -f /var/run/frr/bgpd.pid
type=PATH msg=audit(01/21/2021 09:15:43.987:358) : item=1 name=/var/run/frr/bgpd.pid inode=33039 dev=00:18 mode=file,644 ouid=frr ogid=frr rdev=00:00 obj=system_u:object_r:frr_var_run_t:s0 nametype=DELETE cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=PATH msg=audit(01/21/2021 09:15:43.987:358) : item=0 name=/var/run/frr/ inode=27604 dev=00:18 mode=dir,755 ouid=frr ogid=frr rdev=00:00 obj=system_u:object_r:frr_var_run_t:s0 nametype=PARENT cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(01/21/2021 09:15:43.987:358) : cwd=/
type=SYSCALL msg=audit(01/21/2021 09:15:43.987:358) : arch=x86_64 syscall=unlinkat success=yes exit=0 a0=0xffffff9c a1=0x55614f7d0640 a2=0x0 a3=0x0 items=2 ppid=4755 pid=4780 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=rm exe=/usr/bin/rm subj=system_u:system_r:frr_t:s0 key=(null)
type=AVC msg=audit(01/21/2021 09:15:43.987:358) : avc:  denied  { dac_override } for  pid=4780 comm=rm capability=dac_override  scontext=system_u:system_r:frr_t:s0 tcontext=system_u:system_r:frr_t:s0 tclass=capability permissive=1
`

policy/modules/contrib/frr.te
Koncpa: Oh sorry, I forgot to delete it in PR version. Now It should be everything updated.

policy/modules/contrib/frr.te
WOnder93: Yes, so it's a root process (presumably watchfrr) removing a pid file owned by user frr... in that case the dac_override is justified.

WOnder93 (approved):
Looks good to my untrained eye now, although @zpytela might have further comments.

Koncpa:
Add another permissions for frr policy, which were discover during testing a common selinux policy tests. The build test will fail, because in policy has added interface, which isn't yet in policy, but PR has already made: https://github.com/fedora-selinux/selinux-policy/pull/556

WOnder93:
> The build test will fail, because in policy has added interface, which isn't yet in policy, but PR has already made: #556

In that case why not just include the commit in this PR instead of opening a separate one?

policy/modules.conf
```diff
@@ -3064,3 +3064,11 @@ rrdcached = module
 # stratisd
 #
 stratisd = module
+
+# Layer: contrib
+# Module: frr
+#
+# frr
```
zpytela: Abbreviations usually deserve a few explanatory words.

policy/modules/contrib/frr.fc
```diff
@@ -0,0 +1,25 @@
+/usr/lib/frr(/.*)?              gen_context(system_u:object_r:frr_exec_t,s0)
+
+/usr/lib/systemd/system/frr.*   gen_context(system_u:object_r:frr_unit_file_t,s0)
+
+/etc/frr(/.*)?                  gen_context(system_u:object_r:frr_conf_t,s0)
+
+/var/log/frr(/.*)?              gen_context(system_u:object_r:frr_log_t,s0)
+/var/tmp/frr(/.*)?             gen_context(system_u:object_r:frr_tmp_t,s0)
```
zpytela: Misaligned line.

policy/modules/contrib/frr.fc
```diff
@@ -0,0 +1,25 @@
...
+/var/lock/subsys/isisd          gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/nhrpd          gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ospf6d         gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ospfd          gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/pbrd           gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/pimd           gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ripd           gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/ripngd         gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/staticd        gen_context(system_u:object_r:frr_lock_t,s0)
+/var/lock/subsys/zebra          gen_context(system_u:object_r:frr_lock_t,s0)
```
zpytela: Are all of the subsys locks plain files?

policy/modules/contrib/frr.if
```diff
@@ -0,0 +1,160 @@
...
+## </param>
+#
+interface(`frr_exec',`
+	gen_require(`
+		type frr_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	can_exec($1, frr_exec_t)
+')
```
zpytela, line 39: Blank line separator missing.

policy/modules/contrib/frr.if
```diff
@@ -0,0 +1,160 @@
...
+#
+interface(`frr_read_pid_files',`
+	gen_require(`
+		type frr_var_run_t;
+	')
+
+	files_search_pids($1)
+	read_files_pattern($1, frr_var_run_t, frr_var_run_t)
+')
+
```
zpytela, line 125: Double line separator.

policy/modules/contrib/frr.if
```diff
@@ -0,0 +1,160 @@
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
```
zpytela: $2 is never used in the interface. Are there any plans for using it?

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+files_tmp_file(frr_tmp_t)
+
+type frr_lock_t;
+files_lock_file(frr_lock_t)
+
+type frr_conf_t;
+files_config_file(frr_conf_t)
+
+type frr_unit_file_t;
+systemd_unit_file(frr_unit_file_t)
```
zpytela, line 25: The private unit file type is not used anywhere else, is it supposed to happen so?

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+
+type frr_var_run_t;
+files_pid_file(frr_var_run_t)
+
+########################################
+#
+# frr local policy
+#
+allow frr_t self:capability { dac_override dac_read_search chown kill net_bind_service net_raw setgid setuid };
```
zpytela: I cannot find justification for dac_override/dac_read_search in this PR. Move chown in front of dac_.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+corenet_udp_bind_router_port(frr_t)
+corenet_udp_bind_cmadmin_port(frr_t)
+
+fs_read_nsfs_files(frr_t)
+
+auth_use_nsswitch(frr_t)
+
+domain_use_interactive_fds(frr_t)
+
+logging_send_syslog_msg(frr_t)
```
zpytela: logging is not a part of base module, needs optional block. This comment also applies to usage in interfaces.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+corenet_tcp_bind_qpasa_agent_port(frr_t)
+corenet_tcp_bind_smntubootstrap_port(frr_t)
+corenet_tcp_bind_versa_tek_port(frr_t)
+corenet_udp_bind_bfd_control_port(frr_t)
+corenet_udp_bind_router_port(frr_t)
+corenet_udp_bind_cmadmin_port(frr_t)
+
+fs_read_nsfs_files(frr_t)
+
+auth_use_nsswitch(frr_t)
```
zpytela: First block of interfaces calls is for kernel, then it should be sorted alphabetically.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+corenet_tcp_bind_zebra_port(frr_t)
+corenet_tcp_bind_appswitch_emp_port(frr_t)
+corenet_tcp_bind_firepower_port(frr_t)
+corenet_tcp_bind_priority_e_com_port(frr_t)
+corenet_tcp_bind_qpasa_agent_port(frr_t)
+corenet_tcp_bind_smntubootstrap_port(frr_t)
+corenet_tcp_bind_versa_tek_port(frr_t)
+corenet_udp_bind_bfd_control_port(frr_t)
+corenet_udp_bind_router_port(frr_t)
+corenet_udp_bind_cmadmin_port(frr_t)
```
zpytela: Does this order follow some rule like port numbers? I'd rather see them alphabetically sorted.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+
+sysnet_exec_ifconfig(frr_t)
+
+userdom_read_admin_home_files(frr_t)
+
+optional_policy(`
+	modutils_read_module_config(frr_t)
+	modutils_exec_kmod(frr_t)
+	modutils_getattr_module_deps(frr_t)
+	modutils_read_module_deps_files(frr_t)
```
zpytela, line 109: Can this list also be sorted?

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+# frr local policy
+#
+allow frr_t self:capability { dac_override dac_read_search chown kill net_bind_service net_raw setgid setuid };
+allow frr_t self:process { setcap setpgid };
+allow frr_t self:packet_socket create;
+allow frr_t self:netlink_route_socket rw_netlink_socket_perms;
+allow frr_t self:tcp_socket { connect connected_stream_socket_perms };
+allow frr_t self:udp_socket create_socket_perms;
+allow frr_t self:rawip_socket create_socket_perms;
+allow frr_t self:unix_stream_socket connectto;
```
zpytela, line 41: Try to sort alphabetically whenever possible.

policy/modules/contrib/frr.te
```diff
@@ -0,0 +1,108 @@
...
+files_lock_filetrans(frr_t, frr_lock_t, { file lnk_file })
+
+manage_dirs_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_lnk_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+manage_sock_files_pattern(frr_t, frr_var_run_t, frr_var_run_t)
+files_pid_filetrans(frr_t, frr_var_run_t, { dir file lnk_file })
+
+allow frr_t frr_exec_t:dir search_dir_perms;
+read_lnk_files_pattern(frr_t, frr_exec_t, frr_exec_t)
```
zpytela: Is there an example for executable symlink?

zpytela (changes requested):

policy/modules.conf
Koncpa: Fixed.

policy/modules/contrib/frr.fc
Koncpa: Fixed.

policy/modules/contrib/frr.fc
Koncpa: Yes, It was tested.

policy/modules/contrib/frr.if
Koncpa: Fixed.

policy/modules/contrib/frr.if
Koncpa: Fixed.

policy/modules/contrib/frr.if
Koncpa: It's mistake of sepolicy generate. Now It should be fixed.

policy/modules/contrib/frr.te
Koncpa: @milosmalik  wrote about acceptance criteria of frr in bugzilla ticket, so I added this. And also when I look to /usr/lib/systemd/system/ is here frr.service file which is labeled as frr_unit_file_t.  But when was tested frr policy I didn find any AVC relating with frr_unit_file_t. So what do you think? Can I add some permission for frr_unit_file_t?

policy/modules/contrib/frr.te
Koncpa: About dac_override I wrote in Ondra review, and about dac_read_search the vtysh need boot startup configuration with specific daemon, so I suppose that vtysh need read frr.conf .

Here I attach a example of full audit message about dac_read_search :
`type=PROCTITLE msg=audit(01/29/2021 10:00:36.033:3841) : proctitle=/usr/bin/vtysh -b -d zebra·
type=PATH msg=audit(01/29/2021 10:00:36.033:3841) : item=0 name=/etc/frr/frr.conf inode=262500 dev=fc:01 mode=file,640 ouid=frr ogid=frr rdev=00:00 obj=unconfined_u:object_r:frr_conf_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0·
type=CWD msg=audit(01/29/2021 10:00:36.033:3841) : cwd=/·
type=SYSCALL msg=audit(01/29/2021 10:00:36.033:3841) : arch=x86_64 syscall=openat success=yes exit=9 a0=0xffffff9c a1=0x5587f0b20780 a2=O_RDONLY a3=0x0 items=1 ppid=81191 pid=81243 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=vtysh exe=/usr/bin/vtysh subj=system_u:system_r:frr_t:s0 key=(null)·
type=AVC msg=audit(01/29/2021 10:00:36.033:3841) : avc:  denied  { dac_read_search } for  pid=81243 comm=vtysh capability=dac_read_search  scontext=system_u:system_r:frr_t:s0 tcontext=system_u:system_r:frr_t:s0 tclass=capability permissive=0·
type=AVC msg=audit(01/29/2021 10:00:36.033:3841) : avc:  denied  { dac_read_search } for  pid=81243 comm=vtysh capability=dac_read_search  scontext=system_u:system_r:frr_t:s0 tcontext=system_u:system_r:frr_t:s0 tclass=capability permissive=0`

policy/modules/contrib/frr.te
Koncpa: Fixed.

policy/modules/contrib/frr.te
Koncpa: Fixed.

policy/modules/contrib/frr.te
Koncpa: Fixed.

policy/modules/contrib/frr.te
Koncpa: Fixed.

policy/modules/contrib/frr.te
Koncpa: Fixed.

policy/modules/contrib/frr.te
Koncpa: Yes, here you have example. 

Example : `type=PROCTITLE msg=audit(02/02/2021 06:02:14.242:651) : proctitle=/usr/lib/frr/fabricd -d -A 127.0.0.1 
type=PATH msg=audit(02/02/2021 06:02:14.242:651) : item=0 name=/usr/lib64/frr/libfrr.so.0 nametype=UNKNOWN cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0 
type=CWD msg=audit(02/02/2021 06:02:14.242:651) : cwd=/etc/frr 
type=SYSCALL msg=audit(02/02/2021 06:02:14.242:651) : arch=x86_64 syscall=openat success=no exit=EACCES(Permission denied) a0=0xffffff9c a1=0x7ffcae5eadb0 a2=O_RDONLY|O_CLOEXEC a3=0x0 items=1 ppid=35891 pid=35892 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=fabricd exe=/usr/lib/frr/fabricd subj=system_u:system_r:frr_t:s0 key=(null) 
type=AVC msg=audit(02/02/2021 06:02:14.242:651) : avc:  denied  { read } for  pid=35892 comm=fabricd name=libfrr.so.0 dev="vda1" ino=35651802 scontext=system_u:system_r:frr_t:s0 tcontext=system_u:object_r:frr_exec_t:s0 tclass=lnk_file permissive=0 
`

policy/modules/contrib/frr.fc
zpytela: For plain files the `--` spec can be used.

policy/modules/contrib/frr.te
zpytela: Thanks, I managed to find the previous discussion.

zpytela: Can you post the output of this command?
```
ls -laZ /usr/lib64/frr
```

WOnder93: This seems like a packaging bug in `frr` - the executable files currently in `/usr/lib/frr` should be placed in `/usr/libexec/frr` instead. Did you report it to the maintainer/upstream?

zpytela:
@koncpa, do you want to keep the 2 commit separate, like to distinguish between the initial policy and later enhancements, or will it make sense for you to squash them?

policy/modules/contrib/frr.te
Koncpa: Here:

- drwxr-xr-x.  3 root root system_u:object_r:frr_exec_t:s0     113 Feb  2 04:42 .
- dr-xr-xr-x. 49 root root system_u:object_r:lib_t:s0        24576 Feb  2 04:42 ..
- lrwxrwxrwx.  1 root root system_u:object_r:frr_exec_t:s0      19 Sep  3  2019 libfrrsnmp.so.0 -> libfrrsnmp.so.0.0.0
- -rwxr-xr-x.  1 root root system_u:object_r:frr_exec_t:s0   20792 Sep  3  2019 libfrrsnmp.so.0.0.0
- lrwxrwxrwx.  1 root root system_u:object_r:frr_exec_t:s0      15 Sep  3  2019 libfrr.so.0 -> libfrr.so.0.0.0
- -rwxr-xr-x.  1 root root system_u:object_r:frr_exec_t:s0 1200856 Sep  3  2019 libfrr.so.0.0.0
- drwxr-xr-x.  2 root root system_u:object_r:frr_exec_t:s0     151 Feb  2 04:42 modules

zpytela: This does not conform to the fc file. What do restorecon/matchpathcon say?
```
restorecon -Rvn /usr/lib64/frr
matchpathcon /usr/lib64/frr/libfrr.so.0 
```

Koncpa:
> @Koncpa, do you want to keep the 2 commit separate, like to distinguish between the initial policy and later enhancements, or will it make sense for you to squash them?

What do you think that will be better? I made two commits because on previous commit made @WOnder93 review so I want distinguish later enchancements for next review.

policy/modules/contrib/frr.te
WOnder93: The problem is that in .fc there is `/usr/lib/frr(/.*)?              gen_context(system_u:object_r:frr_exec_t,s0)`, which breaks the labels in `/usr/lib64/frr/...`, since there is an [equivalency](https://src.fedoraproject.org/rpms/selinux-policy/blob/master/f/file_contexts.subs_dist#_8) between `/usr/lib` and `/usr/lib64`.

Koncpa: @zpytela , matchpathcon say:
`/usr/lib64/frr/libfrr.so.0	system_u:object_r:frr_exec_t:s0`

Koncpa: > This seems like a packaging bug in `frr` - the executable files currently in `/usr/lib/frr` should be placed in `/usr/libexec/frr` instead. Did you report it to the maintainer/upstream?

No didn't report that.

Koncpa: > The problem is that in .fc there is `/usr/lib/frr(/.*)? gen_context(system_u:object_r:frr_exec_t,s0)`, which breaks the labels in `/usr/lib64/frr/...`, since there is an [equivalency](https://src.fedoraproject.org/rpms/selinux-policy/blob/master/f/file_contexts.subs_dist#_8) between `/usr/lib` and `/usr/lib64`.

So issue is that this files should be in `/usr/libexec/frr` a also have different label?

zpytela: Executable files which are not supposed to be executed from cmdline should be in /usr/libexec. Shared libraries should be in /usr/lib64 and have the common type lib_t unless there was a reason for having a private type - this matches the purpose of shared libraries.

Koncpa: @WOnder93 So I have already reported this bug to frr maintainer.

Koncpa:
Squash two commits into one.

vmojzis:
@Koncpa Could you please rebase the PR to incoporate https://github.com/fedora-selinux/selinux-policy/pull/556/commits/abf7fa21fa5bb42e773f1255cd724745b05e79d3 ?

policy/modules/contrib/frr.if
```diff
@@ -0,0 +1,161 @@
...
+########################################
+## <summary>
+##	All of the rules required to administrate
+##	an frr environment
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+#
```
WOnder93, line 136: The `<param>` XML tag is left unclosed here, which is breaking the build.

Koncpa:
> @Koncpa Could you please rebase the PR to incoporate [abf7fa2](https://github.com/fedora-selinux/selinux-policy/commit/abf7fa21fa5bb42e773f1255cd724745b05e79d3) ?

Can I rebase PR to incorporate commit in this way?

policy/modules/contrib/frr.if
Koncpa: Thanks Ondřej. :)

policy/modules/contrib/frr.te
zpytela: What I had in mind was: By default the unit file gets the shared `systemd_unit_file_t` type. It is correct as long it is not e. g. required for the application to manage the service, then it is the time for a confined type. I haven't seen such a rule in the policy, so I think it was not necessary; not incorrect though, it may be required later.

zpytela:
Now it generally looks good.

zpytela (approved):

WOnder93 (approved):

PR #548, opened 2021-01-22
Dontaudit vhostmd to write in /var/lib/rpm/ dir and allow signull rpm

Koncpa:
Add rpm_dontaudit_manage_db() interface to do not audit attempts to create, read,
write, and delete the RPM package database for vhostmd_t domain.

Add interface rpm_signull() to send a null signal to rpm for vhostmd_t
domain.

### DIFF
--- a/policy/modules/contrib/vhostmd.te
+++ b/policy/modules/contrib/vhostmd.te
@@ -70,6 +70,10 @@ optional_policy(`
 optional_policy(`
 	rpm_exec(vhostmd_t)
 	rpm_read_db(vhostmd_t)
+	#vhostmd work properly without allowing this permissions, so dontaudit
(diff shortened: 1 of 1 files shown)


WOnder93:
Can you explain why you chose to dontaudit those access vectors? (It should be explained in the commit message and, if possible, also in a short comment in the code, since it doesn't seem obvious, at least to me :)

Koncpa:
> Can you explain why you chose to dontaudit those access vectors? (It should be explained in the commit message and, if possible, also in a short comment in the code, since it doesn't seem obvious, at least to me :)

I tried to explain more in commit message also in code, so I hope that now is more obvious. :)

zpytela:
@koncpa thanks, merging.

PR #549, opened 2021-01-22
ci: Install build dependencies from koji

WOnder93:
The common rawhide composes may not contain the latest packages
referenced from the specfile. To reduce this window, use rawhide
packages from koji.


zpytela:
Merging, thanks.

PR #551, opened 2021-01-25
Allow domain write to systemd-resolved PID socket files

zpytela:
Previously, the permission was allowed for the nsswitch_domain
attribute which turned out not to be sufficient.


WOnder93:
Could it be that geoclue_t [1] should be added to nsswitch_domain instead?

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1900175

zpytela:
This was discusses with systemd developers, the request is considered legit for all domains.

PR #558, opened 2021-02-01
Allow rhsmcertd_t read kpatch lib files

zpytela:
Resolves: rhbz#1895322

### DIFF
--- a/policy/modules/contrib/kpatch.if
+++ b/policy/modules/contrib/kpatch.if
@@ -41,6 +41,26 @@ interface(`kpatch_nnp_domtrans',`
 
 ########################################
 ## <summary>
+##	Read kpatch lib files
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed access
+## </summary>
+## </param>
+#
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/kpatch.if
```diff
@@ -39,6 +39,26 @@ interface(`kpatch_nnp_domtrans',`
...
+## </summary>
+## </param>
+#
+interface(`kpatch_read_lib_files',`
+	gen_require(`
+		type kpatch_var_lib_t;
+	')
+
+	files_search_var_lib($1)
+	allow $1 kpatch_var_lib_t:dir list_dir_perms;
```
WOnder93: Maybe you could use `list_dirs_pattern($1, kpatch_var_lib_t, kpatch_var_lib_t)` instead, to be consistent with the line below? The resulting permission set should be the same.

zpytela: Surely I can. I did not do so because of the duplicate definition inside list_dirs_pattern.

PR #560, opened 2021-02-02
Allow systemd-importd create files in /run/systemd

zpytela:


WOnder93:
I assume this is related to (internal) https://bugzilla.redhat.com/show_bug.cgi?id=1788055, where the file being created is `/run/systemd/machines.lock`, which should be labeled `systemd_machined_var_run_t` according to the .fc [1]. So I think this should be fixed via `init_named_pid_filetrans(..., "machines.lock")` (and the same interface call should be added for `systemd_machined_t` itself, too).

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/systemd.fc#L91

zpytela:
There actually was no fc entry for machines.lock, but it looks more reasonable so I changed the commit.

PR #562, opened 2021-02-04
   Allow unconfined_t and kprop_t to create krb5_0.rcache2 with the right context

Koncpa:
In specific scenarios, processed labeled unconfined_t or kprop_t
may create files with a wrong file context.
Use an interface kerberos_tmp_filetrans_host_rcache() in case of krop_t
which allow kerberos content in the /var/tmp with an correct label.
Add an interface kerberos_tmp_filetrans_host_rcache() to kerberos_filetrans_named_content()
which allow kerberos content in the /var/tmp with an correct label. This
solution in this case avoid duplication of the magic filename in the
future. Then unconfined and kprop can create /var/tmp/krb5_0.rcache2
with proper selinux context.

Bugzillas:
https://bugzilla.redhat.com/show_bug.cgi?id=1874527
https://bugzilla.redhat.com/show_bug.cgi?id=1877044

### DIFF
--- a/policy/modules/contrib/kerberos.if
+++ b/policy/modules/contrib/kerberos.if
@@ -589,6 +589,7 @@ interface(`kerberos_filetrans_named_content',`
 	kerberos_tmp_filetrans_host_rcache($1, "HTTP_23")
 	kerberos_tmp_filetrans_host_rcache($1, "HTTP_48")
 	kerberos_tmp_filetrans_host_rcache($1, "imap_0")
+	kerberos_tmp_filetrans_host_rcache($1, "krb5_0.rcache2")
 	kerberos_tmp_filetrans_host_rcache($1, "nfs_0")
 	kerberos_tmp_filetrans_host_rcache($1, "ldapmap1_0")
 	kerberos_tmp_filetrans_host_rcache($1, "ldap_487")
--- a/policy/modules/contrib/kerberos.te
+++ b/policy/modules/contrib/kerberos.te
@@ -397,4 +397,5 @@ seutil_read_file_contexts(kpropd_t)
 
 sysnet_dns_name_resolve(kpropd_t)
 
+kerberos_tmp_filetrans_host_rcache(kpropd_t, "krb5_0.rcache2")
 kerberos_use(kpropd_t)


zpytela:
Build fails as the old setting is still used.
```
+ DISTGIT_REF=master
```

policy/modules/contrib/kerberos.if
```diff
@@ -595,6 +595,25 @@ interface(`kerberos_filetrans_named_content',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`kerberos_filetrans_tmp_content',`
+	gen_require(`
+		type krb5_host_rcache_t;
+	')
+
+files_tmp_filetrans($1, krb5_host_rcache_t, file, "krb5_0.rcache2")
```
WOnder93: Please indent this line properly.

WOnder93 (changes requested):
And also please use imperative in the commit subject.

policy/modules/contrib/kerberos.if
Koncpa: Now it should be fixed.

WOnder93:
Sorry for being a grammar nazi again, but there are still mistakes in the commit message.
1. Subject line: should be "Allow X to Y" (where Y is an infinitive) and there should be "the" before "right context"
2. The first sentence has wrong word order. Should be: "In specific scenarios, processed labeled unconfined_t or kprop_t may create files with a wrong file context."
3. The second sentence has wrong/missing; should be : "Create an interface" and "with the correct label"
4. "tmp directory" should be "/var/tmp" and "var/tmp/krb5_0.rcache2" is missing the leading slash.

Koncpa:
> Sorry for being a grammar nazi again, but there are still mistakes in the commit message.
> 
>     1. Subject line: should be "Allow X to Y" (where Y is an infinitive) and there should be "the" before "right context"
> 
>     2. The first sentence has wrong word order. Should be: "In specific scenarios, processed labeled unconfined_t or kprop_t may create files with a wrong file context."
> 
>     3. The second sentence has wrong/missing; should be : "Create an interface" and "with the correct label"
> 
>     4. "tmp directory" should be "/var/tmp" and "var/tmp/krb5_0.rcache2" is missing the leading slash.

Np, I like hear feedback. It pushes me forward. :) Now it should be fixed.

Koncpa:
@zpytela find existing interface, which allow similar rules, so it's not necessary to create new one.

policy/modules/roles/unconfineduser.te
```diff
@@ -158,6 +158,10 @@ optional_policy(`
         kpatch_run(unconfined_t,unconfined_r)
     ')
 
+	optional_policy(`
+		kerberos_tmp_filetrans_host_rcache(unconfined_t, "krb5_0.rcache2")
+	')
+
```
WOnder93: Maybe you should rather add the "krb5_0.rcache2" entry to `kerberos_filetrans_named_content()` so that all unconfined domains get this transition (a lot of domains then get this transition, but it could avoid some further duplication of the magic filename in the future).

Koncpa: Yes, sure :)

policy/modules/roles/unconfineduser.te
```diff
@@ -158,6 +158,10 @@ optional_policy(`
         kpatch_run(unconfined_t,unconfined_r)
     ')
 
+	optional_policy(`
+		kerberos_filetrans_named_content(unconfined_t)
+	')
+
```
WOnder93: You can now drop this part, since `unconfined_t` belongs to `named_filetrans_domain`, and there already is `kerberos_filetrans_named_content(named_filetrans_domain)` in policy/modules/kernel/domain.te.

WOnder93 (approved):

zpytela:
LGTM, too, thanks.

PR #563, opened 2021-02-04
Update .copr/make-srpm.sh to use rawhide as DISTGIT_BRANCH

zpytela:
In the src.fedoraproject.org/rpms/selinux-policy packages repository,
the default branch name has changed to "rawhide".
The .copr/make-srpm.sh file was updated to use "rawhide" in the
DISTGIT_REF variable instead of "master".

### DIFF
--- a/.copr/make-srpm.sh
+++ b/.copr/make-srpm.sh
@@ -7,7 +7,7 @@ outdir="$1"; shift
 dirname="$(dirname "$0")"
 
 DISTGIT_URL=https://src.fedoraproject.org/rpms/selinux-policy
-DISTGIT_REF=master
+DISTGIT_REF=rawhide
 
 CONTAINER_URL=https://github.com/containers/container-selinux
 EXPANDER_URL=https://github.com/fedora-selinux/macro-expander


zpytela:
There seems to be something wrong in the build process:
```
policy/modules/roles/unconfineduser.if:177: Error: duplicate definition of unconfined_stub_role(). Original definition on 2442.

policy/modules/system/unconfined.if:239: Error: duplicate definition of unconfined_server_dbus_chat(). Original definition on 2320.
```

WOnder93:
Hm, I don't understand what's going on there... Let me try to re-run the CI just in case it is something transient....

zpytela:
Something must have happened yesterday, #561 suffers from the same issue while #560 is okay.

zpytela:
Today the package builds correctly.

WOnder93 (approved):
Great :) Just for future record, the build failure was caused by https://github.com/containers/container-selinux/pull/117, which is now reverted:
https://github.com/containers/container-selinux/commit/e502a14af67affa6b8fc20e54c2981a983e02f65

PR #565, opened 2021-02-07
Allow systemd and systemd-resolved watch dbus pid object

zpytela:
The init_t domain was allowed to read and watch pid socket files.
The systemd_resolved_t domain was allowed to:
- watch root dirs
- watch /var/run dirs
- watch pid socket files and directories
The following interfaces were added:
- files_watch_root_dirs(systemd_resolved_t)
- files_watch_var_run_dirs(systemd_resolved_t)

### DIFF
--- a/policy/modules/contrib/dbus.if
+++ b/policy/modules/contrib/dbus.if
@@ -544,6 +544,44 @@ interface(`dbus_read_pid_sock_files',`
 
 ########################################
 ## <summary>
+##	Watch system dbus pid socket files
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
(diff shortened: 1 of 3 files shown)


WOnder93 (approved):
I'm a little confused by the inconsistencies here, but I can;t see deep enough to suggest how to improve it.

Though I would at least suggest to reorder the dbus_* interface calls a little so that the _pid_ ones are together (and in consistent order).

zpytela:
I moved the interfaces calls one line up. Not sure though what you mean by inconsistencies, @WOnder93.

It is not easy to keep interface coherent and comprehensible when they provide too complex policy or have a long history.

PR #567, opened 2021-02-09
Allow dmidecode_t integrity lockdown permission

zpytela:
Allow biosdecode, ownership, vpddecode the lockdown permission
from the integrity class.
This permission is required to open /dev/mem, /dev/kmem, and /dev/port.

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -3163,6 +3163,7 @@ interface(`dev_read_raw_memory',`
 
 	read_chr_files_pattern($1, device_t, memory_device_t)
 	allow $1 memory_device_t:chr_file map;
+	allow $1 self:lockdown integrity;
 
 	allow $1 self:capability sys_rawio;
 	typeattribute $1 memory_raw_read;


WOnder93:
How about adding it directly to `dev_read_raw_memory()`?

zpytela:
It makes sense to me, too; but I ran a few tests for domains in memory_raw_read and memory_raw_write attributes, like colord or mcelog, and the permission was not requested.

WOnder93:
They may just not access those files under normal operation. You'd have to try it with `dev_read_raw_memory()` calls removed to see if the tests really trigger it. The permission will be always requested when a process tries to open one of these files [1], so I'd strongly prefer to add it to the interface.

[1] https://elixir.bootlin.com/linux/latest/source/drivers/char/mem.c#L882 - note that `open_port` is aliased as `open_mem` and `open_kmem` a few lines below

zpytela:
I changed the commit in the indicated way and overhauled the commit message (so it is no 30 times longer than the commit itself).

WOnder93 (approved):

PR #569, opened 2021-02-09
Allow kmod_t and udev_t confidentiality lockdown permission

zpytela:
The confidentiality permission from the lockdown class is required
for modprobe and systemd-udevd to use tracefs.

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -248,6 +248,8 @@ allow unconfined_domain_type self:bpf { map_create map_read map_write prog_load 
 
 allow unconfined_domain_type self:lnk_file setattr;
 
+allow unconfined_domain_type self:lockdown confidentiality;
+
 allow unconfined_domain_type self:perf_event manage_perf_event_perms;
 
 # Use/sendto/connectto sockets created by any domain.
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -6894,6 +6894,7 @@ interface(`fs_rw_tracefs_files',`
 	')
 
 	rw_files_pattern($1, tracefs_t, tracefs_t)
+	allow $1 self:lockdown confidentiality;
 ')
 
 ########################################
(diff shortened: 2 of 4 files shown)


WOnder93:
Also here I would prefer to add it to `fs_rw_tracefs_files()` and possibly also to `fs_manage_tracefs_dirs()`. These domains currently don't require these interfaces, since they are part of `files[ystem]_unconfined_type`, but it can't hurt to be explicit and add the `fs_rw_tracefs_files()` call to them.

zpytela:
I see the point in adding it into `fs_rw_tracefs_files()`. It does not solve the problem for kmod or udev though, it needs to be added to I guess `devices_unconfined_type` - the new classes/capabilities are not allowed automatically.

zpytela:
I changed the commit to allow the permission to `devices_unconfined_type`: it is true it is only called from the `unconfined_domain` interface, but it seems to fit semantically here. I added it to the tracefs interfaces, too, and made kmod/udev call `fs_rw_tracefs_files()`.

WOnder93:
Didn't you mean to add it to `filesystem_unconfined_type` instead? `/dev/kmem` and friends are in another PR (and there `integrity` is needed, not `confidentiality`.

Either way, I'd rather we add both permissions to `unconfined_domain_type` directly, since they cover various different things, not just device or filesystem related.

As a side note, given that `unconfined_domain_noaudit()` is the only caller of `fs_unconfined()` and `filesystem_unconfined()`, I wonder if we shouldn't remove the `files[ystem]_unconfined_type` attributes completely (granting the permissions directly in the interfaces) and just call the interfaces directly on `unconfined_domain_type`. Right now, the attributes are pretty much useless and just slow down access vector computation. (But this is of course for a separate PR, ideally starting at refpolicy...)

zpytela:
You are right, I worked on 2 commits at the same time, funny enough I noticed the commit message is incorrect, but not the commit content. Hope now it is consistent.

WOnder93 (approved):
Yes, it looks good now, thanks!

PR #572, opened 2021-02-11
Allow avahi watch etc dirs

zpytela:


WOnder93:
Not sure if it's worth it, but have you considered labeling the /etc/avahi subtree as avahi_conf_t? (If it's not, I'd suggest adding a comment that only /etc/avahi is known to be watched in case someone would want to do that in the future. So they don't need to worry about the other etc_t files.)

zpytela:
It is not /etc/avahi, but rather /etc. This is quite common for /etc and /usr watches. Commit message updated.

ghost:
Let me throw out an argument why you would label /etc/avahi with a private avahi_conf_t type:

There is an avahi_admin() interface in avahi.if. This interface is used for confined avahi administration. Imagine using that in its current shape (i.e. you have no access to write /etc/avahi). That makes this interface pretty pointless. On the other hand if you allow "avahi admin" access to write etc_t type files then that makes avahi_admin() pretty pointless as well.

zpytela:
I actually misread the denial, avahi does chroot. Need to update the commit.

WOnder93 (approved):
LGTM, and I also verified that it fixes the AVC on my system.

PR #573, opened 2021-02-11
Allow systemd-logind integrity lockdown permission

zpytela:
The integrity lockdown permission is required for the hibernation
feature allowing userspace to modify the kernel.

Resolves: rhbz#1927557

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -200,6 +200,13 @@ systemd_domain_template(systemd_sleep)
 #  is for /run/user/$USER ($USER ownership is $USER:$USER)
 allow systemd_logind_t self:capability { chown kill dac_read_search dac_override fowner sys_tty_config sys_admin };
 allow systemd_logind_t self:capability2 block_suspend;
+
+# systemd-logind reads state from /sys/power, which changes output based on
+# whether hibernations is available, which tries to take the lockdown state
+# into account. So the permission is somewhat unnecessary (systemd-logind
+# doesn't actually try to change anything), but it's better to allow it so that
+# systemd-logind sees the right system state.
+allow systemd_logind_t self:lockdown integrity;
 allow systemd_logind_t self:process getcap;
 allow systemd_logind_t self:netlink_kobject_uevent_socket create_socket_perms;
 allow systemd_logind_t self:unix_dgram_socket create_socket_perms;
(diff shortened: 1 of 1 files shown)


policy/modules/system/systemd.te
```diff
@@ -200,6 +200,7 @@ systemd_domain_template(systemd_sleep)
 #  is for /run/user/$USER ($USER ownership is $USER:$USER)
 allow systemd_logind_t self:capability { chown kill dac_read_search dac_override fowner sys_tty_config sys_admin };
 allow systemd_logind_t self:capability2 block_suspend;
+allow systemd_logind_t self:lockdown integrity;
```
WOnder93, line 209: It seems systemd-logind reads /sys/power/state (or similar) to see the available sleep states, which peeks at lockdown integrity status to detect whether hibernation is available, triggering an SELinux check. Let's document it for future reference.

Full AVC from my machine:
```
type=PROCTITLE msg=audit(11.02.2021 23:25:27.196:411) : proctitle=/usr/lib/systemd/systemd-logind 
type=SYSCALL msg=audit(11.02.2021 23:25:27.196:411) : arch=x86_64 syscall=read success=yes exit=16 a0=0x1d a1=0x55873c9d7500 a2=0x1000 a3=0x40000 items=0 ppid=1 pid=1604 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=systemd-logind exe=/usr/lib/systemd/systemd-logind subj=system_u:system_r:systemd_logind_t:s0 key=(null) 
type=AVC msg=audit(11.02.2021 23:25:27.196:411) : avc:  denied  { integrity } for  pid=1604 comm=systemd-logind lockdown_reason="hibernation" scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:system_r:systemd_logind_t:s0 tclass=lockdown permissive=1 
```
```suggestion
# systemd-logind reads state from /sys/power, which changes output based on whether
# hibernations is available, which tries to take the lockdown state into account. So
# the permission is somewhat unnecessary (systemd-logind doesn't actually try to
# change anything), but it's better to allow it so that systemd-logind sees the right
# system state.
allow systemd_logind_t self:lockdown integrity;
```

policy/modules/system/systemd.te
```diff
@@ -1294,6 +1301,7 @@ systemd_read_efivarfs(systemd_userdbd_t)
 
 allow systemd_sleep_t self:capability sys_resource;
 dontaudit systemd_sleep_t self:capability sys_ptrace;
+allow systemd_sleep_t self:lockdown integrity;
```
WOnder93, line 1305: Please add a short comment above this one as well for future reference.

zpytela: A oneliner added.

WOnder93 (approved):

PR #574, opened 2021-02-11
Allow unconfined_domain_type watch domain files

zpytela:
This permission is required for watching files in /proc/pid.

Resolves: rhbz#1927549


zpytela:
I wonder if watching directories and symlinks will be required later, too.

WOnder93:
An alternative could be to add `watch_*_perms` to `manage_*_perms` - if "manage" implies to have a sort of full control over the objects, it would be natural to allow watching for their changes as well. (And it could save us a few similar pull requests in the future.)

zpytela:
I was thinking that way, too - it seems to make sense for file, lnk_file, dir. I would just exclude watch_sb and for lnk_file also watch_mount if I understand properly how they are used.

WOnder93:
Good point, we should probably only add `watch` and `watch_reads` (AFAIK, `watch_with_perm` is used for access control like fapolicyd, which is more dangerous - e.g. SIGSTOP'ing such process can lock up the system). Not sure why you suggest to exclude `watch_mount` only for `lnk_file` - I don't know much about it, so maybe you know something I don't :)

zpytela:
Let's start with the two and time will tell.
I don't think I have any special resources to draw from but the general ones like
https://git.kernel.org/pub/scm/linux/kernel/git/pcmoore/selinux.git/commit/?h=next&id=ac5656d8a4cdd93cd2c74355ed12e5617817e0e7

WOnder93:
Thanks, the changes look good, but you have some diff in the commit message (accidentally pasted from clipboard, I guess).

WOnder93:
I wanted to check one more thing - there are now some interface calls that have become redundant - e.g. `fs_watch_cgroup_files(init_t)` in init.te, because there is also `fs_manage_cgroup_files(init_t)`. I'll try to find all that I can so we can remove them in this PR.

WOnder93:
OK, it seems the init_t one is the only one.

zpytela:
I think miscfiles_watch_localization_symlinks is now redundant, too.

WOnder93:
True, I didn't expect there to be a manage version of it so I didn't even grep for it :)

I can see you also removed `files_mounton_etc(init_t)`, which seems correct, but is not mentioned in the commit message (could even be a separate commit).

zpytela:
Mentioned, misspelled. `Unrelated duplicate files_watch_etc_dirs(init_t) was removed, too.`

WOnder93:
Now the second commit's message doesn't match the content... And it is actually not clear to me which interface covers `files_watch_etc_dirs(init_t)` - could you explain it in the commit message?

WOnder93:
I still can't see the explanation for the `files_watch_etc_dirs(init_t)` removal, (And I'm not nitpicking here, I really want to understand it :D)

zpytela:
I probably had highlighted `files_manage_etc` in editor and didn't read till the end of the string.

WOnder93:
Ah, okay, I thought I was missing something obvious :)

WOnder93 (approved):

zpytela:
The only obvious thing was I had made a mistake ;).

PR #576, opened 2021-02-11
Fix several issues found while dogfooding with latest GitHub policy

WOnder93:
This fixes a few of the AVCs that appeared on my F33 machine after I installed selinux-policy built directly from the rawhide branch [1] (plus updated SELinux userspace from F34 repos). Most of the rest have already been reported, one is related to snappy-selinux, and one will need further investigation.

[1] https://copr.fedorainfracloud.org/coprs/omos/selinux-policy-latest/

### DIFF
--- a/policy/modules/system/sysnetwork.if
+++ b/policy/modules/system/sysnetwork.if
@@ -442,20 +442,18 @@ interface(`sysnet_read_config',`
 	')
 
 	files_search_etc($1)
-	allow $1 net_conf_t:file read_file_perms;
+	read_files_pattern($1, net_conf_t, net_conf_t)
+	read_lnk_files_pattern($1, net_conf_t, net_conf_t)
 
 	ifdef(`distro_debian',`
 		files_search_pids($1)
-		allow $1 net_conf_t:dir list_dir_perms;
-		read_files_pattern($1, net_conf_t, net_conf_t)
+		list_dirs_pattern($1, net_conf_t, net_conf_t)
 	')
 
 	ifdef(`distro_redhat',`
-        files_search_all_pids($1)
-        init_search_pid_dirs($1)
-		allow $1 net_conf_t:dir list_dir_perms;
-		read_files_pattern($1, net_conf_t, net_conf_t)
-		read_lnk_files_pattern($1, net_conf_t, net_conf_t)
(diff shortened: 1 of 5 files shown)


WOnder93:
Ah, the last commit doesn't actually fix the issue... I confused `dnssec_trigger_t` (defined in `dnssec.te`) with `dnssec_triggerd_t` (defined in `dnssectrigger.te`)...

policy/modules/system/sysnetwork.if
```diff
@@ -513,6 +512,7 @@ interface(`sysnet_create_config',`
 
 	files_search_etc($1)
 	allow $1 net_conf_t:file create_file_perms;
+	allow $1 net_conf_t:lnk_file create_lnk_file_perms;
```
zpytela, line 515: As `create_file_perms` is not sufficient for creating files, I'd rather see it replaced with `create_files_pattern`. I suppose it just happens to work because of other rules bringing the required permissions. The same for lnk_files.

WOnder93: Good point! I'll fix it up in a while...

WOnder93: Actually, I would probably just replace `files_search_etc($1)` with `files_rw_etc_dirs($1)`. `create_files_pattern()` would just mean allowing to have `net_conf_t` directories in the chain, which isn't currently allowed in this interface (not even conditionally). All this should be refactored one day, but I think I went deep enough into the rabbit hole for now :)

zpytela: What I had in mind was `create_files_pattern($1, etc_t, net_conf_t)`. It is not clear at first glance `create_file_perms` are not sufficient so it does not give a good example to follow. Surely it works.

WOnder93: Yes, but `etc_t` is defined in another module, so it shouldn't be referenced there directly. And to avoid that one would have to create some partial pattern in files.if and maybe convert the other places to use it... as I said, it would be a deep rabbit hole :)

zpytela: OK, agreed.

zpytela:
Merging, thank you.

PR #578, opened 2021-02-12
Fix docstring for init_watch_dir()

WOnder93:
The lines must begin with two #s, not just one.

Fixes:
support/segenxml.py: warning: unable to find XML for interface init_watch_dir()


zpytela:
Good catch, thanks.

PR #580, opened 2021-02-12
Allow dnsmasq watch /etc dirs

zpytela:


WOnder93:
This is already included in #576 :)

zpytela:
Oops, I just went through the list top down.

PR #585, opened 2021-02-12
Allow pulseaudio watch /dev and systemd-logind session dirs

zpytela:
The dev_watch_generic_dirs() interface was added.

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -321,6 +321,24 @@ interface(`dev_create_generic_dirs',`
 
 ########################################
 ## <summary>
+##	Watch generic device directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dev_watch_generic_dirs',`
+	gen_require(`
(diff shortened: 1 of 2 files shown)


zpytela:
Pulseaudio actually watches `/dev/snd` dir, but currently no directory has a type different to `dev_t`.

policy/modules/kernel/devices.if
```diff
@@ -319,6 +319,24 @@ interface(`dev_create_generic_dirs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dev_watch_generic_dirs',`
+	gen_require(`
+		type device_t;
+	')
+
+	allow $1 device_t:dir watch_dir_perms;
```
WOnder93: Maybe this should rather be `watch_dirs_pattern($1, device_t, device_t)` to ensure that the domain has also search permissions on `device_t` (which I assume is needed to watch e.g. /dev subdirectories).

zpytela: This comment is possibly relevant in other interfaces, too.

WOnder93 (approved):

PR #586, opened 2021-02-12
Build also the container module in CI

WOnder93:
...to make sure that container-selinux will still build after the
change.


zpytela:
Is not `make container.pp` enough once the policy has been built?

WOnder93:
You mean without the `-j $(nproc)`?

zpytela:
No, `-j` is probably useless, but what I meant was
```
       - run: make container.pp -j $(nproc)
```

WOnder93:
Ah, yes, sorry, I missed the extra `policy` target there :) Yes, I'll amend the commit.

zpytela:
Arguments reordering also helped readability, merging, thanks.

PR #588, opened 2021-02-13
Add missing watch permissions for restorecond

WOnder93:
Found via dogfooding. I verified that with these two commits I no longer get any restorecond-related AVCs.


zpytela:
LGTM. Any reason for keeping 2 commits?

WOnder93:
It's two slightly different things - watching the config file vs. watching "random" directories, so I kept it separate. In this case I guess both one or two would be OK, so I would just let it be as it is.

zpytela:
I'd prefer one, but merging now, thanks.

PR #590, opened 2021-02-15
Allow udev integrity lockdown permission

luk1337:
The integrity lockdown permission is required for udev to
load proprietary Nvidia driver.

### DIFF
--- a/policy/modules/kernel/kernel.if
+++ b/policy/modules/kernel/kernel.if
@@ -405,6 +405,21 @@ interface(`kernel_load_module',`
 
 ########################################
 ## <summary>
+##	Allows caller to load unsigned kernel modules
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`kernel_load_unsigned_module',`
+	allow $1 self:lockdown integrity;
+	kernel_load_module($1)
+')
+
+########################################
+## <summary>
 ##	Allow search the kernel key ring.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/system/modutils.te
+++ b/policy/modules/system/modutils.te
@@ -75,7 +75,7 @@ can_exec(kmod_t, update_modules_exec_t)
 manage_files_pattern(kmod_t,kmod_tmpfs_t,kmod_tmpfs_t)
 fs_tmpfs_filetrans(kmod_t,kmod_tmpfs_t,file)
 
-kernel_load_module(kmod_t)
+kernel_load_unsigned_module(kmod_t)
 files_manage_kernel_modules(kmod_t)
 files_map_kernel_modules(kmod_t)
 kernel_read_system_state(kmod_t)
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1163,7 +1163,7 @@ read_lnk_files_pattern(systemd_domain, systemd_home_t, systemd_home_t)
 allow systemd_modules_load_t self:system module_load;
 
 kernel_dgram_send(systemd_modules_load_t)
-kernel_load_module(systemd_modules_load_t)
+kernel_load_unsigned_module(systemd_modules_load_t)
 kernel_ib_access_unlabeled_pkeys(systemd_modules_load_t)
 
 corecmd_exec_bin(systemd_modules_load_t)
--- a/policy/modules/system/udev.te
+++ b/policy/modules/system/udev.te
@@ -87,7 +87,7 @@ allow udev_t udev_var_run_t:dir mounton;
 allow udev_t udev_var_run_t:lnk_file relabel_lnk_file_perms;
 dev_filetrans(udev_t, udev_var_run_t, { file lnk_file } )
 
-kernel_load_module(udev_t)
+kernel_load_unsigned_module(udev_t)
 kernel_read_system_state(udev_t)
 kernel_request_load_module(udev_t)
 kernel_getattr_core_if(udev_t)


luk1337:
Observed this issue on recently branched F34.

zpytela:
Thank you for the PR. Issues like this should be gone with selinux-policy-3.14.7-20 - can you confirm it?

There was a more general solution in place:
https://github.com/fedora-selinux/selinux-policy/pull/569

zpytela:
It was actually the confidentiality permission, sorry for the noise.

WOnder93:
If I understand it correctly, this happens due to udev trying to load a kernel module right? So we should instead modify the `kernel_load_module()` interface to also grant `lockdown { integrity }`, since all domains that need to load modules will need it.

luk1337:
Correct.

WOnder93:
Ah, actually I'm wrong... the integrity permission is only needed when loading unsigned modules. So I think we should add a new `kernel_load_unsigned_module()` interface (which would be `kernel_load_module() + lockdown integrity) and use it in udev.te instead of `kernel_load_module()`.

WOnder93:
@luk1337 BTW, can you please share the full AVC message you got?

luk1337:
```
type=AVC msg=audit(1613422350.218:1035): avc:  denied  { integrity } for  pid=6569 comm="modprobe" lockdown_reason="unsigned module loading" scontext=system_u:system_r:kmod_t:s0 tcontext=system_u:system_r:kmod_t:s0 tclass=lockdown permissive=0
```

luk1337:
oops, wrong one.
```
type=AVC msg=audit(1613422234.465:256): avc:  denied  { integrity } for  pid=607 comm="systemd-udevd" lockdown_reason="unsigned module loading" scontext=system_u:system_r:udev_t:s0-s0:c0.c1023 tcontext=system_u:system_r:udev_t:s0-s0:c0.c1023 tclass=lockdown permissive=0
```

WOnder93:
@luk1337 Thanks, that confirms that it is indeed because of an unsigned module. Do you want to try to update the PR as I suggested or would you rather leave it on us? :)

luk1337:
I'll do it myself in like an hour or so.

luk1337:
Updated PR.

Also unrelated but on F34 I'm also getting following gmain denial every boot:
```
Raw Audit Messages
type=AVC msg=audit(1613430820.26:798): avc:  denied  { watch } for  pid=3300 comm="gmain" path="/var/lib/AccountsService/icons" dev="nvme0n1p3" ino=921879 scontext=system_u:system_r:xdm_t:s0-s0:c0.c1023 tcontext=system_u:object_r:accountsd_var_lib_t:s0 tclass=dir permissive=0
```
should this be addressed or should I just ignore it?

policy/modules/kernel/kernel.te
```diff
@@ -14,6 +14,7 @@ gen_bool(secure_mode_insmod, false)
 
 # assertion related attributes
 attribute can_load_kernmodule;
+attribute can_load_unsigned_kernmodule;
```
WOnder93: I'd say you don't need to create a new attribute for this, since there is no neverallow rule for it (and `neverallow ~can_load_unsigned_kernmodule self:lockdown integrity;` wouldn't make sense, since the permission can be required for various other things). Please just add the allow rule directly to the interface.

luk1337: Done.

WOnder93 (approved):
Thanks, looks good now!

zpytela:
@luk1337 looks good; in the interface description imperative should be used, i. e. Allow rather than Allows; surely I see in the previous just Allows.

More important: In one of the previous comments, you also posted the same AVC for modprobe: is it still in place? I also wonder if systemd-modules-load should have the same permission.

The watch permission needs to be addressed, too.

WOnder93:
The watch AVC has already been reported: https://bugzilla.redhat.com/show_bug.cgi?id=1928546

luk1337:
@zpytela I switched few more domains to `kernel_load_unsigned_module` and now my unsigned modules in `/etc/modules-load.d` are loaded on boot too. I also granted `integrity` for `unconfined_t` so that I can modprobe from shell.
I think it's fine to leave the description as is since it matches the interface right above it.

zpytela:
@luk1337 where did you happen to find the `unconfined_t self:lockdown integrity` perm?

zpytela:
I see now "so that I can modprobe from shell." sysadm has a transition to kmod, so I suppose this is enough. Thank you for your contribution.

policy/modules/roles/unconfineduser.te
```diff
@@ -65,6 +65,8 @@ typealias unconfined_t alias unconfined_crontab_t;
 dontaudit unconfined_t self:dir write;
 dontaudit unconfined_t self:file setattr;
 
+allow unconfined_t self:lockdown integrity;
```
WOnder93: The `integrity` permission should rather be added to `allow unconfined_domain_type self:lockdown confidentiality;` in `domain.te`.

WOnder93 (changes requested):

zpytela:
@WOnder93 are you sure about the scope? It sounds too excessive for me, even when we talk about unconfined domains.

policy/modules/roles/unconfineduser.te
luk1337: Done.

WOnder93:
Yes, I think we should keep it consistent. The things covered by lockdown::integrity have likely been allowed to unconfined domains before. I think we should rather try to reduce the amount of unconfined domains than diverging unconfined_t from unconfined_domain_type.

zpytela:
Well, I believe everything in the world has a purpose :) and the purpose of the lockdown class imho is not to allow potentially dangerous actions to anybody.
BTW both `kmod_t` and `udev_t` are `unconfined_domain_type`.

WOnder93:
Okay, I'm fine with making an exception for lockdown, but then let's go all the way and move also `confidentiality` to just `unconfined_t` (and maybe also `unconfined_service_t`). I think it should be done in a separate PR, though, and resolve only kmod_t and udev_t here.

luk1337:
So you basically want to split this into 3 changes? Like so:
 - confidentiality -> {unconfined_t, unconfined_service_t}
 - integrity -> {kmod_t, udev_t}
 - integrity -> unconfined_t

WOnder93:
My idea is to keep only the changes in kernel.if, modutils.te, systemd.te (missed this one in the previous comment, sorry), and udev.te, and leave the unconfined changes for another PR (so that this one is not held up by the discussion). But let's wait for @zpytela's opinion.

zpytela:
Agree with @WOnder93. We already got to the solution for modutils/udev/systemd-modules-load, let's finish it and merge this PR so that a new build can be created to fix it.
Regarding unconfined_domain_type: I am just saying it seems to me too much now. It is possible we will need to do it in the future this way; but still, I'd rather see as few permission as possible, especially those "dangerous" ones.

luk1337:
Done, removed unconfined changes.

WOnder93 (approved):

zpytela:
@luk1337 thank you, merging.
Feel free to open a new PR for unconfined if you wish.

PR #594, opened 2021-02-16
Enable systemd-sysctl domtrans for udev

WOnder93:
On my machine, systemd-udevd executes systemd-sysctl, so give it the
right domain transition.


zpytela:
The commit is clear, merging, thanks.

PR #596, opened 2021-02-17
Allow upowerd to send syslog messages

WOnder93:
Fixes the following denial found while running with the unconfined
module disabled:
type=AVC msg=audit(1613549055.280:2392): avc:  denied  { sendto } for  pid=2639 comm="upowerd" path="/run/systemd/journal/socket" scontext=system_u:system_r:devicekit_power_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=unix_dgram_socket permissive=1

### DIFF
--- a/policy/modules/contrib/devicekit.te
+++ b/policy/modules/contrib/devicekit.te
@@ -341,6 +341,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	logging_send_syslog_msg(devicekit_power_t)
+')
+
+optional_policy(`
 	modutils_domtrans_kmod(devicekit_power_t)
 ')
 


WOnder93:
I saw this denial every time I suspended my KDE workstation to RAM. I can confirm that with this patch it no longer appears.

policy/modules/contrib/devicekit.te
```diff
@@ -278,6 +278,8 @@ term_use_all_inherited_terms(devicekit_power_t)
 
 auth_use_nsswitch(devicekit_power_t)
 
+logging_send_syslog_msg(devicekit_power_t)
```
zpytela: Strange as it may seem, logging is not in base.
Until recently no policy module had it in the optional block.

WOnder93: OK, I changed it to an optional block.

zpytela:
LGTM, thanks, will merge it once CI finishes.

PR #599, opened 2021-02-17
Handle NVMe devices as in refpolicy

WOnder93:
Refpolicy doesn't have a separate nvme_device_t type and just labels
these devices as fixed_disk_t. That is much more sane than having to add
dev_something_nvme() all over the place, when it could have been covered
by the existing storage_*_fixed_disk() interfaces.

So switch the labeling rules to match refpolicy and deprecate the
dev_read_nvme() and dev_rw_nvme() interfaces. All existing callers of
these already had the equivalent storage_*() calls, with two exceptions:
1. tlp_t had dev_rw_nvme(tlp_t), but only
   storage_raw_read_fixed_disk(tlp_t). I added
   storage_raw_write_fixed_disk(tlp_t) to balance it.
2. xdm_t had dev_read_nvme(xdm_t), but
   storage_dontaudit_read_fixed_disk(xdm_t). I left only the dontaudit
   rules - xdm_t probably doesn't actually need the permission.

### DIFF
--- a/policy/modules/contrib/blkmapd.te
+++ b/policy/modules/contrib/blkmapd.te
@@ -29,7 +29,6 @@ files_pid_filetrans(blkmapd_t, blkmapd_var_run_t, file)
 kernel_read_system_state(blkmapd_t)
 
 dev_list_sysfs(blkmapd_t)
-dev_read_nvme(blkmapd_t)
 
 fs_list_rpc(blkmapd_t)
 fs_rw_rpc_named_pipes(blkmapd_t)
--- a/policy/modules/contrib/dbus.te
+++ b/policy/modules/contrib/dbus.te
@@ -103,8 +103,6 @@ dev_read_sysfs(system_dbusd_t)
 dev_rw_inherited_input_dev(system_dbusd_t)
 dev_rw_inherited_dri(system_dbusd_t)
 
-dev_rw_nvme(system_dbusd_t)
-
 files_rw_inherited_non_security_files(system_dbusd_t)
 files_watch_usr_dirs(system_dbusd_t)
 
--- a/policy/modules/contrib/hddtemp.te
+++ b/policy/modules/contrib/hddtemp.te
@@ -37,8 +37,6 @@ corenet_tcp_bind_hddtemp_port(hddtemp_t)
 corenet_sendrecv_hddtemp_server_packets(hddtemp_t)
 corenet_tcp_sendrecv_hddtemp_port(hddtemp_t)
 
-dev_read_nvme(hddtemp_t)
-
 storage_raw_read_fixed_disk(hddtemp_t)
 storage_raw_read_removable_device(hddtemp_t)
 
--- a/policy/modules/contrib/kdumpgui.te
+++ b/policy/modules/contrib/kdumpgui.te
@@ -46,7 +46,6 @@ dev_dontaudit_getattr_all_chr_files(kdumpgui_t)
 dev_read_sysfs(kdumpgui_t)
 dev_read_urand(kdumpgui_t)
 dev_getattr_all_blk_files(kdumpgui_t)
-dev_read_nvme(kdumpgui_t)
 
 files_manage_boot_files(kdumpgui_t)
 files_manage_boot_symlinks(kdumpgui_t)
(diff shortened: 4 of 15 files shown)


zpytela:
I'd like to see the commit message header as Label nvme devices with fixed_disk_t or otherwise describe the actual change.
Does tlp really need write to the disk devices?
OTOH gnome shell probably needs getattr, see https://bugzilla.redhat.com/show_bug.cgi?id=1380951.
systemd_dbusd_t is now not allowed open/lock/ioctl.

Will the devices files type be assigned correctly by the nvme driver after the change?

WOnder93:
> I'd like to see the commit message header as Label nvme devices with fixed_disk_t or otherwise describe the actual change.

I concur - will update the commit message.

> Does tlp really need write to the disk devices?

Possibly not, at least the dev_read_nvme() seems to have been upgraded to rw based on a false premise that ioctl is not provided by the read interface:
8659df15169ae04f8e92992709feb826fb22016b
https://bugzilla.redhat.com/show_bug.cgi?id=1696943

And the same for the addition of `storage_raw_write_removable_device(tlp_t)`:
ed331699267372bbb62d5e412e131f644d68ae81
https://bugzilla.redhat.com/show_bug.cgi?id=1436830

Would you prefer if I remove both the storage...write calls?

> OTOH gnome shell probably needs getattr, see https://bugzilla.redhat.com/show_bug.cgi?id=1380951.

Hm, indeed... I think it would make sense to allow getattr for both fixed disks and removable devices and leave the rest dontaudited.

> systemd_dbusd_t is now not allowed open/lock/ioctl.

It shouldn't need those. Apparently dbus daemon is just handling disk fds that are being passed between processes:
https://bugzilla.redhat.com/show_bug.cgi?id=1456566
https://bugzilla.redhat.com/show_bug.cgi?id=1323601

> Will the devices files type be assigned correctly by the nvme driver after the change?

Since there are no filename transition rules for NVMe, I'd expect so. AFAIK udevd or something watches for devices appearing in /dev and relabels them based on the policy and also systemd relabels /dev and other filesystems after loading policy. I'll reboot soon into a policy with this patch applied, so we'll see...

WOnder93:
Rebooted and it seems to work just fine:
```
$ ll -Z /dev/nvme*
crw-------. 1 root root system_u:object_r:fixed_disk_device_t:s0 239, 0 feb 18 22:45 /dev/nvme0
brw-rw----. 1 root disk system_u:object_r:fixed_disk_device_t:s0 259, 0 feb 18 22:45 /dev/nvme0n1
brw-rw----. 1 root disk system_u:object_r:fixed_disk_device_t:s0 259, 1 feb 18 22:45 /dev/nvme0n1p1
brw-rw----. 1 root disk system_u:object_r:fixed_disk_device_t:s0 259, 2 feb 18 22:45 /dev/nvme0n1p2
```
No unexpected denials (and none for fixed_disk_device_t) appeared.

zpytela:
> > I'd like to see the commit message header as Label nvme devices with fixed_disk_t or otherwise describe the actual change.
> 
> I concur - will update the commit message.
> 
> > Does tlp really need write to the disk devices?
> 
> Possibly not, at least the dev_read_nvme() seems to have been upgraded to rw based on a false premise that ioctl is not provided by the read interface:
> [8659df1](https://github.com/fedora-selinux/selinux-policy/commit/8659df15169ae04f8e92992709feb826fb22016b)
> https://bugzilla.redhat.com/show_bug.cgi?id=1696943
> 
> And the same for the addition of `storage_raw_write_removable_device(tlp_t)`:
> [ed33169](https://github.com/fedora-selinux/selinux-policy/commit/ed331699267372bbb62d5e412e131f644d68ae81)
> https://bugzilla.redhat.com/show_bug.cgi?id=1436830
> 
> Would you prefer if I remove both the storage...write calls?

I'd remove both, although tlp is not easy to troubleshoot and test as it has multiple ways of invocation so a denial can pop up later.

> 
> > OTOH gnome shell probably needs getattr, see https://bugzilla.redhat.com/show_bug.cgi?id=1380951.
> 
> Hm, indeed... I think it would make sense to allow getattr for both fixed disks and removable devices and leave the rest dontaudited.

Yep.

WOnder93:
Should be better now.

zpytela:
Looks good now, thanks.

PR #600, opened 2021-02-18
Further fixes for unconfined domains

WOnder93:
This PR attempts to fix some of the denials I got after a reboot with the unconfined module disabled.


policy/modules/system/systemd.if
```diff
@@ -102,7 +102,7 @@ interface(`systemd_exec_systemctl',`
 	systemd_login_read_pid_files($1)
 	systemd_passwd_agent_exec($1)
 
-	dontaudit $1 self:capability net_admin;
+	dontaudit $1 self:capability { net_admin sys_ptrace };
```
zpytela, line 105: Given how much troubles there are with dontaudit rules, did you think of a diffferent approach?

zpytela:
Most of it LGTM. For the `Grant libvirtd the fsetid capability` commit:
- almost always we now use allow, not grant
- typo wncountered

WOnder93:
>     * almost always we now use allow, not grant

I though that "allow" should be used only with verbs, but the dictionary says otherwise.

>     * typo wncountered

Both are now fixed.

policy/modules/system/systemd.if
WOnder93: Giving every domain that execs systemctl sys_ptrace seems too excessive to me. An alternative could be to do a transition on systemctl calls (I mentioned that in the commit message), but I'm not sure about the implications... It would probably become impossible to control which domain can control which service (unless each calling domain gets its own something_sysctl_t type...)

zpytela: OK, I don't see any easy solution. Perhaps some dedicated capability?

zpytela:
Merging now, thanks.

PR #601, opened 2021-02-18
Add initial policy for systemd-coredump

WOnder93:
Based on the denials I got with the unconfined module disabled.
Apparently some unprivileged Qt application crashed and systemd-coredump
tried to record its core dump. I tried to keep the policy broad enough
to allow coredumping any process, although there might be some minor
gaps still.

Anyway, I believe this is enough to drop the unconfinedness of the
domain, so I removed the unconfined_domain() call.

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -45,6 +45,9 @@ systemd_domain_template(systemd_coredump)
 type systemd_coredump_tmpfs_t;
 files_tmpfs_file(systemd_coredump_tmpfs_t)
 
+type systemd_coredump_var_lib_t;
+files_type(systemd_coredump_var_lib_t)
+
 systemd_domain_template(systemd_hwdb)
 
 type systemd_hwdb_unit_file_t;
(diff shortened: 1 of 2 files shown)


WOnder93:
(Force-pushed to add & update comments.)

zpytela:
The commit heading is misleading: we already have a (stub) policy for systemd-coredump.

I spotted systemd-tmpfiles needs to have access to the new type, should be covered by a rule for non_auth_file_type unless some permission which is not enumerated is needed.

WOnder93:
> The commit heading is misleading: we already have a (stub) policy for systemd-coredump.

OK, I'll rephrase it to "Complete initial policy".

> I spotted systemd-tmpfiles needs to have access to the new type, should be covered by a rule for non_auth_file_type unless some permission which is not enumerated is needed.

Could you be more specific? Do you have an AVC?

zpytela:
> > The commit heading is misleading: we already have a (stub) policy for systemd-coredump.
> 
> OK, I'll rephrase it to "Complete initial policy".
> 
> > I spotted systemd-tmpfiles needs to have access to the new type, should be covered by a rule for non_auth_file_type unless some permission which is not enumerated is needed.
> 
> Could you be more specific? Do you have an AVC?

No other reference than `/usr/lib/tmpfiles.d/systemd*.conf`, no denial, just a note, hope it's okay.
<rubbish deleted>

WOnder93:
> No other reference than `/usr/lib/tmpfiles.d/systemd*.conf`, no denial, just a note, hope it's okay.

I see. That should be covered by `files_manage_non_auth_files(systemd_tmpfiles_t)` - and the output of `sesearch -A -s systemd_tmpfiles_t -t systemd_coredump_var_lib_t` confirms it (there are lots of permissions, should be enough).

zpytela:
What's happened to this PR?

WOnder93:
Oh, I accidentally deleted the branch :facepalm: Let me get it back...

policy/modules/system/systemd.te
```diff
@@ -1009,14 +1012,38 @@ systemd_read_efivarfs(systemd_sysctl_t)
 #
 # systemd_coredump domains
 #
+
+# dac_read_search - (at least) to read /proc/<pid>/exe symlinks
```
zpytela: Is this really needed? I cannot find any process with the directory permissions different to 0555.

WOnder93: Actually, I was looking at the wrong audit record when writing the description... `sys_ptrace` is the one required for `/proc/<pid>/exe`; and `dac_read_search` is needed for acessing `/proc/<pid>/fd` of a non-root process. Here is the full AVC:
```
type=PROCTITLE msg=audit(18.02.2021 22:35:22.772:1247) : proctitle=/usr/lib/systemd/systemd-coredump 231283 1000 100756 6 1613684122 18446744073709551615 omos.redhat.com 
type=PATH msg=audit(18.02.2021 22:35:22.772:1247) : item=0 name=/proc/231283/fd inode=3496460 dev=00:17 mode=dir,500 ouid=omosnace ogid=omosnace rdev=00:00 obj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0 
type=CWD msg=audit(18.02.2021 22:35:22.772:1247) : cwd=/ 
type=SYSCALL msg=audit(18.02.2021 22:35:22.772:1247) : arch=x86_64 syscall=openat success=yes exit=1 a0=0xffffff9c a1=0x7ffeaae5aaa0 a2=O_RDONLY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC a3=0x0 items=1 ppid=2 pid=231284 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=systemd-coredum exe=/usr/lib/systemd/systemd-coredump subj=system_u:system_r:systemd_coredump_t:s0 key=(null) 
type=AVC msg=audit(18.02.2021 22:35:22.772:1247) : avc:  denied  { dac_read_search } for  pid=231284 comm=systemd-coredum capability=dac_read_search  scontext=system_u:system_r:systemd_coredump_t:s0 tcontext=system_u:system_r:systemd_coredump_t:s0 tclass=capability permissive=1 
```

zpytela: That makes a difference. Will you change the comment?

WOnder93: Yes, it's reworded now.

policy/modules/system/systemd.te
```diff
@@ -1009,14 +1012,38 @@ systemd_read_efivarfs(systemd_sysctl_t)
...
+
+# To read info about the crashed process from /proc
+domain_read_all_domains_state(systemd_coredump_t)
+
+# To be able to mmap the dumped process' executable file for reading
+# (can be basically any file type)
+files_read_non_security_files(systemd_coredump_t)
+files_map_non_security_files(systemd_coredump_t)
+
+logging_send_syslog_msg(systemd_coredump_t)
```
zpytela:
``` $ grep -B4 ^logging policy/modules.conf
# Module: logging
#
# Policy for the kernel message logger and system logging daemon.
#
logging = module
```

WOnder93: Right, this one again... :D

zpytela:
Now it looks promising, thank you.

PR #602, opened 2021-02-18
Allow systemd-sleep get attributes of fixed disk device nodes

zpytela:


WOnder93:
This BZ seems to have some info that could be useful to have as a comment:
https://bugzilla.redhat.com/show_bug.cgi?id=1928539
For example:
```
# systemd-sleep gets the attributes of the swap partition (hibernation will fail if denied)
storage_...
```

zpytela:
Commit message enhanced and policy comment added.

WOnder93 (approved):

PR #604, opened 2021-02-19
Dontaudit NetworkManager write to initrc_tmp_t pipes

zpytela:
Resolves: rhbz#1892556


WOnder93 (approved):
Kind of a weird denial, but can't think of a better resolution. (Plus there is a precedence with the similar kdump denial.)

PR #605, opened 2021-02-19
Allow rngd stat the /sys filesystem

zpytela:


zpytela:
I just wonder should it rather be allowed for a domain as it eventually happened in 0b26432144f0b6f2140b974b9c508d991286bfa4. Currently there are new requests for rngd and pcscd.

WOnder93:
I tried to trace it on my system and it's triggered via `libusbx` (`libusb_init()`):
```
rngd 133312 [004] 49172.922519: avc:selinux_audited: requested=0x8 denied=0x8 audited=0x8 result=0 scontext=system_u:system_r:rngd_t:s0 tcontext=system_u:object_r:sysfs_t:s0 tclass=filesystem
        ffffffff88544f43 avc_audit_post_callback+0x1d3 ([kernel.kallsyms])
        ffffffff88544f43 avc_audit_post_callback+0x1d3 ([kernel.kallsyms])
        ffffffff885666f1 common_lsm_audit+0x111 ([kernel.kallsyms])
        ffffffff885459f9 slow_avc_audit+0x69 ([kernel.kallsyms])
        ffffffff88546098 avc_has_perm+0x138 ([kernel.kallsyms])
        ffffffff88549e6d selinux_sb_statfs+0x5d ([kernel.kallsyms])
        ffffffff885413b2 security_sb_statfs+0x22 ([kernel.kallsyms])
        ffffffff8835e7ac statfs_by_dentry+0x4c ([kernel.kallsyms])
        ffffffff8835ee57 user_statfs+0x47 ([kernel.kallsyms])
        ffffffff8835eec0 __do_sys_statfs+0x10 ([kernel.kallsyms])
        ffffffff88bbadc3 do_syscall_64+0x33 ([kernel.kallsyms])
        ffffffff88c0008c entry_SYSCALL_64_after_hwframe+0x44 ([kernel.kallsyms])
            7f89c0cbffab __GI___statfs+0xb (inlined)
            7f89c0bbe39e [unknown] (/usr/lib64/libusb-1.0.so.0.3.0)
            7f89c0bbce9a libusb_init+0x1ea (/usr/lib64/libusb-1.0.so.0.3.0)
            7f89c117054c rtlsdr_get_device_count+0x2c (/usr/lib64/librtlsdr.so.0.6.0)
            5573a4dda301 [unknown] (/usr/sbin/rngd)
            5573a4dd1a63 [unknown] (/usr/sbin/rngd)
            7f89c0bf6b74 __libc_start_main+0xd4 (/usr/lib64/libc-2.32.9000.so)
            5573a4dd282d [unknown] (/usr/sbin/rngd)
```
pcscd likely uses libusb[x], too.

libusbx page says it's deprecated in favor of libusb, which also has that pattern:
https://github.com/libusb/libusb/blob/f38f09da98acc63966b65b72029b1f7f81166bef/libusb/os/linux_usbfs.c#L394

It is very similar to the procfs one and might be used in other libraries as well, so I'd incline to allow it to `domain` as well. It should be pretty harmless anyway.

zpytela:
Changed to a domain. Added a new commit to remove particular calls made redundant with the first one. To stay on the safe side, I haven't touched the calls in `selinux.if`.

WOnder93:
Actually, looking at the new diff I just thought of another option, which is probably cleaner - we could just add the getattr permission/iface call to `dev_read_sysfs()`. As opposed to the procfs check, this one is not used to check if the process is running in a container (or similar sandbox), but is just used to check for the availability of sysfs by applications/libraries that actually want to use it. (Or maybe we could have taken the same approach with procfs... could be worth looking into it later.)

zpytela:
I think this is exactly the right place to allow the permission.

However, we would also add 300 rules to the policy, not 1 as it currently is in this PR.

WOnder93:
Okay, I can live with the `domain` approach, but could we at least document for the future that `dev_getattr_sysfs_fs()` would best suit inside `dev_read_sysfs()`, but we allowed it to all domains instead as an optimization?

zpytela:
OK, commit message and comment updated.

WOnder93 (approved):

PR #607, opened 2021-02-19
Revert "Allow getty watch its private runtime files"

WOnder93:
This reverts commit 87c07c527954eeb256e022b68d7186b7ca9f78fc.

After commit 3ee12bf25f68 ("Add watch permissions to manage_* object
permissions sets"), the watch permission is provided implictly by the
existing manage_files_pattern(getty_t, getty_var_run_t, getty_var_run_t)
call.

### DIFF
--- a/policy/modules/contrib/policykit.te
+++ b/policy/modules/contrib/policykit.te
@@ -152,8 +152,6 @@ optional_policy(`
 # polkit_auth local policy
 #
 
-allow policykit_auth_t policykit_var_lib_t:dir watch_dir_perms;
-
 allow policykit_auth_t self:capability { sys_nice ipc_lock setgid setuid };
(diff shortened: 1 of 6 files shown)


zpytela:
LGTM, thank you.

policy/modules/contrib/policykit.te
```diff
@@ -152,8 +152,6 @@ optional_policy(`
 # polkit_auth local policy
 #
 
-allow policykit_auth_t policykit_var_lib_t:dir watch_dir_perms;
```
WOnder93, line 155: Actually, this one needs to stay. `policykit_auth_t` is allowed to manage `policykit_var_lib_t` files, but not directories. I double-checked the rest, but this seems to be the only such mistake.

zpytela:
LGTM, thank you.

PR #612, opened 2021-02-23
Allow systemd-sleep transition to appropriate domains

zpytela:
The /usr/lib/systemd/system-sleep directory is a place for local
scripts/binaries, executed immediately before entering and after
leaving system suspend/hibernation.

Resolves: rhbz#1927551

### DIFF
--- a/policy/modules/contrib/sysstat.if
+++ b/policy/modules/contrib/sysstat.if
@@ -54,3 +54,22 @@ interface(`sysstat_admin',`
 	logging_search_logs($1)
 	admin_pattern($1, sysstat_log_t)
 ')
+
+########################################
+## <summary>
+##	Execute sysstat_exec_t in the sysstat domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed to transition.
+## </summary>
+## </param>
+#
+interface(`sysstat_domtrans',`
+	gen_require(`
+		type sysstat_t, sysstat_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, sysstat_exec_t, sysstat_t)
+')
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1349,3 +1349,7 @@ fstools_rw_swap_files(systemd_sleep_t)
 
 # systemd-sleep needs to getattr swap partitions
 storage_getattr_fixed_disk_dev(systemd_sleep_t)
+
+optional_policy(`
+	unconfined_server_domtrans(systemd_sleep_t)
+')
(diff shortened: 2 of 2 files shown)


policy/modules/system/systemd.fc
```diff
@@ -50,6 +50,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /usr/lib/systemd/system/.*sleep.*\.(service|target)		--	gen_context(system_u:object_r:power_unit_file_t,s0)
 /usr/lib/systemd/system/.*shutdown.*\.(service|target)		--	gen_context(system_u:object_r:power_unit_file_t,s0)
 /usr/lib/systemd/system/.*suspend.*\.(service|target)		--	gen_context(system_u:object_r:power_unit_file_t,s0)
+/usr/lib/systemd/system-sleep(/.*)		gen_context(system_u:object_r:systemd_sleep_exec_t,s0)
```
WOnder93: The `()`s are not needed. And there should be a `--` so that directories are not matched.

Also, I'd move it one line below so that it's not in the middle of unit file fcontexts.

policy/modules/system/systemd.fc
```diff
@@ -51,6 +51,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /usr/lib/systemd/system/.*shutdown.*\.(service|target)		--	gen_context(system_u:object_r:power_unit_file_t,s0)
 /usr/lib/systemd/system/.*suspend.*\.(service|target)		--	gen_context(system_u:object_r:power_unit_file_t,s0)
 /usr/lib/systemd/system/systemd-userdbd\.(service|socket)		--	gen_context(system_u:object_r:systemd_userdbd_unit_file_t,s0)
+/usr/lib/systemd/system-sleep(/.*)?		gen_context(system_u:object_r:systemd_sleep_exec_t,s0)
```
WOnder93: I'm not convinced that we want to label also the directory with the `systemd_sleep_exec_t` type... At the very least, `systemd_sleep_t` will now need new permissions to read the directory.

Also, won't this cause a regression for the tlp script? Depending on what it does exactly, systemd_sleep_t might not be a good label for it.

Ideally, each service that installs some script there would have its own "sleep script" domain and file label and would register a transition from `systemd_sleep_t` to this domain. And for custom user scripts there could be a special customizable type (see `customizable_types(5)`) with a transition to some near-unconfined domain (or just `unconfined_service_t`) so they can get their scripts to work easily. Otherwise we'll be playing an infinite catch-up game with what users put there and will want to work and will end up with bloated policy for `systemd_sleep_t`...

zpytela: I changed the PR especially for `/usr/lib/systemd/system-sleep` is expected to be a place for user scripts.

PR #615, opened 2021-02-24
Allow collectd bind tcp sockets

im-0:
This allows using collectd with write_prometheus plugin which binds to TCP port 9103 (bacula_port_t).


zpytela:
Hi @im-0, the content of the commit looks fine, but I don't understand the relation between the commit message and the change. For instance, collectd is already allowed to bind to the bacula port:
```
corenet_tcp_bind_bacula_port(collectd_t)
```
Could you update the message so it is clear what and why is changed?

PR #619, opened 2021-02-25
Allow keepalived read inherited nsfs files

zpytela:
This permission is required when keepalived is configured
to using network namespaces.

The fs_read_inherited_nsfs_files() interface was added.

Resolves: rhbz#1921218

### DIFF
--- a/policy/modules/contrib/keepalived.te
+++ b/policy/modules/contrib/keepalived.te
@@ -84,6 +84,7 @@ dev_read_urand(keepalived_t)
 
 files_dontaudit_mounton_rootfs(keepalived_var_run_t)
 files_mounton_rootfs(keepalived_t)
+fs_read_nsfs_files(keepalived_t)
 fs_unmount_tmpfs(keepalived_t)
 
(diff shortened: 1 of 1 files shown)


WOnder93:
In the BZ [Milos reports](https://bugzilla.redhat.com/show_bug.cgi?id=1921218#c1) that in permissive mode also the `open` permission is needed. And the syscall is `openat()` also in the description, so it really doesn't seem to be an inherited file descriptor.

(Also, s/to using/to use/.)

zpytela:
Right. I had a test run, but apparently also in enforcing mode.

WOnder93:
Hm, the subject got somehow lost from the commit message... And the commit now also has the changes from e9818096e91db46008fc5a0c76d9bbc4f8a55763 (should be easily fixed by `git rebase rawhide`).

zpytela:
Well, working on 3 PRs at the same time for the same files can have unexpected effects.

WOnder93 (approved):

PR #620, opened 2021-02-25
Allow keepalived manage its private tmp files

zpytela:
Previously, keepalived was only allowed create and write patterns.
To use the reload_time_file configuration statement, all permissions
are required.

Resolves: rhbz#1907424


zpytela:
Closing as also the original bz has been closed.

PR #622, opened 2021-02-25
Allow kdump load a new kernel

zpytela:


zpytela:
PR updated to include confidentiality for reading /proc/kcore.

policy/modules/kernel/kernel.if
```diff
@@ -1392,6 +1392,7 @@ interface(`kernel_read_core_if',`
 	')
 
 	allow $1 self:capability sys_rawio;
+	allow $1 self:lockdown confidentiality;
```
WOnder93, line 1395: Could you split this out into a separate commit? (It doesn't seem directly related to the unsigned module loading.)

zpytela:
Commits split, commit message updated.

policy/modules/contrib/kdump.te
```diff
@@ -63,6 +63,7 @@ files_read_kernel_modules(kdump_t)
 files_read_kernel_img(kdump_t)
 files_map_boot_files(kdump_t)
 
+kernel_load_unsigned_module(kdump_t)
```
WOnder93: I looked up the RHEL-9 BZ behind this and the operation that kdump is doing is not a kernel module load, but rather `kexec_load(2)`, which allows to specify a kernel image to reboot into (usually this is used to do analysis/memory dump after a system crash). This currently doesn't support signature verification of the kernel being loaded, so it always asks for lockdown integrity permission.

I'd prefer to create a new interface for this (e.g. `kernel_kexec_load()`), which would grant both the `sys_boot` capability ([also required unconditionally for this syscall](https://elixir.bootlin.com/linux/v5.11/source/kernel/kexec.c#L204)) and the `lockdown` `integrity` permission.

policy/modules/kernel/kernel.if
```diff
@@ -4283,3 +4284,23 @@ interface(`kernel_unlabeled_entry_type',`
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`kernel_kexec_load',`
+	gen_require(`
+		type unlabeled_t;
+	')
```
WOnder93: Forgot to remove this?

zpytela: A result of rebasing when the block appeared back.

policy/modules/kernel/kernel.if
```diff
@@ -4283,3 +4284,19 @@ interface(`kernel_unlabeled_entry_type',`
 	allow $1 unlabeled_t:file { mmap_exec_file_perms ioctl lock };
 ')
 
+
```
WOnder93: This is just a cosmetic detail, but there is an extra empty line being added here. And there seems to be also an extra empty line being added to the end of file.

zpytela: Updated.

WOnder93 (commented):
Thanks, looks good now!

WOnder93 (approved):
Sorry, meant to approve, not comment...

PR #624, opened 2021-03-01
Allow apmd watch generic device directories

zpytela:


policy/modules/contrib/apm.te
```diff
@@ -48,6 +48,7 @@ allow apm_t self:capability { dac_read_search  sys_admin sys_resource };
 kernel_read_system_state(apm_t)
 
 dev_rw_apm_bios(apm_t)
+dev_watch_generic_dirs(apm_t)
```
WOnder93: The commit message mentions `apmd`, but you're adding a rule for the client (should be just `apm` according to the RPM contents). One or the other is probably wrong :)

BTW, it would be nice to include at least the original AVC when the original BZ is private and can't be linked, even if the change looks trivial.

zpytela: Right, and it would help to spot the difference in the same commit.

WOnder93 (approved):

PR #625, opened 2021-03-02
Take back excessive permissions for user_u and staff_u

zpytela:
Since 2010, user_u and staff_u are allowed to run both su and sudo
which contradicts the documentation and disrupts the SELinux users
intended plan.

Since this commit, user in the user_r role is not allowed su neither
sudo and user in the staff_r role is only allowed sudo, along with the
documentation.

Resolves: rhbz#1907517

### DIFF
--- a/policy/modules/contrib/cron.if
+++ b/policy/modules/contrib/cron.if
@@ -13,7 +13,6 @@
 #
 template(`cron_common_crontab_template',`
 	gen_require(`
-		attribute crontab_domain;
 		type crontab_exec_t;
 	')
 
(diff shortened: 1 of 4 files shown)


zpytela:
A copr-build is available: https://copr.fedorainfracloud.org/coprs/zpytela/selinux-policy/build/2037371/

zpytela:
PR is ready for review. The latest copr-build (-203 or newer) can be used for testing.
https://copr.fedorainfracloud.org/coprs/zpytela/selinux-policy/build/2038776/

langeman69:
Confirmed to be working on Fedora rawhide (20210304)
also checked staff_u users: sudo is possible, su not.
Nice work, thanks a lot!

zpytela:
Merging based on positive feedback from users.

PR #627, opened 2021-03-02
Allow systemd watch and watch_reads unallocated ttys

zpytela:
The missing permission prevented early debug-shell from working.

The following interfaces were added:
- term_watch_reads_unallocated_ttys
- term_watch_unallocated_ttys

Resolves: rhbz#1933902

### DIFF
--- a/policy/modules/kernel/terminal.if
+++ b/policy/modules/kernel/terminal.if
@@ -1301,6 +1301,42 @@ interface(`term_dontaudit_ioctl_unallocated_ttys',`
 
 ########################################
 ## <summary>
+##	Watch unallocated tty device nodes.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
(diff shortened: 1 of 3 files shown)


zpytela:
I've just verified both of them were requested.

policy/modules/kernel/terminal.if
```diff
@@ -1299,6 +1299,42 @@ interface(`term_dontaudit_ioctl_unallocated_ttys',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`term_watch_unallocated_ttys',`
+	gen_require(`
+		type tty_device_t;
+	')
+
+	allow $1 tty_device_t:chr_file watch_file_perms;
```
WOnder93: (cosmetic) There is already `watch_chr_file_perms` so it would be better to match the class. And `watch_reads_chr_file_perms` can be added and used below, too,

zpytela: Updated.

WOnder93 (approved):

PR #628, opened 2021-03-02
Add 'make validate' step to CI

WOnder93:
Doesn't seem likely that this would break, but add it to CI just in
case. It takes only ~30 seconds when performed after all modules have
been built.


zpytela:
Merging, thanks.

PR #630, opened 2021-03-03
Add an alias for nvme_device_t

WOnder93:
Commit 9613e80506e7 ("Label NVMe devices as fixed_disk_device_t")
removed the nvme_device_t type in favor of fixed_disk_device_t, which
caused issues after a policy update [1], because the label wasn't fixed
up by fixfiles.

To avoid further problems, add a compatibility type alias, which may be
removed in a later release.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1933226

### DIFF
--- a/policy/modules/kernel/storage.te
+++ b/policy/modules/kernel/storage.te
@@ -16,6 +16,8 @@ attribute storage_unconfined_type;
 # /dev/hd* and /dev/sd*.
 #
 type fixed_disk_device_t;
+# Obsoleted in F34: nvme_device_t was merged into fixed_disk_device_t
+typealias fixed_disk_device_t alias nvme_device_t;
 dev_node(fixed_disk_device_t)
 
 neverallow ~{ fixed_disk_raw_read storage_unconfined_type } fixed_disk_device_t:{ chr_file blk_file } read;


policy/modules/kernel/storage.te
```diff
@@ -16,6 +16,8 @@ attribute storage_unconfined_type;
 # /dev/hd* and /dev/sd*.
 #
 type fixed_disk_device_t;
+# nvme_device_t was merged into fixed_disk_device_t in Fedora 34+
```
zpytela: It was actually not merged but rather renamed.

How about direct statement "will be removed in f36"? If we had some template for changes like this, we could look for them more easily at the right time for them.

WOnder93: > It was actually not merged but rather renamed.

At a high level it was merged into the existing `fixed_disk_device_t` type. But I could change it to "removed in favor" as in the commit message.

> How about direct statement "will be removed in f36"? If we had some template for changes like this, we could look for them more easily at the right time for them.

IMHO, "obsoleted in F34" already gives enough information to know when to remove it, given a strategy. I think it would be better to not hardcode the strategy itself in the comments, just the "input data".

zpytela: Agree this is better.

zpytela:
Thanks, merging.

PR #631, opened 2021-03-04
Fix typo in rsyncd.conf in rsync.if

HarmtH:


zpytela:
@HarmtH good catch. The typo is there since 2013, probably because the file is a part of the package so it is hardly any time created by `rsync_filetrans_named_content` callers. Thank you for your contribution.

PR #635, opened 2021-03-08
Allow to sendmail read/write kerberos host rcache files

Koncpa:
Create interface which allow read/write the kerberos host rcache files.
Use this interface to allow sendmail read/write kerberos host rcache
files.

COPR build: https://download.copr.fedorainfracloud.org/results/pkoncity/selinux-policy/fedora-rawhide-x86_64/02053957-selinux-policy/
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1893333


zpytela:
LGTM, thank you.

PR #636, opened 2021-03-08
Add kerberos object filetrans for nsswitchdomain

Koncpa:
Allow kerberos host rcache filetransition for nsswitch domain.
Filetranstion specify for which object is use, because domains from
local policies can use own general tmp filetrans.
Add for nsswitch domain, because all domains which use kerberos are
part of nsswitch domain.

COPR build: https://download.copr.fedorainfracloud.org/results/pkoncity/selinux-policy/fedora-rawhide-x86_64/02049910-selinux-policy/
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1878348

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -529,6 +529,9 @@ optional_policy(`
 # can not wrap nis_use_ypbind or kerberos_use, but they both have booleans you can turn off.
 optional_policy(`
 	kerberos_use(nsswitch_domain)
+	kerberos_tmp_filetrans_host_rcache(nsswitch_domain, "krb5_0.rcache2")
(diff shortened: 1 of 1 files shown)


zpytela:
I'd expect the new named transition added to `kerberos_filetrans_named_content()` in policy/modules/contrib/kerberos.if: was there any reason do it this way in policy/modules/system/authlogin.te instead, @Koncpa ?

Koncpa:
From my sight of view is more secure to use `kerberos_tmp_filetrans_host_rcache()` intefaces only for nsswitch_domain, then add new named transition to `kerberos_filetrans_named_content()` and  allow it in `nsswitch_domain`, because this named_content interface contain more permission than it's needed for `nsswitch_domain`. What do you suggest @zpytela ?

zpytela:
OK, accepted. thank you.

PR #637, opened 2021-03-08
Allow additional rules for use opencryptoki

Koncpa:
Commit 1:
The var.lib.opencryptoki.* files are located in /dev/shm/ are now labeled as
pkcs_slotd_tmpfs_t.
New interface pkcs_tmpfs_filetrans() allow create and manage objects in
the tmpfs directories with a private type (pkcs_slotd_tmpfs_t) and also
allow manage tmpfs dirs.
In pkcs_use_opencryptoki add permission allowing domain to tcp connect
tcs port.

Commit 2:
Certmonger and Ipsec need more permission for proper usage of opencryptoki.
Allow certmonger_t to get attributes of tmpfs (fs_getattr_tmpfs) and also certmonger allow
create and manage objects in the tmpfs directories with a private type
pkcs_slotd_tmpfs_t (pkcs_tmpfs_filetrans). Also certmonger can proper
use opencryptoki (pkcs_use_opencryptoki).

Allow ipsec_t change owner of file via chown capability.
Also ipsec allow create and manage objects in the tmpfs directories
with a private type pkcs_slotd_tmpfs_t (pkcs_tmpfs_filetrans).

COPR build: https://download.copr.fedorainfracloud.org/results/pkoncity/selinux-policy/fedora-rawhide-x86_64/02052708-selinux-policy/
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1894132

### DIFF
--- a/policy/modules/contrib/pkcs.if
+++ b/policy/modules/contrib/pkcs.if
@@ -118,6 +118,29 @@ interface(`pkcs_getattr_exec_files',`
 
 ########################################
 ## <summary>
+##	Create and manage objects in the tmpfs directories
+##	with a private type.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pkcs_tmpfs_filetrans',`
+	gen_require(`
+		type pkcs_slotd_tmpfs_t;
+	')
+	allow $1 pkcs_slotd_tmpfs_t:file map;
+
+	manage_files_pattern($1, pkcs_slotd_tmpfs_t, pkcs_slotd_tmpfs_t)
+	fs_tmpfs_filetrans($1, pkcs_slotd_tmpfs_t, { file dir })
+
+	fs_manage_tmpfs_dirs($1)
+')
+
+########################################
+## <summary>
 ##	Use opencryptoki services
 ## </summary>
 ## <param name="domain">
@@ -136,6 +159,8 @@ interface(`pkcs_use_opencryptoki',`
 
     kernel_search_proc($1)
     ps_process_pattern(pkcs_slotd_t, $1)
+
+    corenet_tcp_connect_tcs_port($1)
 
     dev_rw_crypto($1)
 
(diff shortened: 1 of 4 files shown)


WOnder93:
I think it would be better to use a named transition specifically for `var.lib.opencryptoki.tpm.root`, otherwise this could result in files e.g. in `/tmp` to get an unexpected label. Hopefully the services won't run as non-root in the near future, at least until I implement prefix/suffix-based filename transitions :)

Also, it would be nicer to split this into multiple commits (I would separate out at least the capability addition and the corenet_tcp_connect_tcs_port() call).

And please don;t forget to double-check the whitespace consistency in your patches ;P

zpytela:
The content looks good, I'd suggest escape dots in the file context specification, and use tabs for indentation in the new interface.
Consider splitting the commit and add reasons for the new permissions when they are available.
Please also change the commit message to make it more clear and comprehensible.

zpytela:
> I think it would be better to use a named transition specifically for `var.lib.opencryptoki.tpm.root`, otherwise this could result in files e.g. in `/tmp` to get an unexpected label. Hopefully the services won't run as non-root in the near future, at least until I implement prefix/suffix-based filename transitions :)

This is a transition for `/dev/shm` with the `tmpfs_t` type, so why are you worried about `/tmp`?

policy/modules/contrib/pkcs.if
```diff
@@ -116,6 +116,29 @@ interface(`pkcs_getattr_exec_files',`
...
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pkcs_tmpfs_filetrans',`
+	gen_require(`
+        type pkcs_slotd_tmpfs_t;
```
zpytela: Please replace 8 spaces with 2 tabs.

zpytela (changes requested):

zpytela:
Last two requests:
- split the "ipsec and certmonger" commit into two to ease backporting
- add a bugzilla reference to each commit

zpytela:
LGTM now, thank you.

PR #642, opened 2021-03-11
install_t: Allow NoNewPriv transition from systemd

travier:
Enable a subset of rpm-ostree commands to run as part of a systemd
service directly under an unprivilged user with DynamicUser.

This will be used for the rpm-ostree-countme service.

### DIFF
--- a/policy/modules/contrib/anaconda.te
+++ b/policy/modules/contrib/anaconda.te
@@ -85,6 +85,7 @@ allow install_t self:capability2 mac_admin;
 systemd_dbus_chat_localed(install_t)
 systemd_dbus_chat_logind(install_t)
 init_dbus_chat(install_t)
+init_nnp_daemon_domain(install_t)
 
 tunable_policy(`deny_ptrace',`',`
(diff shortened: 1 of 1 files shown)


travier:
This is for https://bugzilla.redhat.com/show_bug.cgi?id=1937404

zpytela:
@travier can you ensure no other permission is required, e. g. starting the service in selinux permissive mode?
```
setenforce 0
```

travier:
I got no additional warning running with permissive mode so far. We're discussing if this is the right approach and I'll confirm here.

cgwalters (approved):
Yeah, we need this in order to allow us to drop (DAC) privileges - IOW the overall security benefit here is very high for a small policy addition.

zpytela:
OK, merging then, thank you.

travier:
If possible, we would like to have that in Fedora 34 but this is not a blocker.

zpytela:
It will be a part of the very next f34 and f35 builds.

travier:
Thanks

PR #643, opened 2021-03-12
Allow systemd the audit_control capability conditionally

zpytela:
When the pam_tty_audit pam module is enabled for a session, systemd
needs the audit_control capability to be able to control kernel audit
configuration and and rules.

Without this capability, the systemd user service manager fails to
start, errors are logged and AVC denials audited.
While it is still possible to log in using console or ssh, graphical
display managers fail to start as they use the systemd user session to
spawn the greeter so using graphical user interface is not possible.

The capability is allowed when the init_audit_control boolean is turned on.
This boolean is off by default.

Resolves: rhbz#1861771


zpytela:
Merging based on local testing.

PR #645, opened 2021-03-16
Remove setuid/setgid capabilities from mysqld_t

ljavorsk:
It is not needed when systemd defines user option by itself


zpytela:
Lukasi, could you just replace delete with remove?

ljavorsk:
Changed

zpytela:
Thanks, I'll merge it once CI finishes.

PR #651, opened 2021-03-19
Allow local_login_t get attributes of tmpfs filesystems

zpytela:
This permission is required when the system booted with cgroups v1.

Resolves: rhbz#1898386

### DIFF
--- a/policy/modules/system/locallogin.te
+++ b/policy/modules/system/locallogin.te
@@ -113,6 +113,7 @@ files_create_home_dir(local_login_t)
 
 fs_search_auto_mountpoints(local_login_t)
 fs_getattr_cgroup(local_login_t)
+fs_getattr_tmpfs(local_login_t)
 fs_getattr_xattr_fs(local_login_t)
 
 storage_dontaudit_getattr_fixed_disk_dev(local_login_t)


jlebon:
Yes, seeing this too on f34-based FCOS where we're still using cgroups v1 (but will soon move to v2). FWIW, I think `sssd_t` needs the same treatment:

> Mar 24 00:41:04 localhost audit[716]: AVC avc:  denied  { getattr } for  pid=716 comm="sssd" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:sssd_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=0

Like the login error, this error also does not occur when using cgroups v2.

jlebon:
The weird thing about this is I'm not sure why the kernel thinks the context of `/sys/fs/cgroup` is `tmpfs_t`. systemd relabels `/sys/fs/cgroup` during initialization on switchroot, and injecting an `ExecStart=ls -laZ /sys/fs/cgroup` for `sssd.service` indicates that it is labeled correctly as `cgroup_t`. Unless it ignores the filesystem label and just always uses `tmpfs_t` for tmpfs devices?

ghost:
Jonathan Lebon ***@***.***> writes:

> The weird thing about this is I'm not sure why the kernel thinks the context of /sys/fs/cgroup is
> tmpfs_t. systemd relabels /sys/fs/cgroup during initialization on switchroot, and injecting an
> ExecStart=ls -laZ /sys/fs/cgroup for sssd.service indicates that it is labeled correctly as cgroup_t.
> Unless it ignores the filesystem label and just always uses tmpfs_t for tmpfs devices?

I think that is a cgroup v1 (and hybrid) systemd thing. systemd mounts a
tmpfs on /sys/fs/cgroup so that it can create symlinks there

This should not be applicable if you use cgroup v2 exclusively AFAIK

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

jlebon:
> I think that is a cgroup v1 (and hybrid) systemd thing. systemd mounts a tmpfs on /sys/fs/cgroup so that it can create symlinks there This should not be applicable if you use cgroup v2 exclusively AFAIK

Right, this is definitely from the v1/hybrid setup. What I'm saying is that regardless, the root of the mounted tmpfs is correctly labeled AFAICT.

zpytela:
@jlebon, you are right, the root directory is labeled correctly, but the denial is for the filesystem class.

jlebon:
> @jlebon, you are right, the root directory is labeled correctly, but the denial is for the filesystem class.

Gotcha, that makes sense now! I hadn't paid attention to the `tclass=filesystem` bit.

zpytela:
Merging so that it is in the next build. I will address it for sssd_t, too.

PR #652, opened 2021-03-20
Fix all deprecations with: make policy

maage:
I've included deprecate interfaces etc. to commit messages, allow easy verifications.

pp-files should be the same as without this patchset. base.conf is somewhat reordered.

I did fist three patches by hand.

Rest were just 1-1 replacements made by:
```|sh
#!/bin/bash

set -ex

while read -r a b; do

    [ "$a" ] || continue
    [ "$b" ] || continue

    # Ensure we are not handling regexps
    [[ ! "$a" =~ [\\/()] ]]
    [[ ! "$b" =~ [\\/()] ]]

    # Ensure replacement is the same as original sans comments
    atmp="$(mktemp)"
    btmp="$(mktemp)"
    macro-expander -M "$a(zzz_t)" | sed '/^#/d' > "$atmp"
    macro-expander -M "$b(zzz_t)" | sed '/^#/d' > "$btmp"
    cmp -s -- "$atmp" "$btmp" || {
        diff -u -- "$atmp" "$btmp"
        rm -- "$atmp" "$btmp"
        exit 1
    }
    rm -- "$atmp" "$btmp"

    # Just target files where deprecated entry was used.
    # Ensure we do not modify intenface definitions.
    git grep -l -- "$a" | \
        xargs -r perl -pi -e 's/\b(?<!`)\Q'"$a"'\E(?!'"'"')\b/'"$b"'/'

    printf "change deprecated %s -> %s\n\n" "$a" "$b" > "$atmp"
    git grep -l -- "$a" | xargs -r sed -nr "/(define|interface|template)\(\`$a'/,/^'\)/p" >> "$atmp"

    grep -E " $b[( ]" "$atmp"

    git commit -F "$atmp" .

done <<\EOF

auth_read_all_dirs_except_shadow files_list_non_auth_dirs
auth_read_all_files_except_shadow files_read_non_auth_files
auth_read_all_symlinks_except_shadow files_read_non_auth_symlinks

userdom_user_tmpfs_file userdom_user_tmp_file
userdom_user_tmpfs_content userdom_user_tmp_content
userdom_rw_user_tmpfs_files userdom_rw_user_tmp_files
userdom_read_user_tmpfs_files userdom_read_user_tmp_files
userdom_dontaudit_setattr_user_tmpfs userdom_dontaudit_setattr_user_tmp
userdom_manage_tmpfs_role userdom_manage_tmp_role

xserver_manage_xdm_tmp_dirs userdom_manage_user_tmp_dirs
xserver_relabel_xdm_tmp_dirs userdom_relabel_user_tmp_dirs
xserver_dontaudit_xdm_tmp_dirs userdom_dontaudit_setattr_user_tmp
xserver_search_xdm_tmp_dirs userdom_search_user_tmp_dirs

xserver_rw_xdm_tmp_files userdom_rw_user_tmp_files
xserver_manage_xdm_tmp_files userdom_manage_user_tmp_files
xserver_append_xdm_tmp_files userdom_append_user_tmp_files
xserver_read_xdm_tmp_files userdom_read_user_tmp_files
xserver_dontaudit_read_xdm_tmp_files userdom_dontaudit_read_user_tmp_files

xserver_create_xdm_tmp_sockets userdom_create_user_tmp_sockets
xserver_dontaudit_getattr_xdm_tmp_sockets userdom_dontaudit_user_getattr_tmp_sockets
xserver_xdm_tmp_filetrans userdom_user_tmp_filetrans

miscfiles_read_certs miscfiles_read_generic_certs
miscfiles_manage_cert_files miscfiles_manage_generic_cert_files

mmap_files_pattern mmap_exec_files_pattern

automount_exec_config files_exec_etc_files

EOF
```


WOnder93:
Thank you for the effort! I have just one comment: I think you should move the new userdom_*() calls out from the xserver `optional_policy()` blocks so that they are together with the other userdom_*() calls (if any). And a few other interface calls could be moved together with the other calls with the same prefix. Granted, it's extra work, but like this it's easier to miss them when browsing the policy files.

maage:
Okay, now there is next try. I tried to group userdom_ macros and as I did that, I noticed those modules had duplicate entries and I added also patches to clean up those too. Dupes are mostly dropped on modules where there was already modifications in this patch set.

I added also init.te enabled_mls -> enable_mls change.

And some gone entries and then some macros which get type as an arg, not attribute.

As there is those modifications, there is differencies in end results at least in CIL format.

PR #656, opened 2021-03-22
Makefile fixes and M4 optimization

maage:
First fix seusers dependency, then simplify all_interfaces.conf creation as tmp file is unnecessary.

Then try to create most directories only once and then rely on make. This should speed up compilation slightly.

Last one uses m4 frozen files to speed up compilation.
https://www.gnu.org/software/m4/manual/m4-1.4.15/html_node/Frozen-files.html#Frozen-files

In my machine normal: make -j is done around 1/3rd of a time than without this patchset.

There is no difference of the result files. Temp files differ, but they are not installed.


WOnder93:
This looks nice! Would you consider contributing this also to the upstream refpolicy [1]? We could then just cherry-pick the commits when it's merged there (though there might be significant differences so let's keep this PR open for now). It's true that we don't do that too often any more, but in case of the build infra I think it's better to not leave the original repo behind :)

[1] https://github.com/SELinuxProject/refpolicy

PR #660, opened 2021-03-24
Allow sssd_t get attributes of tmpfs filesystems

zpytela:
This permission is required when the system booted with cgroups v1.


jlebon (approved):
Makes sense to me!

zpytela:
@jlebon thanks for confirmation.

PR #679, opened 2021-04-06
Allow systemd-modules-load read/write tracefs files

zpytela:
Using the fs_rw_tracefs_files() interface, systemd-modules-load is also
allowed the confidentiality lockdown permission.

type=AVC msg=audit(1616133452.608:146): avc:  denied  { confidentiality }
for  pid=763 comm="systemd-modules" lockdown_reason="use of tracefs"
scontext=system_u:system_r:systemd_modules_load_t:s0
tcontext=system_u:system_r:systemd_modules_load_t:s0 tclass=lockdown permissive=0

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1207,6 +1207,8 @@ init_read_pid_files(systemd_modules_load_t)
 files_map_kernel_modules(systemd_modules_load_t)
 files_read_kernel_modules(systemd_modules_load_t)
 
+fs_rw_tracefs_files(systemd_modules_load_t)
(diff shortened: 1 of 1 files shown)


WOnder93:
I suspect this may be the same issue as https://bugzilla.redhat.com/show_bug.cgi?id=1945581

zpytela:
There is a subtle difference, this time it runs as a regular service and the purpose of systemd-modules-load is similar to kmod; surely we can wait.

zpytela:
Closing given the latest development in kernel removing the lockdown class.

PR #680, opened 2021-04-06
allow systemd-sleep to set timer for suspend-then-hibernate

MartinNowak:
- tested with `HibernateDelaySec=15min` in `/etc/systemd/sleep.conf`
- suspend-then-hibernate not working in F34 beta
  (systemd-sleep fails with `Error creating timerfd: Operation not permitted`)
- fixed by allowing wake_alarm capability

fixup of #426

Also see: [[Fedora 34 beta] suspend-then-hibernate fails with Error creating timerfd: Operation not permitted - Discussions in English / on using Fedora - Ask Fedora](https://ask.fedoraproject.org/t/fedora-34-beta-suspend-then-hibernate-fails-with-error-creating-timerfd-operation-not-permitted/13179)

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1333,6 +1333,8 @@ systemd_read_efivarfs(systemd_userdbd_t)
 #
 
 allow systemd_sleep_t self:capability sys_resource;
+# systemd-sleep needs to set timer for suspend-then-hibernate
+allow systemd_sleep_t self:capability2 wake_alarm;
 dontaudit systemd_sleep_t self:capability sys_ptrace;
 # systemd-sleep needs the permission to change sleep state
 allow systemd_sleep_t self:lockdown integrity;


zpytela:
@MartinNowak This is a new issue. Will you be able to isolate when this problem started to happen, like with which packages update?
Is allowing the `wake_alarm` permission enough to make the service working?

zpytela:
Could you also include the information regarding timerfd and HibernateDelaySec into the commit message?

MartinNowak:
> Is allowing the `wake_alarm` permission enough to make the service working?

Yes.

> @MartinNowak This is a new issue. Will you be able to isolate when this problem started to happen, like with which packages update?

Will give it a try. The respective changes are really old ([systemd@v243](https://github.com/systemd/systemd/commit/1bbbefe7a68059eb55d864c3e0e670d00269683a), [kernel@v4.8](https://github.com/torvalds/linux/commit/2895a5e5b3ae78d9923a91fce405d4a2f32c4309)), so unlikely to be the causes.

MartinNowak:
> Could you also include the information regarding timerfd and HibernateDelaySec into the commit message?

Done

zpytela:
I assume then this particular feature has not been used by anybody so far.

Thank you for your contribution, merging.

PR #686, opened 2021-04-07
Move (most) dist-git config files to the source repo

WOnder93:
See the commit messages for details. The last commit is just to demonstrate that the RPM build passes with the proposed corresponding dist-git changes.


WOnder93:
Corresponding dist-git PR:
https://src.fedoraproject.org/rpms/selinux-policy/pull-request/175

bachradsusi:
> It will also make it easier for other distributions seeking to package a Fedora-like SELinux policy.

Do you expect that other distribution would be happy with Fedora defaults and Fedora split to -minimum, -targeted and -mls?

WOnder93:
Not necessarily, but if the differences are small, they apply patches as usual or override specific files with dist-git versions, as it done now. If the differences are larger, they would either have their configs in dist-git like now, or we could add another level of nesting under `dist/` and maintain their configs alongside Fedora ones.

bachradsusi:
It will be merged in https://github.com/fedora-selinux/selinux-policy/pull/2345 Thanks!

PR #694, opened 2021-04-14
Allow nsswitch_domain read cgroup files

zpytela:
This permission is required when the systemd nss module is used
in nsswitch.conf for users or groups. The module checks whether
the current process is running in the root cgroup, or if rather
cgroup namespaces are in place.

Resolves: rhbz#1895061


zpytela:
Merging a cherry-pick request.

PR #695, opened 2021-04-15
Allow system dbusd read /var/lib symlinks

zpytela:
Resolves: rhbz#1949785


zpytela:
This is a draft PR. It has not been made clear yet if `/var/lib/flatpak/exports/share` will get a private type.

zpytela:
Regardless of the bz resolution, this permission seems to be reasonable.

PR #697, opened 2021-04-20
Allow ipsec the lockdown confidentiality permission

zpytela:
XFRM is an IP framework for transforming packets.
The concept is used in the IPsec protocol implementation.
The permission is required for the ipsec_t domain for pluto
to be able to redact XFRM SA secret.

Added into kernel:
https://www.spinics.net/lists/linux-security-module/msg39063.html

type=PROCTITLE msg=audit(04/20/2021 03:10:29.406:636) :
proctitle=/usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --nofork
type=SYSCALL msg=audit(04/20/2021 03:10:29.406:636) : arch=x86_64
syscall=write success=yes exit=40 a0=0xb a1=0x7ffe31e05000 a2=0x28
a3=0x7f3f8be843e0 items=0 ppid=1 pid=8102 auid=unset uid=root gid=root
euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none)
ses=unset comm=pluto exe=/usr/libexec/ipsec/pluto
subj=system_u:system_r:ipsec_t:s0 key=(null)
type=MAC_IPSEC_EVENT msg=audit(04/20/2021 03:10:29.406:636) : op=SAD-delete
auid=unset ses=unset subj=system_u:system_r:ipsec_t:s0 src=10.0.138.148
dst=10.0.138.68 spi=754605297(0x2cfa5cf1) res=yes
type=AVC msg=audit(04/20/2021 03:10:29.406:636) : avc:  denied  { confidentiality }
for  pid=8102 comm=pluto lockdown_reason="xfrm SA secret"
scontext=system_u:system_r:ipsec_t:s0 tcontext=system_u:system_r:ipsec_t:s0
tclass=lockdown permissive=0

### DIFF
--- a/policy/modules/system/ipsec.te
+++ b/policy/modules/system/ipsec.te
@@ -82,6 +82,7 @@ files_type(ipsec_mgmt_devpts_t)
 
 allow ipsec_t self:capability { net_admin dac_read_search dac_override setpcap sys_admin sys_nice net_raw setuid setgid };
 dontaudit ipsec_t self:capability sys_tty_config;
+allow ipsec_t self:lockdown confidentiality;
 allow ipsec_t self:process { getcap setcap getsched signal signull setsched sigkill };
 allow ipsec_t self:tcp_socket create_stream_socket_perms;
 allow ipsec_t self:udp_socket create_socket_perms;


WOnder93:
> The permission is required for the ipsec_t domain for pluto
> to be able to redact XFRM SA secret.

Actually, it's the other way around - the secret is redacted from the output *unless* (in case of SELinux) the domain has the lockdown::confidentiality permission granted.

Such AVC could be addressed in two ways:
1. If the process doesn't need to see the actual value of the secret, it should be just dontaudited.
2. If it needs to see the actual value for proper operation (unlikely), then it would need to be allowed.

This case is strange though, since it seems that this is triggered during a delete operation, where this shouldn't be hit... Do you have a reproducer for this AVC? (If it was found by an internal test, please point me to it via internal channels.)

WOnder93:
Ok, so after fighting with Beaker and my own stupidity, I finally managed to reproduce it and get a kernel backtrace... What happens here is actually that the userspace process sends a XFRM_MSG_ADDSA/XFRM_MSG_DELSA control message to add/delete an IPSec Security Association (IIUC), which then triggers a notification message that is broadcasted to listening userspace processes (possibly different from the calling process), and the redacting is applied on these broadcasted messages. So the SELinux check doesn't really make sense here and should be just skipped. I think this will need a similar kernel fix as in https://bugzilla.redhat.com/show_bug.cgi?id=1945581... I'll try to address both of these together upstream. In the meantime, this permission can be allowed (seems to make more sense in light of this discovery than dontauditing) in the policy if it's urgent; otherwise I'd suggest to waive these denials and wait for the kernel fix.

zpytela:
Closing given the latest development wrt the lockdown class.

PR #698, opened 2021-04-21
Add anon_inode class to the policy

zpytela:


policy/modules/kernel/domain.te
```diff
@@ -122,6 +122,7 @@ neverallow ~{ domain unlabeled_t } *:process *;
 #
 
 # read /proc/(pid|self) entries
+allow domain self:file create_file_perms;
```
wrabcak: @zpytela  
How is this  connected to anon_inode? 

CC @WOnder93

policy/modules/kernel/domain.te
```diff
@@ -122,6 +122,7 @@ neverallow ~{ domain unlabeled_t } *:process *;
 #
 
 # read /proc/(pid|self) entries
+allow domain self:anon_inode create_file_perms;
```
WOnder93: According to the documentation and the selinux-testsuite test, `create getattr ioctl read` should be the only permissions needed (for userfaultfd). `create_file_perms` map to `getattr create open`, which wouldn't be enough. I'd prefer if we create a new perm set (e.g. `use_userfaultfd_perms` or `userfaultfd_anon_inode_perms`) and use that here.

zpytela: I changed the commits accordingly. I am a bit surprised open is not among the permissions, but the test suite seems to test exactly this four-piece set.

PR #700, opened 2021-04-26
Allow systemd-coredump getattr nsfs files and net_admin capability

zpytela:
The net_admin capability is required by systemd-coredump to increase
the send buffer in setsockopt(); refer to socket(7):

       SO_SNDBUFFORCE (since Linux 2.6.14)
              Using this socket option, a privileged (CAP_NET_ADMIN)  process  can  perform  the
              same task as SO_SNDBUF, but the wmem_max limit can be overridden.

Resolves: rhbz#1952163


oturpe:
I tested this. Installing the rpms from actions fixes rhbz#1952163.

zpytela:
Merging, thanks for the confirmation.

PR #709, opened 2021-04-28
cockpit: allow cockpit socket to bind nodes

lnussel:
Looks like this setting is implicit with kerberos enabled.
cockpit.socket fails to start if kerberos_enabled=false

### DIFF
--- a/policy/modules/contrib/cockpit.te
+++ b/policy/modules/contrib/cockpit.te
@@ -52,7 +52,9 @@ can_exec(cockpit_ws_t,cockpit_session_exec_t)
 dev_read_urand(cockpit_ws_t) # for authkey
 dev_read_rand(cockpit_ws_t)  # for libssh
 
+# cockpit-ws allows connections on websm port
 corenet_tcp_bind_websm_port(cockpit_ws_t)
+corenet_tcp_bind_generic_node(cockpit_ws_t)
 
 # cockpit-ws can connect to other hosts via ssh
 corenet_tcp_connect_ssh_port(cockpit_ws_t)


lnussel:
this is my first encounter with selinux. Maybe @martinpitt  wants to confirm this is correct :)

martinpitt:
@lnussel: We just pulled in the cockpit SELinux policy into cockpit itself in https://github.com/cockpit-project/cockpit/pull/15707 . Once this is in all supported Fedoras, we'll send a PR to drop it from this selinux-policy project.

Can you please re-do this as a PR against cockpit? Thank you, and sorry for the confusion!

`corenet_tcp_bind_generic_node()` certainly sounds plausible, but we didn't run into any trouble with this in RHEL/Fedora. If it helps you on SUSE, then by all means let's pull it in.

lnussel:
> @lnussel: We just pulled in the cockpit SELinux policy into cockpit itself in [cockpit-project/cockpit#15707](https://github.com/cockpit-project/cockpit/pull/15707) . Once this is in all supported Fedoras, we'll send a PR to drop it from this selinux-policy project.
> 
> Can you please re-do this as a PR against cockpit? Thank you, and sorry for the confusion!

Sure, will do.
 
> `corenet_tcp_bind_generic_node()` certainly sounds plausible, but we didn't run into any trouble with this in RHEL/Fedora. If it helps you on SUSE, then by all means let's pull it in.

I had a brief look at Fedora 34. It has `kerberos_enabled` set to true by default. If you turn it off you see the same issue on Fedora.

lnussel:
closing in favor of https://github.com/cockpit-project/cockpit/pull/15758

PR #710, opened 2021-04-30
Ajusted path for SUSE based distros

jsegitz:
I want to upstream the changes we currently have for SUSE. Quite a few path adjustments. I have a few questions:
- Are you willing to take those?
- If so: do you want them wrapped in ifdef(`distro_suse', ` statements? Might be helpful on other systems too?
- Do you want individual pull requests per module or rather one for all .fc changes?

### DIFF
--- a/policy/modules/contrib/rabbitmq.te
+++ b/policy/modules/contrib/rabbitmq.te
@@ -85,6 +85,7 @@ kernel_dgram_send(rabbitmq_t)
 
 kernel_read_system_state(rabbitmq_t)
 kernel_read_fs_sysctls(rabbitmq_t)
+kernel_read_net_sysctls(rabbitmq_t)
 
 corecmd_exec_bin(rabbitmq_t)
 corecmd_exec_shell(rabbitmq_t)


zpytela:
@jsegitz yes, we want to cover suse-specific configuration, too.

The policy regarding `ifdef` usage is not strict. However, if there is a path clash with FHS, I'd like to have them inside an ifdef block. Currently, I am not aware of other distros using Fedora selinux-policy.

If the changeset is not "million lines", I see the benefit of having related changes in one PR.

jsegitz:
First batch :) I added 
ifdef(`distro_suse', `
where I thought it was unlikely that other users would benefit.

I have other, non filesystem label, changes that I'll submit over time

zpytela:
@jsegitz I have one general note: fixfiles converts regular expressions to wildcards. It has unexpected consequences especially when the regexp is early in the file path as restorecon is called for the files with changed context on the package update, so I'd like to avoid changes like
 ```
-/usr/libexec/geoclue           --      gen_context(system_u:object_r:geoclue_exec_t,s0)
+/usr/lib(exec)?/geoclue                --      gen_context(system_u:object_r:geoclue_exec_t,s0)
```
to prevent from large filesystem relabeling.

This thing seems to me not to be completely consistent: sometimes the changes to lib instead of libexec are wrapped with ifdef, sometimes not - is it based on some experience?

Conan-Kudo:
Note that openSUSE Tumbleweed uses `/usr/libexec` instead of `/usr/lib`, so a lot of these are probably not going to last beyond the SLE 15 code stream.

PR #713, opened 2021-05-03
Kernel perf interfaces

zpytela:


WOnder93:
Looks good :+1: Although note that the confidentiality permission will (most likely) not be needed after this kernel patch:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=08ef1af4de5fe7de9c6d69f1e22e51b66e385d9b

I have requested for it to be backported to stable kernels, so we should add a note to the interfaces to remove the confidentiality permission after some grace period (say, in F36 after F35 gets branched off):
https://lore.kernel.org/selinux/CAFqZXNvrYW+oqBez6wUxWPbMiMSPBbNnvVNpDXST47-i0qS6ZQ@mail.gmail.com/T/

WOnder93 (approved):

zpytela:
A comment in the interface added.

PR #718, opened 2021-05-05
Allow NetworkManager watch its config dir

zpytela:
This permission is required for the inotify_add_watch() syscall on the
/etc/NetworkManager/VPN directory which does not exist by default.
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/8232c3473f93a87e2eb736bf4b63049001a20926/src/core/vpn/nm-vpn-manager.c#L217

Resolves: rhbz#1934725


WOnder93 (commented):
How about `watch_dirs_pattern(NetworkManager_t, NetworkManager_etc_t, NetworkManager_etc_t)`?

zpytela:
Surely a better option.

WOnder93 (approved):

PR #719, opened 2021-05-05
policy module for rtorrent

jsegitz:


policy/modules/contrib/rtorrent.if
```diff
@@ -0,0 +1,111 @@
...
+	# transition from the userdomain to the derived domain
+	domtrans_pattern($2, rtorrent_exec_t, rtorrent_t)
+
+	# allow ps to show rtorrent
+	ps_process_pattern($2, rtorrent_t)
+	allow $2 rtorrent_t:process { signull sigstop signal sigkill };
+
+	ifdef(`hide_broken_symptoms',`
+		#Leaked File Descriptors
+		dontaudit rtorrent_t $2:fifo_file rw_fifo_file_perms;
```
ghost, line 35: This does not make sense as this is allowed by default in the `domtrans_pattern()` here: https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/misc_patterns.spt#L37

That is, without the "open" (ie `rw_inherited_fifo_file_perms`.

If it really was a "#Leaked File Descriptors" then it wouldnt need the "open" permission provided by `rw_fifo_file_perms` in the first place.

policy/modules/contrib/rtorrent.if
```diff
@@ -0,0 +1,111 @@
...
+')
+
+######################################
+## <summary>
+##  Make rtorrent an entrypoint for
+##  the specified domain.
+## </summary>
+## <param name="domain">
+##  <summary>
+##  The domain for which cifs_t is an entrypoint.
```
ghost, line 83: Copy and paste issue "cifs_t" but you seem to not be using this interface anyway so why add it?

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+gen_tunable(rtorrent_send_mails, false)
+
+## <desc>
+## <p>
+## Enable necessary permissions for rutorrent
+## </p>
+## </desc>
+gen_tunable(rtorrent_enable_rutorrent, false)
+
+attribute rtorrentdomain;
```
ghost, line 21: unused type attribute

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+
+attribute_role rtorrent_roles;
+roleattribute system_r rtorrent_roles;
```
ghost, line 24: You should not directly reference `system_r` here. In which scenario is `rtorrent` run by "the system"?

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+role rtorrent_roles types rtorrent_t;
+
+########################################
+#
+# rtorrent local policy
+#
+
+type rtorrent_port_t;
+corenet_port(rtorrent_port_t)
+allow rtorrent_t rtorrent_port_t:tcp_socket name_bind;
```
ghost, line 38: the port declaration should go into `corenetwork` module, and then you cal call here:
```
corenet_tcp_bind_rtorrent_ports(rtorrent_t)
```

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+
+userdom_read_user_home_content_symlinks(rtorrent_t)
+
+allow rtorrent_t self:process setpgid;
+allow rtorrent_t self:netlink_audit_socket { create_socket_perms nlmsg_relay };
```
ghost, line 43: That does not make sense. This suggests that `rtorrent` logs to audit, but its incomplete because you need access to `CAP_AUDIT_WRITE` to be able to log to audit.

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+allow rtorrent_t self:fifo_file rw_fifo_file_perms;
+allow rtorrent_t self:tcp_socket create_stream_socket_perms;
+allow rtorrent_t self:unix_stream_socket connectto;
```
ghost, line 46: This is questionable. Where is the `sock_file` write access used to connect?

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+
+allow rtorrent_t self:netlink_route_socket { bind create nlmsg_read };
+allow rtorrent_t self:udp_socket { connect create getattr };
+nscd_shm_use(rtorrent_t)
```
ghost, line 50: These above 3 lines are provided with the more suitable `auth_use_nsswitch()` interface. Call that instead.

ghost: O, nevermind you called it below. So these 3 lines are redundant and can be removed.

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+corenet_udp_sendrecv_generic_if(rtorrent_t)
+corenet_tcp_sendrecv_generic_node(rtorrent_t)
+corenet_udp_sendrecv_generic_node(rtorrent_t)
+corenet_tcp_sendrecv_all_ports(rtorrent_t)
+corenet_udp_sendrecv_all_ports(rtorrent_t)
+corenet_tcp_connect_all_ports(rtorrent_t)
+corenet_sendrecv_all_client_packets(rtorrent_t)
+corenet_udp_bind_all_unreserved_ports(rtorrent_t)
+
+domain_use_interactive_fds(rtorrent_t)
```
ghost, line 68: redundant, already provided with `userdom_user_application_domain()`

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+userdom_manage_user_home_content_files(rtorrent_t)
+userdom_manage_user_home_content_dirs(rtorrent_t)
+userdom_home_manager(rtorrent_t)
+userdom_filetrans_home_content(rtorrent_t)
+userdom_stream_connect(rtorrent_t)
+
+optional_policy(`
+	tunable_policy(`rtorrent_send_mails',`
+		userdom_exec_user_bin_files(rtorrent_t)
+		userdom_exec_user_home_content_files(rtorrent_t)
```
ghost, line 83: Why is this needed, can this content not just be labeled `home_bin_t` instead?

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+		files_manage_generic_tmp_files(rtorrent_t)
```
ghost, line 84: This is suboptimal, there should not be files with the generic `tmp_t` type ideally

policy/modules/contrib/rtorrent.te
```diff
@@ -0,0 +1,95 @@
...
+		mta_send_mail(rtorrent_t)
+	')
+')
+
+optional_policy(`
```
ghost, line 89: This whole block seems questionable to me. rtorrent is a user application (AFAIK). Why would the system run this app and why would this app need to manage system content?

jsegitz:
Thank you for the many notes. I'll work through them and test the resulting module before pushing a new version

jsegitz:
I currently don't find time to do this. I'll sent a new pull request once I have a module in a better state

PR #724, opened 2021-05-06
Remove legacy anon_inodefs and inotifyfs rules

WOnder93:
inotifyfs has been dropped in favor of anon inodes and anon_inodefs inodes have been excluded from LSM checking (except the ones now checked via the new `anon_inode` class) for a very long time.

My workstation has been running with a policy including these patches for a few days now. Not a single related AVC observed.


zpytela:
LGTM, thank you.

Providing the kernel commits links for the convenience:
https://github.com/torvalds/linux/commit/3836a03d978e
https://github.com/torvalds/linux/commit/c44dcc56d2b5

PR #728, opened 2021-05-10
Allow privoxy execmem

zpytela:
The execmem permission is required for mapping a memory region as
executable. This is not common and is possibly insecure so it is
disabled by default in selinux-policy.

Just-in-time (JIT) compiling is a heavyweight optimization that can
greatly speed up pattern matching. JIT support applies only to the
traditional Perl-compatible regular expressions (pcre).

Privoxy 3.0.29 introduces JIT support for pcre which requires the execmem
permission and cannot be disabled dinamically, just at compile time.

Resolves: rhbz#1917099


WOnder93 (approved):

zpytela:
An ugly typo fixed.

PR #734, opened 2021-05-12
Allow fcoemon create sysfs files

zpytela:
This permission is required when fcoemon wants to write to the
/sys/bus/fcoe/devices/ctlr_X/enabled file and this file does not exist.

Resolves: rhbz#1952292


WOnder93 (approved):
I have checked that fcoemon really creates this file. I was wondering if it didn't get created indirectly by a kernel module, but not. I find it strange that the kernel didn't create this file along with the rest of the device entry, but it seems to be expected...

PR #747, opened 2021-05-19
Cleanup setrlimit permission for executing systemctl

WOnder93:
Rather than adding this permission one by one to individual domains, it should have been added to the interface instead (as dontaudit),


zpytela:
LGTM, thank you.

PR #750, opened 2021-05-24
Label /dev/trng with random_device_t

zpytela:
The trng device (trng stands for "true random-number generator")
is used on some architectures like s390x where the Central Processor
Assist for Cryptographic Functions (CPACF) with Message-Security-Assist
Extension 7 (MSA 7) is available to generate high-entropy random numbers.
The label for the device node is backed by a file transition.


WOnder93 (approved):
Looks good. https://www.ibm.com/docs/en/linux-on-systems?topic=t-setup also mentions /dev/hwrng, but that one is already covered.

PR #754, opened 2021-05-26
Allow nsswitch_domain read init pid lnk_files

zpytela:
This permission is required when a service uses the systemd nss module
to resolve user records and a symlink in the /run/systemd/dynamic-uid
directory needs to be dereferenced.

type=PROCTITLE msg=audit(07/27/2020 09:06:58.618:335) : proctitle=/usr/bin/dbus-daemon
--system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
type=PATH msg=audit(07/27/2020 09:06:58.618:335) : item=0
name=/run/systemd/dynamic-uid/direct:62803 inode=38503 dev=00:18 mode=link,777 ouid=root
ogid=root rdev=00:00 obj=system_u:object_r:init_var_run_t:s0 nametype=NORMAL cap_fp=none
cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(07/27/2020 09:06:58.618:335) : cwd=/
type=SYSCALL msg=audit(07/27/2020 09:06:58.618:335) : arch=x86_64 syscall=readlinkat
success=no exit=EACCES(Permission denied) a0=0xffffff9c a1=0x7ffe508dcb30
a2=0x55974a9ca6f0 a3=0x63 items=1 ppid=1 pid=673 auid=unset uid=dbus gid=dbus euid=dbus
suid=dbus fsuid=dbus egid=dbus sgid=dbus fsgid=dbus tty=(none) ses=unset
comm=dbus-daemon exe=/usr/bin/dbus-daemon
subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 key=(null)
type=AVC msg=audit(07/27/2020 09:06:58.618:335) : avc:  denied  { read }
for  pid=673 comm=dbus-daemon name=direct:62803 dev="tmpfs" ino=38503
scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023
tcontext=system_u:object_r:init_var_run_t:s0 tclass=lnk_file permissive=0

Resolves: rhbz#1860924

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -525,6 +525,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	init_read_pid_lnk_files(nsswitch_domain)
+')
+
+optional_policy(`
 	likewise_stream_connect_lsassd(nsswitch_domain)
 ')
 
--- a/policy/modules/system/init.if
+++ b/policy/modules/system/init.if
@@ -2891,6 +2891,24 @@ interface(`init_manage_pid_files',`
 	manage_files_pattern($1, init_var_run_t, init_var_run_t)
 ')
 
+#######################################
+## <summary>
+##	Read init pid lnk_files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`init_read_pid_lnk_files',`
+	gen_require(`
+		type init_var_run_t;
+	')
+
+	read_lnk_files_pattern($1, init_var_run_t, init_var_run_t)
+')
+
 ########################################
 ## <summary>
 ##	Read init unnamed pipes.


WOnder93 (approved):

ghost:
Ondrej Mosnáček ***@***.***> writes:

> @WOnder93 approved this pull request.

I think this is questionable. this seems dbus-daemon (may or may not
apply to dbus-broker as well) specific. it is
also old and it may be that this code has changed (earlier systemd used
dbus for dynamic user resolving (this might be related to that) nowaday's
is uses varlink instead. I suspect that this no longer applies and if it
does then it might only applies to dbus (we would have seen more related
reports, i mean why would this be reported and nothing else, there are
surely many more nss clients that resolve uids)

Aside, its reading an inherited symlink and allowing this pattern is
probably too much there is no need to traverse /run/systemd and no need
to open the lnk file.

Its not a big deal but auth_use_nsswitch is widely used.

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

zpytela:
Discussed with systemd folks internally:
- it is valid in current Fedora,
- any domain using nss can run into this issue.

You are right open is not needed, but it was not allowed: the interface calls read_lnk_files_pattern which means `{ getattr read }`. I could have named the interface read_inherited_... Traversing /run/systemd was already allowed:
```
# sesearch -A -s system_dbusd_t -t init_var_run_t -c dir
allow nsswitch_domain pidfile:dir { getattr open search };
allow system_dbusd_t init_var_run_t:dir { ioctl lock read };
allow systemprocess init_var_run_t:dir { getattr open search };
```

ghost:

> Discussed with systemd folks internally:
>
> * it is valid in current Fedora,
> * any domain using nss can run into this issue.

Okay it is not bad to have that clarified

>
> You are right open is not needed, but it was not allowed: the interface calls read_lnk_files_pattern which means {
> getattr read }. I could have named the interface read_inherited_... Traversing /run/systemd was already allowed:

My bad, youre right open is not checked at all on symlinks I did not
expect that.

>
> # sesearch -A -s system_dbusd_t -t init_var_run_t -c dir
> allow nsswitch_domain pidfile:dir { getattr open search };
> allow system_dbusd_t init_var_run_t:dir { ioctl lock read };
> allow systemprocess init_var_run_t:dir { getattr open search };
>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

PR #757, opened 2021-05-27
Drop cockpit

marusak:
Cockpit is now shipping its own policy.


marusak:
```
policy/modules/contrib/certmonger.te:125:ERROR 'syntax error' at token 'cockpit_manage_pid_dirs' on line 6696:
    cockpit_manage_pid_dirs(certmonger_t)
```

build fails. I will need some guidance.
This is declared in `policy/modules/contrib/cockpit.if`. Should I keep this in this file or should I replace it where it is being used? And then we can probably drop this from [cockpit directly](https://github.com/cockpit-project/cockpit/blob/master/selinux/cockpit.if)?

marusak:
@zpytela you might know whats the best approach?

zpytela:
@marusak we don't have a general answer or guidance or yet. There are 2 options:
- keep the module in selinux-policy
- include ifndef statements into the other custom policy module (certmonger in this case), similar to https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Backwards_compatibility

each of them with pros and cons.

Would you like to have the cockpit module removed from selinux-policy? Anyway in that case we recommend some transition period (1 release).

@vmojzis here we have an example which needs to be covered in the documentation.

marusak:
> Would you like to have the cockpit module removed from selinux-policy? Anyway in that case we recommend some transition period (1 release).

Cockpit ships these selinux rules on its own, now for already a couple of releases so we think there is no point for selinux-policy to ship it as well.

> include ifndef statements into the other custom policy module (certmonger in this case)

We don't ship any rules for certmonger - wouldn't that then mean that certmonger would stop working with selinux enforcing?

zpytela:
> > Would you like to have the cockpit module removed from selinux-policy? Anyway in that case we recommend some transition period (1 release).
> 
> Cockpit ships these selinux rules on its own, now for already a couple of releases so we think there is no point for selinux-policy to ship it as well.

You are right, we just need to figure out the right approach for modules in selinux-policy which depend on cockpit.

> 
> > include ifndef statements into the other custom policy module (certmonger in this case)
> 
> We don't ship any rules for certmonger - wouldn't that then mean that certmonger would stop working with selinux enforcing?

When cockpit is installed, certmonger needs some rules to access cockpit types, which is done using cockpit interfaces in certmonger policy.

martinpitt:
I had a look at this particular rule:
```
optional_policy(`
    cockpit_manage_pid_dirs(certmonger_t)
    cockpit_manage_pid_files(certmonger_t)
')
```
This was introduced in commit dab4b50b7d and 9c5ff0923a 3 years ago, but unfortunately there is neither a justification nor a reference. I honestly don't see any reason to give any program other than cockpit-{tls,ws} write access to /run/cockpit. There are two kinds of files there:

  * dynamically generated motd (symlinked from /etc/motd.d/cockpit, read by PAM)
  * wsinstance/*.sock from cockpit-tls, an internal implementation detail.

There is nothing at all certificate related there. It *does* make sense to give certmonger access to /etc/cockpit/ws-certs.d/, so that you can manage its TLS certificate  -- but that's not what that rule does. So my gut feeling is that we should just  remove that bit entirely.

However, that's not the only place where policies refer to cockpit:
```
policy/modules/services/ssh.te:    cockpit_read_pid_files(sshd_t)
policy/modules/services/ssh.te:    cockpit_rw_pipes(ssh_agent_type)
policy/modules/system/getty.te:    cockpit_read_pid_files(getty_t)
policy/modules/system/userdomain.te:    cockpit_manage_unix_stream_sockets(userdomain)
```
Plus a bunch of user role privileges like

    config/appconfig-standard/sysadm_u_default_contexts:system_r:cockpit_session_t:s0               sysadm_r:sysadm_t:s0

These are all legit. I suppose for these we still need to keep a stripped down version of the cockpit policy with the interfaces above, and then re-use them from cockpit itself instead of copying them?

zpytela:
> I had a look at this particular rule:
> 
> ```
> optional_policy(`
>     cockpit_manage_pid_dirs(certmonger_t)
>     cockpit_manage_pid_files(certmonger_t)
> ')
> ```
> 
> This was introduced in commit [dab4b50](https://github.com/fedora-selinux/selinux-policy/commit/dab4b50b7d2268b6cfb675754903b1a413008bba) and [9c5ff09](https://github.com/fedora-selinux/selinux-policy/commit/9c5ff0923ac5149a038dfe61693bd4a1b107c5f2) 3 years ago, but unfortunately there is neither a justification nor a reference. I honestly don't see any reason to give any program other than cockpit-{tls,ws} write access to /run/cockpit. 

I happened to succeed in looking for the change reason, it is https://bugzilla.redhat.com/show_bug.cgi?id=1629685 (a private bz, created by guess-who). The justification is "Cockpit's ability to request and install an SSL certificate for its webserver from FreeIPA". Possibly the manage permissions can be lowered to create files in the cockpit runtime directory?

> I suppose for these we still need to keep a stripped down version of the cockpit policy with the interfaces above, and then re-use them from cockpit itself instead of copying them?

We need to come up with a general solution in our team.

martinpitt:
Haha, thanks for digging out the certs-in-/run/cockpit bz -- indeed I remember. This is still current, although we can now in principle make this easier.

policy/modules/contrib/certmonger.te
```diff
@@ -124,8 +124,12 @@ optional_policy(`
 ')
 
 optional_policy(`
-    cockpit_manage_pid_dirs(certmonger_t)
-    cockpit_manage_pid_files(certmonger_t)
+	gen_require(`
```
marusak, line 127: Does this approach make sense?

marusak: Or maybe should I move these into cockpit.te in cockpit repo?

Similar to https://github.com/cockpit-project/cockpit/pull/16223

marusak: @zpytela @vmojzis WDYT?

vmojzis:
Sorry for the late reply.
We finally put together a solution that will be painless for all DSP adopters. A copy of the interface file will stay in this repository, with all interfaces enclosed in "ifndef" statements. This way we are not introducing a regression for anyone using any of the interfaces in their custom policy.
See https://github.com/fedora-selinux/selinux-policy/pull/853

marusak:
Closing this as there is now #853 
Thank you @vmojzis for taking over!

PR #758, opened 2021-05-27
Allow iscsid read all process stat

HuaxinLu:
iscsid needs to read all process stat to find and set the priority of one thread.
My environment:
```
# cat /etc/fedora-release 
Fedora release 34 (Thirty Four)
```
Use iscsiadm to connect a target node:
```
# iscsiadm -m discovery -t st -p xxx -l
```
Then watch systemd journal:
```
......
fedora audit[2529]: AVC avc:  denied  { search } for  pid=2529 comm="iscsid" name="2496" dev="proc" ino=70130 scontext=system_u:system_r:iscsid_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=dir permissive=0
fedora audit[2529]: AVC avc:  denied  { search } for  pid=2529 comm="iscsid" name="2497" dev="proc" ino=70131 scontext=system_u:system_r:iscsid_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=dir permissive=0
fedora audit[2529]: AVC avc:  denied  { search } for  pid=2529 comm="iscsid" name="2603" dev="proc" ino=38486 scontext=system_u:system_r:iscsid_t:s0 tcontext=system_u:system_r:rpm_t:s0 tclass=dir permissive=0
fedora audit[2529]: AVC avc:  denied  { search } for  pid=2529 comm="iscsid" name="3392" dev="proc" ino=70132 scontext=system_u:system_r:iscsid_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=dir permissive=0
fedora iscsid[2529]: iscsid: Could not set session1 priority. READ/WRITE throughout and latency could be affected.
......
```
After install a policy module as follow, the problem can be solved:
```
# cat demo.te 
policy_module(demo, 1.0)

require
{
	type iscsid_t;
};

domain_read_all_domains_state(iscsid_t)
```

The relevant code:
https://github.com/open-iscsi/open-iscsi/blob/master/usr/initiator.c
function session_increase_wq_priority

### DIFF
--- a/policy/modules/contrib/iscsi.te
+++ b/policy/modules/contrib/iscsi.te
@@ -105,7 +105,7 @@ dev_rw_userio_dev(iscsid_t)
 dev_map_userio_dev(iscsid_t)
 
 domain_use_interactive_fds(iscsid_t)
-domain_dontaudit_read_all_domains_state(iscsid_t)
+domain_read_all_domains_state(iscsid_t)
 
 files_read_kernel_modules(iscsid_t)
 files_map_kernel_modules(iscsid_t)


policy/modules/contrib/iscsi.te
```diff
@@ -120,6 +120,7 @@ modutils_read_module_config(iscsid_t)
 modutils_read_module_deps_files(iscsid_t)
 
 mount_read_pid_files(iscsid_t)
+domain_read_all_domains_state(iscsid_t)
```
WOnder93: There is already `domain_dontaudit_read_all_domains_state(iscsid_t)` a few lines above - you should just replace that line with this one. (The `domain_*()` calls should be all together in one block.)

HuaxinLu: Thanks. I have fixed it. This rule was not added before, probably because the failure of priority setting will not affect the function.

zpytela:
@HuaxinLu could you also put some relevant information into the commit message? The avc denial, link to the source code, reason why it was previously sufficient to dontaudit and now not, or anything.

HuaxinLu:
> @HuaxinLu could you also put some relevant information into the commit message? The avc denial, link to the source code, reason why it was previously sufficient to dontaudit and now not, or anything.

When I submitted this code, I did not notice the above dontaudit rules. So this may be a question that has already been considered.

iscsid will increase a xmit thread priority for performance [code here](https://github.com/open-iscsi/open-iscsi/commit/175a817ac84d3651d38bebb55d03d62f6e45e370). The current selinux policy prohibits this behavior, but it will not affect the function of iscsid. I just found a fail log in systemd journal and add the missing rules.

On the other hand, reading all process state may be an excessive permission.

zpytela:
Merging, thanks.

PR #771, opened 2021-06-04
Revert "Remove openhpi module"

zpytela:
This reverts commit 1e6afd814132ddb080635307bfa24ef728e72515.

There should be a transition period, so it should not be done
during F34 development phase.


zpytela:
No need for a transition period, the module is not active since a long time ago.

PR #782, opened 2021-06-16
Allow kdump create /var/lib/kdump with a transition

zpytela:
With commit af7e4b6492b ("Label /var/lib/kdump with kdump_var_lib_t"),
the kdump service was allowed to write initramfs files in case /boot
was read only. In that case though the kdump service creates
the /var/lib/kdump directory as it is not shipped in the rpm package,
so it requires appropriate permissions, together with a named file
transition, too.


zpytela:
Made this PR a draft as it is possible the directory will be packaged soon.

zpytela:
This directory is a part of new versions of kdump, so this commit is not needed any longer.

PR #784, opened 2021-06-17
Label /usr/bin/arping plain file with netutils_exec_t

zpytela:
In iputils newer than iputils-20180629, the arping command moved
from /usr/sbin to /usr/bin, keeping the symlink in sbin.

### DIFF
--- a/policy/modules/admin/netutils.fc
+++ b/policy/modules/admin/netutils.fc
@@ -7,6 +7,7 @@
 /bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
 /sbin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
 
+/usr/bin/arping		--	gen_context(system_u:object_r:netutils_exec_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
Aligned with other similar commands, this is a duplicated entry in Fedora since the policy defines "/sbin = /usr/sbin" equivalency, but not the same for /bin. The proper solution would consist of another equivalency rule and removing all duplicates. I plan to do it later, but in rawhide only.

ghost:

> Aligned with other similar commands, this is a duplicated entry in Fedora since the policy defines "/sbin =
> /usr/sbin" equivalency, but not the same for /bin. The proper solution would consist of another equivalency rule
> and removing all duplicates. I plan to do it later, but in rawhide only.

You could take that further (it is what i do) and make /bin /sbin
/usr/sbin /usr/libexec /usr/local/bin /usr/local/sbin /usr/local/libexec all
equivalents of that single /usr/bin

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

PR #763, opened 2021-05-31
Label /dev/dma_heap with dma_device_dir_t

zpytela:
With commit a091bcda8d9 (Label /dev/dma_heap/* char devices with dma_device_t)
a new dma_device_t type was assigned to the /dev/dma_heap directory
and all files in it. The basic dev_node() interface just assigns the
type to the device_node attribute, which prevents many domains from
searching the directory.

This commits adds the dma_device_t type to file_type, non_auth_file_type,
and non_security_file_type attributes which should allow the access for
domains requiring this access.

An example AVC denial:

type=PROCTITLE msg=audit(05/31/2021 09:03:08.452:397) :
proctitle=/usr/bin/python3 -Es /usr/sbin/setroubleshootd -f
type=PATH msg=audit(05/31/2021 09:03:08.452:397) : item=0 name=/dev/dma_heap/*
nametype=UNKNOWN cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=SYSCALL msg=audit(05/31/2021 09:03:08.452:397) : arch=x86_64 syscall=newfstatat
success=no exit=EACCES(Permission denied) a0=0xffffff9c a1=0x7fd607775ec0
a2=0x7fd60774bf60 a3=0x0 items=1 ppid=1 pid=2498 auid=unset uid=setroubleshoot
gid=setroubleshoot euid=setroubleshoot suid=setroubleshoot fsuid=setroubleshoot
egid=setroubleshoot sgid=setroubleshoot fsgid=setroubleshoot tty=(none) ses=unset
comm=setroubleshootd exe=/usr/bin/python3.9 subj=system_u:system_r:setroubleshootd_t:s0
type=AVC msg=audit(05/31/2021 09:03:08.452:397) : avc:  denied  { search }
for  pid=2498 comm=setroubleshootd name=dma_heap dev="devtmpfs" ino=102
scontext=system_u:system_r:setroubleshootd_t:s0
tcontext=system_u:object_r:dma_device_t:s0 tclass=dir permissive=0

Resolves: rhbz#1965743

PR #787, opened 2021-06-18
Label /dev/dma_heap/* char devices with dma_device_t 

zpytela:


zpytela:
Taking the discussion in https://github.com/fedora-selinux/selinux-policy/pull/763 and the current Fedora state into account, I decided to revert the current solution in favor of the least complex one, and not label /dev/dma_heap with a private, unique type.

I am afraid though we will need to get back to it later unless we decide to drop all transitions as also discussed. There already is a request for `/dev/pktcdvd/[0-7]` support and this will hardly be a unique filename.

Note to myself: In that case, a dozen of interfaces like dev_manage_all_dev_nodes will need to be updated.

zpytela:
Testing the new build, I haven't run into any issue, so undrafting.

PR #799, opened 2021-07-07
allow init watch all files

zpytela:


rhatdan:
LGTM

rhatdan:
@wrabcak PTAL

zpytela:
Merging. Frankly I expected an objection non_security_file_type is not enough, so just putting it here as a comment.

PR #804, opened 2021-07-16
Allow systemd-coredump read and write usermodehelper state

zpytela:
Resolves: rhbz#1982961

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1045,6 +1045,8 @@ manage_files_pattern(systemd_coredump_t, systemd_coredump_var_lib_t, systemd_cor
 mmap_files_pattern(systemd_coredump_t, systemd_coredump_var_lib_t, systemd_coredump_var_lib_t)
 init_var_lib_filetrans(systemd_coredump_t, systemd_coredump_var_lib_t, dir, "coredump")
 
+kernel_rw_usermodehelper_state(systemd_coredump_t)
+
 dev_write_kmsg(systemd_coredump_t)
 
 # To read info about the crashed process from /proc


ghost:
might be a leaked file descriptor? There is no open and read mentioned in the bugzilla

ghost:
Yes... i suspect this is wrong as systemd-coredump@.service is configured with `ProtectKernelTunables=yes`.

To me that at least indicates that systemd thinks that systemd-coredump has no business in `/proc/sys`.

zpytela:
Making this PR draft, had a chat with local systemd folks, needs to be clarified in upstream.

zpytela:
Very likely it is an unimportant bug in systemd code or a leaked file descriptor. systemd-coredump is only expected to read core_pattern.

keszybz:
Actually this pull request is good: systemd-coredump will write the `core_pattern` file when it detects that it is PID 1 that has crashed. If PID1 crashes it will freeze and systemd services cannot be started, so coredump handling with systemd-coredump will not work either. And frozen PID1 does not collect zombies. So we want to avoid spawning further processes as much as possible. Thus we try to disable coredumping by adjusting `core_pattern` to something like `|/bin/false`.

Please reopen.

ghost:
Excuse my ignorance but how would that work when it has ProtectKernelTunables=yes set in https://github.com/systemd/systemd/blob/main/units/systemd-coredump%40.service.in#L34

ghost:
Never mind, I think it starts to dawn on me that coredump is not run by pid1 (its spawned by core_pattern entry) and so that service unit does not apply in this scenario.

ghost:
... and there is no open because the fd is probably inherited from that same task that read /proc/sys/kernel/core_pattern in the first place.

zpytela:
Not reopened, but continuing here:
https://github.com/fedora-selinux/selinux-policy/pull/999

PR #810, opened 2021-07-28
Allow mdadm read iscsi pid files

zpytela:
The mdadm command needs to read /run/initiatorname.iscsi file which is
created by dracut-cmdline executed from initramfs. After loading SELinux
policy and switchroot, /run is relabeled, so no file transition needs to
be set, having the default file context entry should be enough.

The iscsi_read_pid_files() interface was added.

type=AVC msg=audit(22.1.2021 18:23:19.833:17) : avc:  denied  { getattr } for  pid=972 comm=mdadm path=/run/initiatorname.iscsi dev="tmpfs" ino=12569 scontext=system_u:system_r:mdadm_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=0
type=SYSCALL msg=audit(22.1.2021 18:23:19.833:17) : arch=x86_64 syscall=stat success=no exit=EACCES(Operace zamítnuta) a0=0x55e616c2ce90 a1=0x7ffc88132660 a2=0x7ffc88132660 a3=0x100 items=0 ppid=1 pid=972 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=mdadm exe=/usr/sbin/mdadm subj=system_u:system_r:mdadm_t:s0 key=(null)

### DIFF
--- a/policy/modules/contrib/iscsi.if
+++ b/policy/modules/contrib/iscsi.if
@@ -208,3 +208,22 @@ interface(`iscsi_admin',`
 	files_search_tmp($1)
 	admin_pattern($1, iscsi_tmp_t)
 ')
+
+########################################
+## <summary>
+##	Read iscsi PID files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`iscsi_read_pid_files',`
+	gen_require(`
+		type iscsi_var_run_t;
+	')
+
+	allow $1 iscsi_var_run_t:file read_file_perms;
+	files_search_pids($1)
+')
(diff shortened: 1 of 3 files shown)


policy/modules/contrib/iscsi.if
```diff
@@ -208,3 +208,22 @@ interface(`iscsi_admin',`
...
+##	</summary>
+## </param>
+#
+interface(`iscsi_read_pid_files',`
+	gen_require(`
+		type iscsi_var_run_t;
+	')
+
+	files_search_pids($1)
+	read_files_pattern($1, iscsi_var_run_t, iscsi_var_run_t)
```
ghost: nit but i will mention it anyway. The applicable file context specification do not mention any directories with type iscsi_var_run_t and so this interface adds a redundant rule to the policy for each caller. It is inefficient but it also distracts when reading the policy because there is a inconsistency when comparing the file context with the rules expanded from this interface.

ie. i would use the read_file_perms permission set instead of the read_files_pattern pattern

zpytela: Thank you for the feedback. I considered using just the permission set and then chose pattern instead for the sake of consistency as almost every _read_pid_files interface uses pattern even when not necessary.

PR #819, opened 2021-07-30
Allow cockpit_ws_t get attributes of fs_t filesystems

zpytela:
This permission is required by systemctl executed in the cockpit-ws
caller domain.

Resolves: rhbz#1979182

### DIFF
--- a/policy/modules/contrib/cockpit.te
+++ b/policy/modules/contrib/cockpit.te
@@ -82,6 +82,7 @@ auth_use_nsswitch(cockpit_ws_t)
 
 corecmd_exec_bin(cockpit_ws_t)
 
+fs_getattr_xattr_fs(cockpit_ws_t)
 fs_read_efivarfs_files(cockpit_ws_t)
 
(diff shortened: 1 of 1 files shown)


WOnder93:
> This permission is required by systemctl executed in the cockpit-ws
> caller domain.

If that's the case, shouldn't it be added to `systemd_exec_systemctl()` instead?

zpytela:
> > This permission is required by systemctl executed in the cockpit-ws
> > caller domain.
> 
> If that's the case, shouldn't it be added to `systemd_exec_systemctl()` instead?

Actually not, it is a one-off usage of the systemctl command in a systemd unit. I've updated the commit message appropriately.

marusak:
Cockpit is already shipping its own policy, so this should be part of cockpit. I sent https://github.com/cockpit-project/cockpit/pull/16226

I am picking up https://github.com/fedora-selinux/selinux-policy/pull/757 again so we can prevent these situations.

PR #821, opened 2021-08-02
Allow vmtools_unconfined_t domain transition to rpm_script_t

Koncpa:
VMware allow option to installing packages outside
the VMware guest through vmtools_unconfined file.
So vmtools_unconfined_t need allow domain transition·
for installing rpm packages.Allow  process transition·
from vmtools_unconfined_t domain to rpm_script_t domain.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1872245


zpytela:
Please change WMware for VMware to make the keyword searchable. Additionally, try to make the commit message body content more comprehensible - which permissions are allowed to which subject.

zpytela:
Merged, thank you.

PR #822, opened 2021-08-02
Disable seccomp on CI containers

WOnder93:
This is a workaround for the bug described in:
https://bugzilla.redhat.com/show_bug.cgi?id=1985499


zpytela:
Merging, thank you.

PR #824, opened 2021-08-04
Allow sssd to set samba setting

Koncpa:
Sssd execute adcli and need set
samba Active Directory setting.
Allow sssd to set samba net AD
setting. Also creating needed
interfaces for this.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1855215

### DIFF
--- a/policy/modules/contrib/samba.if
+++ b/policy/modules/contrib/samba.if
@@ -178,6 +178,25 @@ interface(`samba_domtrans_unconfined_net',`
 
 ########################################
 ## <summary>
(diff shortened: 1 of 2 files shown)


zpytela:
Have you considered a transition instead of just `samba_exec()`?

Koncpa:
> Have you considered a transition instead of just `samba_exec()`?

Yes of course, but when I tried the solution with transition, it appeared a lot of more AVC's.

zpytela:
Merging, thank you.

PR #829, opened 2021-08-11
Policy fixes for udisksd

WOnder93:
I hit these missing rules when connecting removable storage media into my laptop (USB HDD and a flash disk) with the `unconfined` module disabled.

(Creating as a draft since I haven't yet verified by manual testing that all denials are really addressed.)


WOnder93:
I got a few more denials  related to dmestup and devicekit that didn't appear before. I have verified that with these commits the mentioned denials no longer appear (except for some fwupd ones, but that's for another PR...).

zpytela:
LGTM, thank you. (I won't be nitpicking e. g. about order in mount.if.)

WOnder93:
Force-pushed to replace the merge commit with a rebase (and fixed the commit message of the first commit, part of which became obsolete by another commit that is already in the rawhide branch).

zpytela:
Merging (finally), thank you.

PR #836, opened 2021-08-17
Allow rhsmcertd to create cache file in /var/cache/cloud-what

jirihnidek:
* The `rhsmcertd` uses new python package cloud-what. This
  package tries to create cache file in `/var/cache/cloud-what`
* This commit create new label `cloud_what_var_cache_t` for this
  directory and it allows `rhsmcertd` service to manage files and
  directories in this cache directory.

### DIFF
--- a/policy/modules/contrib/rhsmcertd.if
+++ b/policy/modules/contrib/rhsmcertd.if
@@ -195,6 +195,63 @@ interface(`rhsmcertd_manage_lib_dirs',`
 
 ########################################
 ## <summary>
+##	Read cloud-what cache files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cloud_what_read_cache_files',`
+	gen_require(`
+		type cloud_what_var_cache_t;
+	')
+
+	files_search_var($1)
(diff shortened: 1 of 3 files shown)


jirihnidek:
More details about this issue could be found here: https://bugzilla.redhat.com/show_bug.cgi?id=1922151#c6

policy/modules/contrib/rhsmcertd.if
```diff
@@ -193,6 +193,63 @@ interface(`rhsmcertd_manage_lib_dirs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cloud_what_read_cache_files',`
+	gen_require(`
+		type cloud_what_var_cache_t;
+	')
+
+	files_search_var_lib($1)
```
zpytela: The purpose of a line like this in the interface is to allow access to the type of the parent directory. In this case it is var_t, so `files_search_var($1)` should be used instead.
Same for the next interface.
```
# ls -lZd /var/cache /var/cache/httpd
drwxr-xr-x. 1 root   root   system_u:object_r:var_t:s0         238 Jul 16 15:02 /var/cache
drwx------. 1 apache apache system_u:object_r:httpd_cache_t:s0  10 Jun  2 14:08 /var/cache/httpd
```

zpytela (changes requested):

zpytela:
Merging, thank you.

PR #837, opened 2021-08-18
Allow D-bus communication between avahi and sosreport

Koncpa:
Avahi_t need communicate with sosreport_t via D-bus.
Allow D-bus communication between avahi_t and
sosreport_t.

rhbz#1916397


zpytela:
Merging, thank you.

PR #839, opened 2021-08-18
Allow systemd to delete fwupd var cache files

Koncpa:
Systemd can clean fwupd cache data via·
'systemctl clean --what=cache fwupd.service'.
Allow init_t domain deleting cache files·
with label fwupd_cache_t. Part of this fix is also
create new interface for this purpouses.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1759554

### DIFF
--- a/policy/modules/contrib/fwupd.if
+++ b/policy/modules/contrib/fwupd.if
@@ -78,6 +78,26 @@ interface(`fwupd_search_cache',`
 
 ########################################
 ## <summary>
+##	Allow the specified domain to delete
+##	fwupd cache.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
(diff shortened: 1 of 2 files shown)


travier:
Thanks for working on this. This takes care of the deletion but does this also includes the creation during first start?

travier:
I.e. `systemctl start fwupd.service` after that should work without denials.

Koncpa:
Hi Timothée,

I tested  `systemctl start fwupd.service` on F34 during first start and also after cleaning cache and I didn't find any AVC's denials and dir was created properly.

travier (approved):
LGTM

zpytela:
The create permission are in the policy since commit 3f0e03f65a6e7363aeb664439f993b96e1db39b2.
Merging, thanks for the patch and review.

PR #849, opened 2021-08-25
Allow braille printing (cups-brf) in selinux

zdohnal:
cups-brf gets data from getpwnam() to switch itself to the user who
issued the print job and then creates BRF dir and output files in user
home directories (/home/<username>/BRF for regular user, /root/BRF for
root).

### DIFF
--- a/policy/modules/contrib/cups.te
+++ b/policy/modules/contrib/cups.te
@@ -61,6 +61,10 @@ files_tmp_file(cupsd_lpd_tmp_t)
 type cupsd_lpd_var_run_t;
 files_pid_file(cupsd_lpd_var_run_t)
 
+type cups_brf_t, cups_domain;
+type cups_brf_exec_t;
+cups_backend(cups_brf_t, cups_brf_exec_t)
+
 type cups_pdf_t, cups_domain;
 type cups_pdf_exec_t;
 cups_backend(cups_pdf_t, cups_pdf_exec_t)
@@ -592,6 +596,30 @@ logging_send_syslog_msg(cupsd_lpd_t)
 
 optional_policy(`
 	inetd_service_domain(cupsd_lpd_t, cupsd_lpd_exec_t)
+')
+
+########################################
+#
+# BRF (Braille) local printing policy
+#
+
+# when printing as a root - to create /root/BRF/<file>
+allow cups_brf_t self:capability dac_override;
+userdom_manage_admin_dirs(cups_brf_t)
+userdom_manage_admin_files(cups_brf_t)
+
+# the backend gets user info from getpwnam() and
+# sets its uid/gid to be run under user
+auth_read_passwd(cups_brf_t)
+init_search_pid_dirs(cups_brf_t)
+
+# for writing ~/BRF/<file> for normal user
+userdom_manage_user_home_content_dirs(cups_brf_t)
+userdom_manage_user_home_content_files(cups_brf_t)
+userdom_filetrans_home_content(cups_brf_t)
+
+optional_policy(`
+	sssd_search_lib(cups_brf_t)
 ')
 
 ########################################
--- a/policy/modules/contrib/cups.fc
+++ b/policy/modules/contrib/cups.fc
@@ -30,6 +30,7 @@
 
 /usr/lib/cups/daemon/cups-lpd -- gen_context(system_u:object_r:cupsd_lpd_exec_t,s0)
 /usr/lib/cups/backend/cups-pdf -- gen_context(system_u:object_r:cups_pdf_exec_t,s0)
+/usr/lib/cups/backend/cups-brf -- gen_context(system_u:object_r:cups_brf_exec_t,s0)
 /usr/lib/cups/backend/hp.* --	gen_context(system_u:object_r:cupsd_exec_t,s0)
 
 /usr/libexec/cups-pk-helper-mechanism -- gen_context(system_u:object_r:cupsd_config_exec_t,s0)


sthibaul:
(this is very much like cups-pdf)

policy/modules/contrib/cups.te
```diff
@@ -594,6 +598,27 @@ optional_policy(`
 	inetd_service_domain(cupsd_lpd_t, cupsd_lpd_exec_t)
 ')
 
+########################################
+#
+# BRF (Braille) local printing policy
+#
+
+# when printing as a root - to create /root/BRF/<file>
+allow cups_brf_t self:capability dac_override;
```
zpytela, line 607: Can you share some details? What happens on the filesystem when printing as root?

policy/modules/contrib/cups.te
```diff
@@ -594,6 +598,27 @@ optional_policy(`
...
+userdom_manage_admin_dirs(cups_brf_t)
+userdom_manage_admin_files(cups_brf_t)
+
+# the backend gets user info from getpwnam() and
+# sets its uid/gid to be run under user
+auth_read_passwd(cups_brf_t)
+init_search_pid_dirs(cups_brf_t)
+sssd_search_lib(cups_brf_t)
```
zpytela: sssd is a module, its interfaces calls should be in an optional block.
authlogin, init, userdomain actually are modules, too, but they hardly can be removed from the system, so it's better to leave them where they are.

zpytela (changes requested):

policy/modules/contrib/cups.te
zdohnal: cups-brf backend calls getpwnam() to get info about the user who started print job and then the backend sets UID and GID to match with the issuer.
Then the backend uses umask() to 'act as the user' by forbidding permissions for 'group' and 'other' classes and creates BRF directory in user home directory (```/root``` for root user), creates a file in it, write filters' output in it and closes it.

To be honest, this is my first time writing a selinux policy, so I followed the suggestions from ```audit2allow``` command after I'd got a selinux denial, so I was updating the policy until the print job passed. Do feel free to let me know if I should approach this differently.

policy/modules/contrib/cups.te
zdohnal: @zpytela do you mean like this:

```
optional_policy(`
  sssd_search_lib(cups_brf_t)
')
```

?

zdohnal:
@zpytela ping questions in review

policy/modules/contrib/cups.te
zpytela: Yes, just move it down (L621 or so).

policy/modules/contrib/cups.te
zpytela: Please also update the commit message with the information for future reference.

zpytela:
> @zpytela ping questions in review

I am sorry, I completely missed the notifications.

policy/modules/contrib/cups.te
zdohnal: Moved down.

policy/modules/contrib/cups.te
zdohnal: The commit message was updated to:

```
   Allow braille printing in selinux

    Description of the process:

    cups-brf backend calls getpwnam() to get info about the user who started
    print job and then the backend sets its UID and GID to match with the
    issuer.
    After that the backend uses umask() to 'act as the user' by forbidding
    permissions for 'group' and 'other' classes and creates BRF directory in
    user home directory (/root for root user, /home/<username> for others),
    creates a file in the directory, write filters' output in the file
    and closes it.
```

zdohnal:
@zpytela updated, would you mind looking into the current PR?

zpytela:
The commit well matches cups-pdf so as such it is fine. However, there is one troublesome thing: with `userdom_manage_user_home_content_dirs(cups_brf_t)` we are giving to the plugin unnecessary permissions to manage user's home.

Will it be possible to change the software behaviour to use some XDG adviced directory, e. g. ~/.config/cups/brf?

zdohnal:
> Will it be possible to change the software behaviour to use some XDG adviced directory, e. g. ~/.config/cups/brf?

@zpytela  The file in the created BRF dir is a printout of a document - IMHO it doesn't make much sense to put it into a dir which is meant for configurations...

Is it possible to somehow define in selinux policy that the backend can only create and access only BRF directory in home dir, and then open/write/close files in BRF dir?

zpytela:
> > Will it be possible to change the software behaviour to use some XDG adviced directory, e. g. ~/.config/cups/brf?
> 
> @zpytela The file in the created BRF dir is a printout of a document - IMHO it doesn't make much sense to put it into a dir which is meant for configurations...

What I meant was ~/.cache, but there are other "well-known" directories, too.

> 
> Is it possible to somehow define in selinux policy that the backend can only create and access only BRF directory in home dir, and then open/write/close files in BRF dir?

Policy can give access to a particular directory, but if it does not exist, permissions to modify user's home dir are required if it is in user's home dir. That's why a dedicated data/cache/runtime directory is a better choice.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

zdohnal:
@sthibaul Hi Sam,

as cups-brf author, would you mind looking into the proposal @zpytela mentioned? Is it acceptable to change the dir?

sthibaul:
> Is it acceptable to change the dir?

Well, as much as it would be for the cups-pdf filter. The problem is that if the file is stuffed into a hidden directory, there is very little probability that the user will find it...

zdohnal:
@zpytela ,

based on @sthibaul review, the changing of dir would reduce the output visibility, so cups-brf cannot use a XDG dir for output.

zpytela:
I see. This needs to be discussed in our team, I'll let you know.

zdohnal:
Hi @zpytela ,

any news regarding the PR?

zpytela:
It is still considered unfortunate and we will appreciate using a standardized directory in the future in order not to allow arbitrary domains manage user's home. Merging though as the risk has been assessed minimal. Thank you for your contribution.

PR #857, opened 2021-09-01
Allow postfix_domain to sendto unix dgram sockets.

Koncpa:
Postfix from new version allow option to logging into file.
There is one process in postfix which handle it and other
postfix processes need communicate it through socket.
Allow postfix_domain sendto unix_dgram_socket to
postfix_master_t.

Bugzilla:https://bugzilla.redhat.com/show_bug.cgi?id=1920521


zpytela:
Please do not insert the new line between a comment and the code line it belongs to.

zpytela:
The bz link at L793 relates to the very next line L794 with `allow postfix_domain postfix_master_t:file read;`, so they should not be separated with another line.

Koncpa:
Sure, DONE. :)

zpytela:
Merging, thank you.

PR #859, opened 2021-09-02
Label /usr/sbin/virtproxyd as virtd_exec_t

Koncpa:
The virtproxyd is server side daemon component of the
libvirt virtualization management system.As it is part
of libvirt, label it as virtd_exec_t.

Bugzilla:https://bugzilla.redhat.com/show_bug.cgi?id=1854332


zpytela:
Merging, thank you.

PR #866, opened 2021-09-09
Allow firewalld the sys_module capability

zpytela:
This capability needs to be allowed when a service which requires
a kernel module loaded is added to the configuration:

firewall-cmd --add-service ftp

Resolves: rhbz#1999152

### DIFF
--- a/policy/modules/contrib/firewalld.te
+++ b/policy/modules/contrib/firewalld.te
@@ -72,6 +72,7 @@ can_exec(firewalld_t, firewalld_var_run_t)
 kernel_read_network_state(firewalld_t)
 kernel_read_system_state(firewalld_t)
 kernel_rw_net_sysctls(firewalld_t)
+kernel_load_module(firewalld_t)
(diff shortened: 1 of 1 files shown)


policy/modules/contrib/firewalld.te
```diff
@@ -69,6 +69,7 @@ manage_dirs_pattern(firewalld_t, firewalld_var_run_t, firewalld_var_run_t)
 files_pid_filetrans(firewalld_t, firewalld_var_run_t, { file dir })
 can_exec(firewalld_t, firewalld_var_run_t)
 
+kernel_load_module(firewalld_t)
```
WOnder93: Just a nit, but it would be nicer to move this line close to the `kernel_request_load_module(firewalld_t)` a few lines below.

zpytela: Right, but it is against the alphabetical-order rule. One can never satisfy everybody.

PR #874, opened 2021-09-14
Allow GUI programs execute dbus in the caller domain

zpytela:


zpytela:
The installation now passes without an issue.

policy/modules/contrib/gnome.if
```diff
@@ -1965,6 +1965,24 @@ interface(`gnome_command_domtrans_gkeyringd', `
     type_transition $1 gkeyringd_exec_t:process $2;
 ')
 
+########################################
+## <summary>
+##	Execute gnome-atspi services in the caller comain
```
freevatar, line 1970: Typo in `domain` word

zpytela: Thanks for spotting this, fixed.

PR #876, opened 2021-09-16
Fix typo in the gnome_exec_atspi() interface summary

zpytela:


zpytela:
Looks like the last change was not pushed to the previous PR.

PR #856, opened 2021-09-01
Drop cockpit module and stop ignoring standalone interface files

vmojzis:

PR #891, opened 2021-09-22
Drop glusterd and openhpid

vmojzis:
openhpid should only be dropped in rawhide. Glusterd can be dropped in F35 as well (not necessarily).
Both modules still need to be disabled in distgit.


vmojzis:
Final sync-up of glusterd module - https://github.com/gluster/glusterfs-selinux/pull/25

vmojzis:
Requires https://github.com/fedora-selinux/selinux-policy/pull/856 to work.

PR #894, opened 2021-09-23
Multiple commits which allow proper function soreport in sysadmin role

Koncpa:
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1965251

### DIFF
--- a/policy/modules/contrib/iotop.te
+++ b/policy/modules/contrib/iotop.te
@@ -20,8 +20,10 @@ role iotop_roles types iotop_t;
 #
 
 allow iotop_t self:capability net_admin;
+allow iotop_t self:netlink_generic_socket create_socket_perms;
 allow iotop_t self:netlink_route_socket r_netlink_socket_perms;
 allow iotop_t self:netlink_socket create_socket_perms;
+allow iotop_t self:unix_dgram_socket create_socket_perms;
 
 kernel_read_system_state(iotop_t)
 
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/iotop.te
```diff
@@ -20,8 +20,10 @@ role iotop_roles types iotop_t;
 #
 
 allow iotop_t self:capability net_admin;
+allow iotop_t self:netlink_generic_socket { bind create getattr setopt };
```
zpytela: Please use object permissions sets when they are available, like `create_socket_perms`.

policy/modules/contrib/iotop.te
```diff
@@ -20,8 +20,10 @@ role iotop_roles types iotop_t;
...
 allow iotop_t self:netlink_route_socket r_netlink_socket_perms;
 allow iotop_t self:netlink_socket create_socket_perms;
+allow iotop_t self:unix_dgram_socket { create getopt setopt } ;
```
zpytela: Same here.

zpytela (changes requested):
Also please change "sysadm domain" to "sysadm_t domain" in one of the commit messages.

policy/modules/contrib/iotop.te
Koncpa: DONE

policy/modules/contrib/iotop.te
Koncpa: DONE

zpytela:
Merging, thank you.

zpytela (approved):

PR #898, opened 2021-09-24
Allow management of lnk_files if similar access to regular files

rhatdan:
container-selinux has container_logreader_t which needs to be able
to read symbolic liks.

### DIFF
--- a/policy/modules/system/logging.if
+++ b/policy/modules/system/logging.if
@@ -159,6 +159,7 @@ interface(`logging_read_audit_log',`
 
 	files_search_var($1)
 	read_files_pattern($1, auditd_log_t, auditd_log_t)
+	read_lnk_files_pattern($1, auditd_log_t, auditd_log_t)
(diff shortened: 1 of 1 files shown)


policy/modules/system/logging.if
```diff
@@ -1339,6 +1341,7 @@ interface(`logging_write_generic_logs',`
 	files_search_var($1)
 	allow $1 var_log_t:dir list_dir_perms;
 	write_files_pattern($1, var_log_t, var_log_t)
+	write_lnk_files_pattern($1, var_log_t, var_log_t)
```
zpytela: Shouldn't this also be `read_lnk_file_patterns`?

rhatdan: Changed, it is arguable either way, but I doubt it is important.  Changed to read.

Please merge and generate a new release.

rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2005997

zpytela:
Merging, thank you.

PR #899, opened 2021-09-29
Allow unconfined_t transition to mozilla_plugin_t with NoNewPrivileges

zpytela:
Resolves: rhbz#2007418

### DIFF
--- a/policy/modules/contrib/mozilla.if
+++ b/policy/modules/contrib/mozilla.if
@@ -273,6 +273,24 @@ interface(`mozilla_domtrans_plugin',`
 
 ########################################
 ## <summary>
+##	Allow caller to transition to mozilla_plugin_t with NoNewPrivileges
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed access.
+## </summary>
+## </param>
+#
+interface(`mozilla_nnp_domtrans_plugin',`
+	gen_require(`
+		type mozilla_plugin_t;
+	')
+
+	allow $1 mozilla_plugin_t:process2 nnp_transition;
+')
+
+########################################
+## <summary>
 ##	Execute mozilla_plugin in the mozilla_plugin domain, and
 ##	allow the specified role the mozilla_plugin domain.
 ## </summary>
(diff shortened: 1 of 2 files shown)


WOnder93:
I don't think this is a good idea... If I understand the semantics behind the nnp_transition permission correctly, it's supposed to be allowed only when the target domain is verified to not be able to transition to another domain that has more privileges (i.e. wider permissions granted). This verification is left to the policy writer, because there is no easy way to determine this automatically.

With this patch you are 100% allowing some NNP transition that isn't really NNP. NNP transitions should be always allowed on a case-by-case basis at least at the level of target (never only source!) domains.

zpytela:
Can you point me to some resource regarding the semantics? I followed the original commit description with "systemd enables NNP not only for services whose unit files explicitly specify NoNewPrivileges=yes but also for services whose unit files specify any of the following options in combination with running without CAP_SYS_ADMIN etc." and was probably mislead.
https://patchwork.kernel.org/project/selinux/patch/20170710202553.20668-1-sds@tycho.nsa.gov/
https://www.kernel.org/doc/html/latest/userspace-api/no_new_privs.html
https://www.paul-moore.com/blog/d/2017/11/linux_v414.html
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#NoNewPrivileges=

WOnder93:
Ok, I'm not sure anymore if I understand it well... I suppose it comes down to how the policy writer decides to treat NNP transition - If they decide to treat it as a constraint only for DAC, then they could just allow it globally and be done with it. If they decide that its semantics should be preserved for SELinux, they can either unset the nnp_nosuid_transition and rely on type bounds, or set it, but be careful about which domains can be a target of an NNP/nosuid transition (possibly with justified exceptions to the strict NNP semantics). Until now, Fedora policy has been selective about where the nnp_transtions is allowed (to some extent); IMHO if we want to change the the approach, we should do it consistently, not just in this specific case.

(But then again, maybe I'm misunderstanding it completely...)

zpytela:
Ok, thanks. I understand the approach requires more attention in the future.

PR #900, opened 2021-09-30
Allow ModemManager create and use netlink route socket

zpytela:
Resolves: rhbz#2008755


oturpe (approved):
Original reporter of the Bugzilla issue here.
This update fixes the denial I get.

zpytela:
@oturpe, thanks for confirmation.

PR #907, opened 2021-10-08
authlogin: add fcontext for tcb

besser82:
tcb is an alternative password shadowing scheme used by some Linux distributions, like ALT Linux, Mandriva, OWL, and some others.

The `/etc/tcb` directory tree is used to store a single shadow file inside of a subdirectory created for every local user.

The `tcb_chkpwd` binary is meant to provide the same functionality as the `unix_chkpwd` binary.

The `tcb_convert` and `tcb_uncovert` binaries are used for conversions from a UNIX shadow file to the tcb password shadowing scheme and vice-versa.

This is already part of the SELinux reference-policy:
https://github.com/SELinuxProject/refpolicy/commit/bc88a1ca4b37df37c3654fc9e5368d7d96b11548


policy/modules/system/authlogin.fc
```diff
@@ -13,6 +13,7 @@ HOME_DIR/\.google_authenticator~		gen_context(system_u:object_r:auth_home_t,s0)
 /etc/gshadow.*		--	gen_context(system_u:object_r:shadow_t,s0)
 /etc/nshadow.*		--	gen_context(system_u:object_r:shadow_t,s0)
 /etc/shadow.*		--	gen_context(system_u:object_r:shadow_t,s0)
+/etc/tcb(/.*)?		--	gen_context(system_u:object_r:shadow_t,s0)
```
zpytela: The `--` specifier means the entry applies to a plain file only.

besser82: So removing it would be the solution in this case?

WOnder93: But I don't think we want to have directories labeled as `shadow_t` - it would likely lead to new denials, as there are no rules in the policy that would allow accessing `shadow_t` directories (except as part of the `file_type` attribute). So the current "label all regular files under /etc/tcb as `system_u:object_r:shadow_t:s0`" is exactly what we want here, IMO.

zpytela: Right. My point was to remove the '()' regexp part, but this is also reasonable as there are directories and symlinks in /etc/tcb, not just files. Thanks for pointing this out.

@besser82, is it expected a plain file which is not shadow-type appears in /etc/tcb? I understand there should only be `/etc/tcb/username/shadow` or `/etc/tcb/other/path/shadow` , and `username` can be a directory or a symlink somewhere else inside the tree. The regexp can be more strict then, like `/etc/tcb/[^/]+/.+` if it does not make it too hard.

besser82: @zpytela Exactly.  There are no other files to be expected in there execpt for the `shadow` file placed inside a distinct sub-directory for each system user; during changes of the password one may expect a temporary file or a lock file, but labeling those `shadow_t` shouldn't be an issue as the corresponding process has at least `updpwd_t` capabilities anyways.

I've updated the commit to reflect your suggested regex.

zpytela: So the regexp can even end with shadow, not .+, or is it too much already?

besser82: That's perfectly fine either.  I've changed that in the rebased commit.

zpytela: Well this regexp `/etc/tcb/[^/]+/shadow.*` now supports only one directory level.

besser82: Using `/etc/tcb/(.*/)?shadow.*` should be fine then, I guess?

zpytela:
Actually `/etc/tcb/.+/shadow.*`, but it's not a big deal.

besser82:
> Actually `/etc/tcb/.+/shadow.*`, but it's not a big deal.

Fixed.

zpytela:
Merging, thank you and sorry for nitpicking.

besser82:
np.  You're welcome!

PR #910, opened 2021-10-12
Allow keepalived setattr on scripts

Koncpa:


zpytela:
It is still not clear why would keepalived call utimensat() on notify scripts and the referred bz was closed, hence closing this pr, too.

PR #912, opened 2021-10-13
Allow unconfined_t transition to mozilla_plugin_t with NoNewPrivileges

zpytela:
The mozilla_nnp_domtrans_plugin() interface was added.

Resolves: rhbz#2007418


zpytela:
Build is expected to fail for f34 build.

PR #921, opened 2021-10-19
Allow fedora-third-party run flatpak post-install actions

zpytela:
In gnome-initial-setup, fedora-third-party executes flatpak which needs:
- work with temporary files
- execute gpg/gpgsm
- network resolution

Resolves: rhbz#2001837


zpytela:
This is meant a temporary workaround until flatpak is confined.

PR #929, opened 2021-10-28
filesystem: add fs_use_trans for ramfs

cgzones:
Enable extended attributes for inodes on ramfs filesystems, similar to
tmpfs filesystems.

For example systemd uses ramfs for service credentials[1], and xattr
support is needed for per service based labeling[2].

[1]: https://www.freedesktop.org/software/systemd/man/systemd-creds.html
[2]: https://github.com/systemd/systemd/pull/21158

Signed-off-by: Christian Göttsche <cgzones@googlemail.com>


zpytela:
@cgzones Could you please add the justification and the links to the commit message as well?

cgzones:
> Could you please add the justification and the links to the commit message as well?

Aren't they already?

zpytela:
Merging, thank you. You are right, I must have overlooked that.

PR #940, opened 2021-11-12
Report warning on duplicate definition of interface

vmojzis:
Duplicate definition of interface or template should only be a warning,
since it does not block the compilation.
The warning should also specify which file contains the original
definition.

Old:
 ipa.if:284: Error: duplicate definition of
  ipa_cert_filetrans_named_content(). Original definition on 284.
New:
 ipa.if:284: Warning: duplicate definition of
  ipa_cert_filetrans_named_content(). Original definition in
  /usr/share/selinux/devel/include/contrib/ipa.if:284.

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>

### DIFF
--- a/policy/support/loadable_module.spt
+++ b/policy/support/loadable_module.spt
@@ -60,7 +60,7 @@ define(`policy_m4_comment',`
 # template(name,rules)
 #
 define(`template',` dnl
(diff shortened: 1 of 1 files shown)


zpytela:
Merging, thank you.

PR #941, opened 2021-11-12
Sync the policy_capabilities file with refpolicy and enable genfs_seclabel_symlinks

WOnder93:
See commit messages for more details.


ghost (commented):
I suspect that the /proc/net symlink label could potentially cause slight disruption. It probably changed from proc_t to proc_net_t with this patch

WOnder93:
Indeed it is now `proc_net_t`, but `kernel_read_network_state()` already defines the necessary `lnk_file` permissions and there is also `kernel_read_network_state_symlinks()` used with a couple domains, so this shouldn't cause much, if any, disruption.

ghost:
Ondrej Mosnáček ***@***.***> writes:

> Indeed it is now proc_net_t, but kernel_read_network_state() already defines the necessary lnk_file
> permissions and there is also kernel_read_network_state_symlinks() used with a couple domains, so this
> shouldn't cause much, if any, disruption.


Alright, nice.

Strange though that the interfaces facilitated reading proc_net_t
symlinks as I would imagine that without this policy capability there
would never be any, since proc_t is not a seclabel fs.

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

zpytela:
Merging, thank you.

PR #959, opened 2021-11-30
Ensure that `/run/systemd/*` are properly labeled

DemiMarie:
`/run/systemd/generator.{early,late}` were not covered by the type_transition rules.


zpytela:
You are right, did you see them somewhere mislabeled though?

DemiMarie:
I did, though that was systemd/systemd#18076 (now fixed).  Still, if those rules are going to be present at all they should be consistent.

zpytela:
Merging, thank you.

PR #962, opened 2021-12-03
Make postfix_postdrop_exec_t type an MTA executable file

zpytela:
Previously, it was postfix_postdrop_t which was assigned to the
mta_exec_type attribute which is incorrect. Since this commit it rather
is postfix_postdrop_exec_t instead of postfix_postdrop_t.
This change makes it aligned with other mta types like sendmail_exec_t,
while postfix_postdrop_t remains in the mta_user_agent attribute.

Resolves: rhbz#2004843


zpytela:
Now there are 2 interfaces which cannot be applied to the same type: postfix_domtrans_postdrop() and mta_send_mail(). Used only for asterix_t and spamc_t.

PR #966, opened 2021-12-06
Allow httpd named transitions in tmpfs directories

zpytela:
A part of this commit are also rules to manage pkcs_slotd_tmpfs_t type files.

Addresses AVC denials like these:

type=PROCTITLE msg=audit(12/02/2021 15:49:36.552:674) : proctitle=/usr/sbin/httpd -DFOREGROUND
type=PATH msg=audit(12/02/2021 15:49:36.552:674) : item=0 name=/dev/shm/var.lib.opencryptoki.swtok inode=64710 dev=00:16 mode=file,660 ouid=apache ogid=pkcs11 rdev=00:00 obj=unconfined_u:object_r:pkcs_slotd_tmpfs_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(12/02/2021 15:49:36.552:674) : cwd=/
type=SYSCALL msg=audit(12/02/2021 15:49:36.552:674) : arch=x86_64 syscall=openat success=yes exit=19 a0=0xffffff9c a1=0x7ffeb38e2e60 a2=O_RDWR|O_NOFOLLOW|O_CLOEXEC a3=0x0 items=1 ppid=1 pid=21325 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=httpd exe=/usr/sbin/httpd subj=system_u:system_r:httpd_t:s0 key=(null)
type=AVC msg=audit(12/02/2021 15:49:36.552:674) : avc:  denied  { open } for  pid=21325 comm=httpd path=/dev/shm/var.lib.opencryptoki.swtok dev="tmpfs" ino=64710 scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:pkcs_slotd_tmpfs_t:s0 tclass=file permissive=1
type=AVC msg=audit(12/02/2021 15:49:36.552:674) : avc:  denied  { read write } for  pid=21325 comm=httpd name=var.lib.opencryptoki.swtok dev="tmpfs" ino=64710 scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:pkcs_slotd_tmpfs_t:s0 tclass=file permissive=1

Resolves: rhbz#2028637

### DIFF
--- a/policy/modules/contrib/apache.te
+++ b/policy/modules/contrib/apache.te
@@ -886,6 +886,10 @@ tunable_policy(`httpd_use_fusefs',`
 	fs_manage_fusefs_symlinks(httpd_t)
 ')
 
+tunable_policy(`httpd_use_opencryptoki',`
+	allow httpd_t self:capability fowner;
+')
+
 tunable_policy(`httpd_setrlimit',`
 	allow httpd_t self:process setrlimit;
 	allow httpd_t self:capability sys_resource;
@@ -930,6 +934,9 @@ optional_policy(`
 ')
 
 optional_policy(`
+	# type transitions with a filename not allowed inside conditionals
+	pkcs_tmpfs_named_filetrans(httpd_t)
+
     tunable_policy(`httpd_use_opencryptoki',`
         pkcs_use_opencryptoki(httpd_t)
     ')
--- a/policy/modules/contrib/pkcs.if
+++ b/policy/modules/contrib/pkcs.if
@@ -132,12 +132,11 @@ interface(`pkcs_tmpfs_named_filetrans',`
 		type pkcs_slotd_tmpfs_t;
 	')
 
-	allow $1 pkcs_slotd_tmpfs_t:file map;
-
-	manage_files_pattern($1, pkcs_slotd_tmpfs_t, pkcs_slotd_tmpfs_t)
 	fs_tmpfs_filetrans($1, pkcs_slotd_tmpfs_t, file, "var.lib.opencryptoki.ccatok")
 	fs_tmpfs_filetrans($1, pkcs_slotd_tmpfs_t, file, "var.lib.opencryptoki.ep11tok")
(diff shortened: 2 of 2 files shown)


DemiMarie:
httpd_t should not be allowed to manage PKCS#11 stuff by default.  PKCS#11 stuff is trusted significantly more.

What would work instead is splitting up the policy so that the allow rules were under a conditional.

zpytela:
> httpd_t should not be allowed to manage PKCS#11 stuff by default. PKCS#11 stuff is trusted significantly more.
> 
> What would work instead is splitting up the policy so that the allow rules were under a conditional.

I agree. The question now is: can httpd be the first one to create the tokens? If yes, it needs to be allowed the named file transitions, and to back that also manage_files_pattern(), at least a part of it. Named file transitions cannot be inside tunable_policy().

DemiMarie:
> I agree. The question now is: can httpd be the first one to create the tokens? If yes, it needs to be allowed the named file transitions, and to back that also manage_files_pattern(), at least a part of it. Named file transitions cannot be inside tunable_policy().

What about including the `type_transition` rules, but *not* the corresponding `allow:file create` rules, so that `httpd_t` gets errors because it cannot create the file with the correct context?  Users who need this could change the httpd systemd service file to specify `PrivateTmp=yes`, which is a good idea anyway.

DemiMarie:
@zpytela should httpd just be using p11-kit?  Most PKCS#11 modules require permissions (such as the ability to access devices or credentials for network HSMs) that httpd should not have.

zpytela:
httpd needs to work with files like /dev/shm/var.lib.opencryptoki.swtok
I am currently rewriting the previous commit in order to give as low number of permissions as possible.

zpytela:
I believe the current state is in a better shape.

PR #975, opened 2021-12-15
New devices

zpytela:


ghost (commented):
"The same type is also used for /dev/nvme-subsys[0-9]+ block devices."

Either the description is wrong or the filecon is wrong because:

The description say's is a block device
The filecon say's is a char device

zpytela:
They are char devices, thank you for spotting this. I wanted to pinpoint they are char devices although they are used for disks.

PR #976, opened 2021-12-20
Sudo and sysadm commits

zpytela:


ghost (commented):
I read the bugzilla but this does not make sense

1. this "solution" would only address `sudo passwd` and the reporter states that he would like this to work for any command.
2. you shouldnt have to run sudo from a sysadm shell because the sysadm_t domain is meant for root, so if youre sysadm_t then you are already root (no need to run sudo then)
3. but even if it would make sense somehow then i guess you could either specify `-r sysadm_r -t sysadm_t` on the sudo command like prompting it to run `sesh` or add an entry to `sudoers` "root ALL=(ALL) ROLE=sysadm_r TYPE=sysadm_t ALL"

But again, sysadm_t should only ever be used with root, so if you are in the sysadm_t domain and you aren't root already then youre doing it wrong.

zpytela:
> I read the bugzilla but this does not make sense
> 
>     1. this "solution" would only address `sudo passwd` and the reporter states that he would like this to work for any command.

If issues like this are to be resolved in selinux-policy, they need to be addressed one by one.

>     2. you shouldnt have to run sudo from a sysadm shell because the sysadm_t domain is meant for root, so if youre sysadm_t then you are already root (no need to run sudo then)

It seems to be a common case sysadm_u is assigned to non-root users. In that case they need to execute commands through sudo, which then, except a few explicit rules, end up in `sysadm_sudo_t.`

>     3. but even if it would make sense somehow then i guess you could either specify `-r sysadm_r -t sysadm_t` on the sudo command like prompting it to run `sesh` or add an entry to `sudoers` "root ALL=(ALL) ROLE=sysadm_r TYPE=sysadm_t ALL"

These two approaches are proved to work, additionally running the command through a shell works, too. Sudo without explicit role goes a different code path and request to go like if there was "-r DEFAULTROLE" on the command line was not accepted in sudo upstream.

> But again, sysadm_t should only ever be used with root, so if you are in the sysadm_t domain and you aren't root already then youre doing it wrong.

The only other approach to this I can see is document it properly, but I am afraid we've got too far and cannot change it easily. Looking for any documentation mentioning strict forbidding, I haven't succeeded, there are only positive examples. I quite agree and I can even think of not stopping here and recommend (not enforce) not to use sysadm_u as a login user, but that is definitely unfeasible. Having said that, at the moment I can't see any other direct solution to the problem "non root sysadm_r user executes sudo command".

ghost:
Okay, looks like you know what you are doing,  This is not an issue that should be addressed in the sudo code though. Sysadm_t was never designed to be used with users other than root. The fact that people decide to use sysadm_t with non-root does not change that.

zpytela:
> Okay, looks like you know what you are doing, This is not an issue that should be addressed in the sudo code though. Sysadm_t was never designed to be used with users other than root. The fact that people decide to use sysadm_t with non-root does not change that.

I'd be grateful to any resource to this statement. Does not seem to be in common ones like selinux-notebook, danwalsh.livejournal, etc. On the contrary, SELinux administration (2013)  gives `semanage login -a -s sysadm_u "%admins"` as an example.

And surely the fact I cannot find another way out does not mean it does not exist.

ghost:
Yes It might not have been documented clearly but if you read "between the lines" then it might become clear. Take this table for example:

http://www.selinuxproject.org/page/RefpolicyBasicRoleCreation

It states that staff_r is the "Administrator's unprivileged user role". What this means is that this is the administrators role to be used for "non-root" administrators. The other "administrator" roles do not mention this "unprivileged" aspect (which kind of implies that those are "privileged administrator roles").

But the policy itself kind of tells the story and the fact that for example this does not currently work basically kind of indicates that this was never supposed to work in the first place.

You could ask @pebenito and maybe he can confirm or deny my take on this matter.

Regardless though, I do not mind that you add this support. I am just glad that the proposal to try to support this in the sudo code was rejected.

zpytela:
One commit added and for the passwd one the commit message rewritten so it is clearer why and could possibly be reverted later.

ghost:

> One commit added and for the passwd one the commit message rewritten so it is clearer why and could
> possibly be reverted later.

I guess systemctl_exec_t is already an entrypoint to (atleast) sysadm_t?
Because that would probably (atleast) be needed to get this working. But
to make this work for all of $3 I guess you'd need a
systemctl_entrypoint($3) and that then implies that anyone who is
allowed to run sudo with a domain transition is also allowed to execute
systemctl (if systemctl module is enabled).

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

Koncpa (approved):

zpytela:
For 4 login domains entrypoint was added in ec151331fea8e6a4dc65664d6118c0aa478c654a (different problem), but I see only 2 of them applied. Seems reasonable add it to the current commit anyway.

ghost:
That makes sense because unconfined_t and user_t do not call that sudo role template.

Anyway, I suspect that this might still not be enough and that you might need to at least add a `allow $1_sudo_t $3:process noatsecure;' (if that is not allowed already) to make this work consistently.

zpytela:
> That makes sense because unconfined_t and user_t do not call that sudo role template.

Yes and no, there are explicit rules for user_t and staff_t outside sudo template because there was another request for that.

> Anyway, I suspect that this might still not be enough and that you might need to at least add a `allow $1_sudo_t $3:process noatsecure;' (if that is not allowed already) to make this work consistently.

It seems not to be present, but the use cases work now.

ghost:
Okay, I think the "noatsecure" is needed for the systemctl functionality that relies on dbus (and the dbus socket path in its environment). So It might work for the functionality that does not require DBUS.

I could be wrong though.

PR #1014, opened 2022-01-19
Revert "Label /etc/cockpit/ws-certs.d with cert_t"

zpytela:
This reverts commit 0ac09c6d3b987d034e759936b92565831405e395.

Unfortunately duplicate file context entries cannot co-exist
even in modules installed with a different priority.

### DIFF
--- a/policy/modules/system/miscfiles.fc
+++ b/policy/modules/system/miscfiles.fc
@@ -9,7 +9,6 @@ ifdef(`distro_gentoo',`
 # /etc
 #
 /etc/avahi/etc/localtime --	gen_context(system_u:object_r:locale_t,s0)
-/etc/cockpit/ws-certs\.d(/.*)?	gen_context(system_u:object_r:cert_t,s0)
 /etc/docker/certs\.d(/.*)?          gen_context(system_u:object_r:cert_t,s0)
 /etc/httpd/alias(/.*)?	        gen_context(system_u:object_r:cert_t,s0)
 /etc/localtime		-l	gen_context(system_u:object_r:locale_t,s0)


martinpitt:
@zpytela : Is there still something which either you or me need to do about the original issue? #856 has been stuck for a while, but is there anything wrt. ws-certs.d labels which is still broken right now?

Thank you!

zpytela:
> @zpytela : Is there still something which either you or me need to do about the original issue? #856 has been stuck for a while, but is there anything wrt. ws-certs.d labels which is still broken right now?
> 
> Thank you!

No, a new policy build is in CI pipeline and should be ready in a few hours.

martinpitt:
I didn't mean that revert -- that's obviously the right thing to do for unbreaking the world. I meant for the original commit 0ac09c6d3b987d034e759936b92565831405e395 -- I suppose that was done for a reason, to fix https://bugzilla.redhat.com/show_bug.cgi?id=1907473 . My question is, is that actually still necessary? We ship our custom cockpit policy in RHEL 8 now, so I would have thought selinux-policy wouldn't have to care about this any more?

zpytela:
> I didn't mean that revert -- that's obviously the right thing to do for unbreaking the world. I meant for the original commit [0ac09c6](https://github.com/fedora-selinux/selinux-policy/commit/0ac09c6d3b987d034e759936b92565831405e395) -- I suppose that was done for a reason, to fix https://bugzilla.redhat.com/show_bug.cgi?id=1907473 . My question is, is that actually still necessary? We ship our custom cockpit policy in RHEL 8 now, so I would have thought selinux-policy wouldn't have to care about this any more?

Absolutely, it was one of today's findings that cockpit's policy it is not in cockpit-selinux which had made me believe cockpit does not ship its own policy in RHEL 8.

PR #1018, opened 2022-01-21
Label /var/lib/plocate with locate_var_lib_t

zpytela:
plocate is a new locate implementation to replace mlocate in the near
future. It uses /var/lib/plocate as the data directory so the directory
needs to have the correct label.

### DIFF
--- a/policy/modules/contrib/slocate.fc
+++ b/policy/modules/contrib/slocate.fc
@@ -3,5 +3,6 @@
 /usr/bin/updatedb.*	--	gen_context(system_u:object_r:locate_exec_t,s0)
 
 /var/lib/[sm]locate(/.*)?	gen_context(system_u:object_r:locate_var_lib_t,s0)
+/var/lib/plocate(/.*)?		gen_context(system_u:object_r:locate_var_lib_t,s0)
 
 /var/run/mlocate\.daily\.lock	--	gen_context(system_u:object_r:locate_var_run_t,s0)


policy/modules/contrib/slocate.fc
```diff
@@ -3,5 +3,6 @@
 /usr/bin/updatedb.*	--	gen_context(system_u:object_r:locate_exec_t,s0)
 
 /var/lib/[sm]locate(/.*)?	gen_context(system_u:object_r:locate_var_lib_t,s0)
+/var/lib/plocate(/.*)?		gen_context(system_u:object_r:locate_var_lib_t,s0)
```
keszybz, line 6: Or maybe do `/var/lib/[smp]locate(/.*)?	gen_context(system_u:object_r:locate_var_lib_t,s0)` instead? That'd be more consistent.

zpytela: You're right. However, fixfiles converts all regexps to .* wildcard which means /var/lib would be relabeled. Been there already.

keszybz: I don't understand. If line 5 is ok, then just adding one more character there shouldn't change anything.

zpytela: On selinux-policy updates, rpm script identifies changed files and runs fixfiles for them. If any filename contains a regexp, it is converted to "*" without further distinguishing and passed to restorecon.

https://github.com/SELinuxProject/selinux/blob/master/policycoreutils/scripts/fixfiles#L170

keszybz: Wow.

zpytela: I dare to read that as approved.

keszybz: Correct :)

PR #1021, opened 2022-01-22
Add missing equivalence rules

DemiMarie:
I found this problem while adding SELinux support to Qubes OS and while
looking at upstream refpolicy.

### DIFF
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -9,14 +9,24 @@
 # example, but aliasing.
 # 
 /etc/init.d /etc/rc.d/init.d
+/usr/etc /etc
+/etc/systemd /usr/lib/systemd
 /lib32 /lib
 /lib64 /lib
 /run /var/run
 /run/lock /var/lock
 /usr/lib32 /usr/lib
 /usr/lib64 /usr/lib
+/usr/local/etc /etc
 /usr/local/lib32 /usr/lib
 /usr/local/lib64 /usr/lib
 /usr/local/lib /usr/lib
+/usr/local/include /usr/include
+/usr/local/bin /usr/bin
+/usr/local/src /usr/src
+/usr/local/sbin /usr/sbin
+/usr/local/share /usr/share
+/usr/local/libexec /usr/libexec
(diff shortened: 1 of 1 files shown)


config/file_contexts.subs_dist
```diff
@@ -9,6 +9,8 @@
 # example, but aliasing.
 # 
 /etc/init.d /etc/rc.d/init.d
+/etc /usr/local/etc
```
bachradsusi: This looks wrong.

WOnder93:
The contents of this file are not currently used in the Fedora package. In Fedora it is replaced by a downstream version:
https://src.fedoraproject.org/rpms/selinux-policy/blob/rawhide/f/file_contexts.subs_dist

I have proposed to shift most of the downstream files to this repo in #686, but it's currently stalling...

config/file_contexts.subs_dist
DemiMarie: This one comes from Qubes OS, where `/usr/local/etc` is used to store per-VM configuration that overrides configuration in `/etc`.  At a minimum, I think that `/usr/local/etc` needs to have better labeling than `usr_t`.  Right now, it is very difficult for downstream packages to add their own substitutions, which is annoying.

ghost:
Demi Marie Obenour ***@***.***> writes:

> @DemiMarie commented on this pull request.
>
> ------------------------------------------------------------------------------------------------------
>
> In config/file_contexts.subs_dist:
>
>> @@ -9,6 +9,8 @@
>  # example, but aliasing.
>  # 
>  /etc/init.d /etc/rc.d/init.d
> +/etc /usr/local/etc
>
> This one comes from Qubes OS, where /usr/local/etc is used to store per-VM configuration that
> overrides configuration in /etc. At a minimum, I think that /usr/local/etc needs to have better
> labeling than usr_t. Right now, it is very difficult for downstream packages to add their own
> substitutions, which is annoying.

Yes but its the other way around:

/usr/local/etc /etc

So the entry is just technically wrong (this idea is good)

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

PR #1023, opened 2022-01-25
Allow domain transition to sssd_t and fix badly indented used interfaces

Koncpa:
**Commit 1:**

Allow domain transition to sssd_t

When installing some rpm packages, new users or 
groups are added to the system using
the groupadd and useradd tools. Then the sss_cache 
file with the bin_t label is run and on this file 
groupadd and useradd want to setgid and this 
trigger SELinux denials. Label the sss_cache binary 
as sssd_exec_t and enabling the transition from 
groupadd_t and useradd_t to sssd_t. Sssd policy 
allowed setgid on this binary.
    
Bugzilla:https://bugzilla.redhat.com/show_bug.cgi?id=2022690

**Commit 2:**

Fix badly indented used interfaces

In policy is few badly indented used
iterfaces in optional blocks. Correction
of indentation of interface names.

### DIFF
--- a/policy/modules/admin/usermanage.te
+++ b/policy/modules/admin/usermanage.te
@@ -296,10 +296,10 @@ optional_policy(`
 
 optional_policy(`
 	sssd_domtrans(groupadd_t)
-    sssd_manage_lib_files(groupadd_t)
-    sssd_manage_public_files(groupadd_t)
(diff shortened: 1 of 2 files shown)


zpytela:
Merging, thank you.

PR #1028, opened 2022-01-26
Allow systemd services watch dbusd pid directory and its parents

zpytela:
The following services were allowed to watch /run/dbus and all its
parents:
- systemd-resolved
- systemd-networkd
- systemd-timesyncd and systemd-timedated

These permissions are required when the services start earlier than
dbus-broker manages to establish the socket communication, e. g. after

  $ systemctl disable NetworkManager && systemctl enable systemd-networkd

The dbus_watch_pid_dir_path() interface was added. The redundant
dbus_watch_pid_dirs() calls were removed.

Resolves: rhbz#2031668

### DIFF
--- a/policy/modules/contrib/dbus.if
+++ b/policy/modules/contrib/dbus.if
@@ -606,6 +606,27 @@ interface(`dbus_watch_pid_dirs',`
 
 ########################################
 ## <summary>
+##	Watch system dbusd pid directory and all its parents
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dbus_watch_pid_dir_path',`
+	gen_require(`
+		type system_dbusd_var_run_t;
+	')
+
+	files_watch_root_dirs($1)
+	files_watch_var_run_dirs($1)
+	files_search_pids($1)
+	allow $1 system_dbusd_var_run_t:dir watch_dir_perms;
+')
+
+########################################
+## <summary>
 ##	Read and write system dbus tmp socket files.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -458,6 +458,7 @@ getty_start_services(systemd_machined_t)
 optional_policy(`
 	dbus_connect_system_bus(systemd_machined_t)
 	dbus_system_bus_client(systemd_machined_t)
+	dbus_watch_pid_dir_path(systemd_machined_t)
 ')
 
 optional_policy(`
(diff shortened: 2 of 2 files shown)


ghost:
Also applies to systemd-oomd AFAIK

zpytela:
@keszybz are there more systemd services which can start so early that they need to start with watching fs root?

ghost:
I think it probably actually applies to any daemon that owns a name on the bus. Strictly speaking DBUS is optional.

keszybz:
This functionality is implemented in a generic fashion in sd-bus. So any service can start using this feature if it is started when dbus is not running. As to which service could be started like this: I would just allow this everywhere, because it makes things more robust if dbus-broker is delayed or not started for any reason, and I don't see any downsides to allowing it.

zpytela:
It would be quite easy to allow it for the `systemd_domain` attribute, but I don't have a good feeling about it. Not for a particular reason.

systemd_bootchart_t systemd_coredump_t systemd_gpt_generator_t systemd_hostnamed_t systemd_hwdb_t systemd_importd_t systemd_initctl_t systemd_journal_upload_t systemd_localed_t systemd_logger_t systemd_logind_t systemd_machined_t systemd_modules_load_t systemd_networkd_t systemd_notify_t systemd_passwd_agent_t systemd_resolved_t systemd_rfkill_t systemd_sleep_t systemd_sysctl_t systemd_timedated_t systemd_tmpfiles_t systemd_userdbd_t

ghost:
fwiw In my personal policy i limited this access only to systemd services that own a name on the system bus

```
root@brutus:~# seinfo -xadbus.monitoravailability.typeattr

Type Attributes: 1
   attribute dbus.monitoravailability.typeattr;
        systemd.hostname.subj
        systemd.import.subj
        systemd.locale.subj
        systemd.login.subj
        systemd.machine.subj
        systemd.network.subj
        systemd.oom.subj
        systemd.portable.subj
        systemd.resolve.subj
        systemd.timedate.subj
        systemd.timesync.subj
root@brutus:~# 
```

```
root@brutus:~# sesearch -A -s dbus.monitoravailability.typeattr -ds
allow dbus.monitoravailability.typeattr dbus.run.file:dir { getattr ioctl lock read search watch };
allow dbus.monitoravailability.typeattr dbus.run.file:sock_file { getattr ioctl lock read watch };
allow dbus.monitoravailability.typeattr root.file:dir { getattr ioctl lock read search watch };
allow dbus.monitoravailability.typeattr run.file:dir { getattr ioctl lock read search watch };
root@brutus:~#
```
its not pretty but ce'st la vie

zpytela:
I took all information into account and decided to add 2 more services, systemd-hostnamed and systemd-machined, where it matches this commit description. In Fedora we do not confine homed, oomd, portabled yet.
I see there might be other good reasons to have it allowed for all the others.

@doverride note systemd-homed is also the case.

ghost:
Thanks, yes. Debian does not ship systemd-home

PR #1030, opened 2022-01-28
Allow confined users to use kinit,klist and etc.

Koncpa:
Confined users need to use kerberos authentication
 for NFS. Allow confined users to write gssd keys and
use tools like kinit, klist and kdestroy. Allow for
confined users like staff_t, admin types and user_t.
For admin types allow read kerberos keytab file via
adding interfaces kerberos_read_keytab() for
confined_admindomain attribute.
    
 Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2026598

### DIFF
--- a/policy/modules/system/userdomain.te
+++ b/policy/modules/system/userdomain.te
@@ -578,6 +578,14 @@ tunable_policy(`selinuxuser_execstack',`
 optional_policy(`
 	fs_list_cgroup_dirs(confined_admindomain)
 ')
+
+optional_policy(`
+	kerberos_read_keytab(confined_admindomain)
+')
+
+optional_policy(`
+	rpc_rw_gssd_keys(confined_admindomain)
+')
 	
 optional_policy(`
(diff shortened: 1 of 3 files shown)


policy/modules/system/userdomain.te
```diff
@@ -578,6 +578,14 @@ tunable_policy(`selinuxuser_execstack',`
 optional_policy(`
 	fs_list_cgroup_dirs(confined_admindomain)
 ')
+
+optional_policy(`
+        kerberos_read_keytab(confined_admindomain)
```
zpytela: This looks like space-formatted.

zpytela (changes requested):
Please also fix the confide_admindomain typo in the commit message.

Koncpa:
> Please also fix the confide_admindomain typo in the commit message.

Sure, now It should be ok. :)

zpytela:
Merging now, thank you.

PR #1047, opened 2022-02-04
Fix koji repo URL pattern

WOnder93:
For non-rawhide releases, 'f$releasever-build' tag name needs to be used
instead of '$releasever'.


zpytela:
Mergin, thank you.

PR #1053, opened 2022-02-07
rabbitmq: add 15671 to rabbitmq_port_t

ktdreyer:
`rabbitmq-server-3.8.3-1.el8` fails to start unless it can bind to tcp sockets on `unreserved_port_t`.


ktdreyer:
I've published CentOS 8 RPMs with this change at https://fedorapeople.org/~ktdreyer/selinux-rabbitmq/

I've verified that this change allows me to start rabbitmq-server from the CentOS Messaging SIG's `centos-release-rabbitmq-38`

zpytela:
@ktdreyer Thank you for the contribution. Do you know which ports are actually needed?

ktdreyer:
Thanks, I should have looked more carefully at the exact port. I researched this more, and TL;DR, this works for me:
```
semanage port -a -t rabbitmq_port_t -p tcp 15671
```

The problem is with RabbitMQ's management web UI port. When I enable the management web UI, I get this AVC denial, and when I comment it out, rabbitmq-server can start properly.

Here are the relevant settings in `/etc/rabbitmq/rabbitmq.conf`:
```
management.ssl.port       = 15671
management.ssl.certfile   = /etc/pki/koji/kojidev.ktdreyer.com.chain.crt
management.ssl.keyfile    = /etc/pki/koji/kojidev.ktdreyer.com.key
management.ssl.cacertfile = /etc/pki/koji/koji-ca.crt
```

From https://www.rabbitmq.com/management.html#single-listener-port:
* tcp 15672 is plaintext HTTP
* tcp 15671 is HTTPS

In 8e0c4d3faa8ad8c572be3524b67da8cc8ea0f512, we added port 15672 as `amqp_port_t`. [rhbz#1107873](https://bugzilla.redhat.com/show_bug.cgi?id=1107873) talks about qpid, not rabbitmq, but it just so happens that RabbitMQ uses TCP 15672 as well, so that works nicely.

In conclusion, I've rewritten this PR to add tcp 15671 as `rabbitmq_port_t` instead of using `corenet_tcp_bind_all_unreserved_ports`.

ktdreyer:
I've updated my RPMs at https://fedorapeople.org/~ktdreyer/selinux-rabbitmq/ to have this latest `rabbitmq_port_t` patch. I tested these RPMs in a CentOS 8 VM, and I can bring up rabbitmq-server with `management.ssl.port = 15671` now.

kajinamit (approved):

ktdreyer:
Would you mind merging this change?

PR #1056, opened 2022-02-08
New policy for insight-client

5umm3r15:


policy/modules/contrib/insights_client.fc
```diff
@@ -0,0 +1,9 @@
+/etc/insights-client/			-d	gen_context(system_u:object_r:insights_client_etc_t,s0)
+/etc/insights-client/[^/]*		--	gen_context(system_u:object_r:insights_client_etc_t,s0)
```
zpytela: Possibly '+' can be used instead of '*'

policy/modules/contrib/insights_client.fc
```diff
@@ -0,0 +1,9 @@
...
+
+/usr/bin/insights-client		--	gen_context(system_u:object_r:insights_client_exec_t,s0)
+
+/var/lib/insights/newest\.egg(/.*)?	--	gen_context(system_u:object_r:insights_client_var_lib_t,s0)
```
zpytela: This won't work - `--` matches plain files only.

policy/modules/contrib/insights_client.fc
```diff
@@ -0,0 +1,9 @@
...
+/var/lib/insights/newest\.egg.asc	--	gen_context(system_u:object_r:insights_client_var_lib_t,s0)
+
+/var/log/insights-client(/.*)?		--	gen_context(system_u:object_r:insights_client_var_log_t,s0)
```
zpytela: Same here.

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+files_pid_filetrans(insights_client_t, insights_client_var_run_t, { dir file })
+
+auth_read_passwd(insights_client_t)
+corecmd_exec_bin(insights_client_t)
+corenet_tcp_bind_generic_node(insights_client_t)
+corenet_tcp_connect_http_port(insights_client_t)
+ 
+domain_use_interactive_fds(insights_client_t)
+
+files_read_etc_files(insights_client_t)
```
zpytela: Why is this needed?

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+gpg_exec(insights_client_t)
+gpg_manage_home_content(insights_client_t)
```
zpytela: gpg interfaces should be in an optional block

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+
+libs_exec_ldconfig(insights_client_t)
+
+miscfiles_read_generic_certs(insights_client_t)
+miscfiles_read_localization(insights_client_t)
+
+rhsmcertd_read_config_files(insights_client_t)
```
zpytela: rhsmcertd interface should be in an optional block

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+allow insights_client_t self:udp_socket { connect create getattr setopt };
+allow insights_client_t self:unix_stream_socket create_stream_socket_perms;
+
+manage_dirs_pattern(insights_client_t, insights_client_etc_t, insights_client_etc_t)
+manage_files_pattern(insights_client_t, insights_client_etc_t, insights_client_etc_t)
+files_etc_filetrans(insights_client_t, insights_client_etc_t, { dir file })
+
+manage_dirs_pattern(insights_client_t, insights_client_var_log_t, insights_client_var_log_t)
+manage_files_pattern(insights_client_t, insights_client_var_log_t, insights_client_var_log_t)
+logging_log_filetrans(insights_client_t, insights_client_var_log_t, { file dir })
```
zpytela: Both file and dir are needed? Anyway dir file would be better.
And if the log dir is packaged, it will be labeled on the package install.

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+# insights_client local policy
+#
+allow insights_client_t self:fifo_file rw_fifo_file_perms;
+allow insights_client_t self:tcp_socket { bind connect create getattr getopt setopt };
+allow insights_client_t self:udp_socket { connect create getattr setopt };
+allow insights_client_t self:unix_stream_socket create_stream_socket_perms;
+
+manage_dirs_pattern(insights_client_t, insights_client_etc_t, insights_client_etc_t)
+manage_files_pattern(insights_client_t, insights_client_etc_t, insights_client_etc_t)
+files_etc_filetrans(insights_client_t, insights_client_etc_t, { dir file })
```
zpytela: Both file and dir are needed for filetrans?
Does insights client need to write to only some of etc files? It'd be better to indicate with the type like `...etc_rw_t`.

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+
+manage_dirs_pattern(insights_client_t, insights_client_var_log_t, insights_client_var_log_t)
+manage_files_pattern(insights_client_t, insights_client_var_log_t, insights_client_var_log_t)
+logging_log_filetrans(insights_client_t, insights_client_var_log_t, { file dir })
+
+manage_dirs_pattern(insights_client_t, insights_client_tmp_t, insights_client_tmp_t)
+manage_files_pattern(insights_client_t, insights_client_tmp_t, insights_client_tmp_t)
+files_tmp_filetrans(insights_client_t, insights_client_tmp_t, { dir file })
```
zpytela, line 52: Both file and dir are needed for filetrans?

policy/modules/contrib/insights_client.te
```diff
@@ -0,0 +1,75 @@
...
+
+manage_dirs_pattern(insights_client_t, insights_client_var_lib_t, insights_client_var_lib_t)
+manage_files_pattern(insights_client_t, insights_client_var_lib_t, insights_client_var_lib_t)
+files_var_lib_filetrans(insights_client_t, insights_client_var_lib_t, { dir file })
```
zpytela, line 60: Both file and dir are needed for filetrans? Isn't the lib dir packaged? In that case it would be labeled on the package install.

zpytela (changes requested):
Additionally, please use tabs instead of spaces to indent.

policy/modules/contrib/insights_client.te
5umm3r15: thanks, it is not needed, I removed it

policy/modules/contrib/insights_client.te
5umm3r15: thanks, I removed file from it

policy/modules/contrib/insights_client.te
5umm3r15: I created etc_rw_t for certain files.

policy/modules/contrib/insights_client.te
5umm3r15: thanks, fixed

policy/modules/contrib/insights_client.te
5umm3r15: yes

policy/modules/contrib/insights_client.fc
```diff
@@ -0,0 +1,13 @@
...
+/etc/insights-client/[^/]+				--	gen_context(system_u:object_r:insights_client_etc_t,s0)
+/etc/insights-client/insights-client-egg-release	--	gen_context(system_u:object_r:insights_client_etc_rw_t,s0)
+/etc/insights-client/\.cache\.json\.asc			--	gen_context(system_u:object_r:insights_client_etc_rw_t,s0)
+/etc/insights-client/\.cache\.json			--	gen_context(system_u:object_r:insights_client_etc_rw_t,s0)
+/etc/insights-client/\.insights-core-gpg-sig\.etag	--	gen_context(system_u:object_r:insights_client_etc_rw_t,s0)
+
+/usr/bin/insights-client				--	gen_context(system_u:object_r:insights_client_exec_t,s0)
+
+/var/lib/insights/newest\.egg(/.*)?				gen_context(system_u:object_r:insights_client_var_lib_t,s0)
+/var/lib/insights/newest\.egg.asc			--	gen_context(system_u:object_r:insights_client_var_lib_t,s0)
```
zpytela: One more dot to escape.

policy/modules/contrib/insights_client.if
```diff
@@ -0,0 +1,39 @@
...
+##    Execute insights_client_exec_t in the insights_client domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##    Domain allowed to transition.
+## </summary>
+## </param>
+#
+interface(`insights_client_domtrans',`
+    gen_require(`
```
zpytela: Is it indented by spaces?

policy/modules/contrib/insights_client.fc
```diff
@@ -0,0 +1,13 @@
...
+
+/var/log/insights-client(/.*)?					gen_context(system_u:object_r:insights_client_var_log_t,s0)
```
zpytela, line 16: I also miss path entry in /var/run.

zpytela (changes requested):
If /var/lib/insights/ and /var/log/insights are a part of the package, they should be labelled properly at install time and there should be no need for manage_dirs_* and *_filetrans, in case of /var/lib also manage_files_*. It is different to transient files like those in /tmp and /run where it is needed.

5umm3r15:
> If /var/lib/insights/ and /var/log/insights are a part of the package, they should be labelled properly at install time and there should be no need for manage_dirs_* and __filetrans, in case of /var/lib also manage_files__. It is different to transient files like those in /tmp and /run where it is needed.

These files and dirs are not a part of package. They are created after a start of the service.

zpytela (approved):

zpytela:
> > If /var/lib/insights/ and /var/log/insights are a part of the package, they should be labelled properly at install time and there should be no need for manage_dirs_* and **filetrans, in case of /var/lib also manage_files**. It is different to transient files like those in /tmp and /run where it is needed.
> 
> These files and dirs are not a part of package. They are created after a start of the service.

All right. I think this is not correct and we should report it, but it is needed now. So thank you for the contribution, I'll merge it so that it goes to the next build.

PR #1057, opened 2022-02-08
Allow unconfined to run virtd bpf

5umm3r15:
Create virtd interface to allow caller domain to run bpf.
Apply virt_prog_run_bpf interface in unconfined domain.

Fixed BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2033504


zpytela:
Merging, thank you.

PR #1063, opened 2022-02-11
Allow alsa bind mixer controls to led triggers

zpytela:
Since v5.13, the kernel has support to bind certain alsa mixer controls
to LED triggers from userspace to control the mute-LEDS found on some
devices (typically embedded inside the keyboard's mute keys).

To allow that, alsa needs to be able to execute "modprobe snd_ctl_led"
and write to /sys/class/sound/ctl-led/speaker/ and .../mic.

Resolves: rhbz#1958210

### DIFF
--- a/policy/modules/contrib/alsa.te
+++ b/policy/modules/contrib/alsa.te
@@ -82,11 +82,13 @@ corecmd_exec_bin(alsa_t)
 
 dev_getattr_fs(alsa_t)
 dev_read_sound(alsa_t)
-dev_read_sysfs(alsa_t)
+dev_rw_sysfs(alsa_t)
 dev_read_urand(alsa_t)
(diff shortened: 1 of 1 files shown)


bdaase:
Is it by design that the pipeline is creating an F36 rpm when doing an F35 backport (and merging into the F35 branch)?

zpytela:
Almost always rawhide is a superset of rules of older branches, so they need to be backported to each f35-like branch if applicable there, too. Does not necessarily correlate with creating builds.

bdaase:
Ah, I thought those RPMs were supposed to be fore testing the changes by users (and then it would be good to have a "fitting" RPM).

zpytela:
While the changes go to rawhide right away with the build, it is still possible to get F35/F34 builds ready in advance.

PR #1067, opened 2022-02-11
filesystem.te: add genfscon rule for ntfs3 filesystem

WOnder93:
Kernel 5.15 introduced a new NTFS filesystem driver. Add the same
genfscon rule for it as there is for other NTFS filesystems.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2029039


zpytela:
Merging, thank you.

PR #1092, opened 2022-02-22
Allow sosreport dbus chat abrt systemd timedatex

5umm3r15:
Create sosreport dbus chat interface.

Allow abrt, systemd and timedatex to dbus chat sosreport


zpytela:
LGTM, thank you.
Note the checks failures are a result of new userspace tools.

5umm3r15:
thanks

Issue #1108, opened 2022-03-14
Add selinux policy for cloud-init-hotplug


yangzz-97:
Cloud-init introduces cloud-init-hotplug.socket requires systemd have add_name capability in the directory /run/cloud-init/hook-hotplug-cmd, I think we need to add a security policy for cloud init hotplug

https://bugs.launchpad.net/cloud-init/+bug/1936229

yangzz-97:
@zpytela Could you help me? Thank you

zpytela:
@yangzz-97 I am sorry for the delay in answering. Support for cloud-init was enhanced in selinux-policy in the past months, anyway I cannot reproduce the problem now. Please let me know if it works for you too.

zpytela:
This problem is not reproducible any longer, so closing the issue. Feel free to reopen it if you can.

PR #1110, opened 2022-03-15
systemd should be allowed add_name access on the hook-hotplug-cmd dir…

yangzz-97:
#1108


yangzz-97:
@zpytela Can you review the code? thank you

PR #1111, opened 2022-03-21
Label /var/run/ecblp0 pipe with cupsd_var_run_t

zpytela:
With the edce3e31ec2 (Label /var/run/ecblp0 as cupsd_var_run_t) commit,
default file context for /var/run/ecblp0 was defined for a plain file
instead of a named pipe which is actually used by epson drivers.

Resolves: rhbz#2061427


zpytela:
There is only this failure:
```
make: *** No rule to make target 'insights_client.if'
```
Seems still some rawhide fragments are used for the checks builds.

PR #1122, opened 2022-04-01
Add the io_uring class

WOnder93:
This adds the new io_uring class used to control the ability to use
special io_uring features by processes.

Initially allow general usage to unconfined domains. Further rules can
be granted to individual domains as needed (such io_uring usage should
be very rare).

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -1095,3 +1095,9 @@ class lockdown
     integrity
     confidentiality
 }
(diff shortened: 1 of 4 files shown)


zpytela:
Thank you. Supposing this should be backported to F35, too.

WOnder93:
Yes. Possibly even F34, which also has a 5.16.x kernel now.

PR #1145, opened 2022-04-12
New policy for stalld

5umm3r15:
New policy for stalld, a daemon that monitors threads on a system running low latency applications.
It checks for job threads that have been on a run-queue without being scheduled onto a CPU for a specified threshold.

When it detects a stalled thread, stalld temporarily changes the scheduling policy to SCHED_DEADLINE
and assigns the thread a slice of CPU time to make forward progress.

Copr build: https://copr.fedorainfracloud.org/coprs/nknazeko/selinux-policy/


zpytela:
Merging, thank you.

PR #1156, opened 2022-04-20
Update github actions to more strict rules of git 2.36

zpytela:


.github/workflows/build.yml
```diff
@@ -22,6 +22,7 @@ jobs:
     steps:
       - run: dnf install --nogpgcheck -y make git-core rpm-build 'dnf-command(builddep)'
       - uses: actions/checkout@v2
+      - run: git config --global --add safe.directory /__w/selinux-policy/selinux-policy
```
WOnder93: What about `chown -R $(id -u):$(id -g) .`?

zpytela: Well, this was one of the options while the PR was in a draft state. Now also the commit messages has been elaborated on, so it is ready for review.

zpytela:
Anyway thanks for the comment. I need to merge this PR now, but I understand there is still room for improvement.

PR #1159, opened 2022-04-22
Fix users for SELinux userspace 3.4

bachradsusi:
Latest yet to be released userspace version 3.4 added new validation and
discovered several issues in current implementation. This patch tries to
address them:

- move guest and xguest module from contrib to roles - refpolicy did
this change long time ago

- roles guest_r and xguest_r need to be defined in kernel.te

- gen_user() is supposed to be in policy/users, not in modules

- drop role multiple definitions from userdom_base_user_template as it's
and is supposed to be defined in kernel.te


zpytela:
LGTM, I will merge it soon. Thank you for your effort.

PR #1185, opened 2022-05-11
F36 nm dispatcher

zpytela:


zpytela:
Merging, the reported problem with build requires backporting rawhide changes for confined users to f36.

PR #1188, opened 2022-05-12
Allow sa-update to get init status and start systemd files

5umm3r15:
Allow sa-update, automate SpamAssassin rule updates, to
status init_t and start systemd_unit_file_t

Fixed: bz#2061844

### DIFF
--- a/policy/modules/contrib/spamassassin.te
+++ b/policy/modules/contrib/spamassassin.te
@@ -59,6 +59,12 @@ files_type(spamd_var_lib_t)
 type spamd_var_run_t;
 files_pid_file(spamd_var_run_t)
 
+type spamd_update_unit_file_t;
+systemd_unit_file(spamd_update_unit_file_t)
+
+type spamd_unit_file_t;
+systemd_unit_file(spamd_unit_file_t)
+
 ifdef(`distro_redhat',`
 	# spamassassin client executable
 	type spamc_t;
@@ -625,8 +631,6 @@ kernel_read_system_state(spamd_update_t)
 
 can_exec(spamd_update_t, spamd_update_exec_t)
 
-can_exec(spamd_update_t, spamd_update_exec_t)
-
 # for updating rules
 corenet_tcp_connect_http_port(spamd_update_t)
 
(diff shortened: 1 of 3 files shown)


policy/modules/contrib/spamassassin.te
```diff
@@ -641,6 +641,9 @@ domain_read_all_domains_state(spamd_update_t)
 auth_use_nsswitch(spamd_update_t)
 auth_dontaudit_read_shadow(spamd_update_t)
 
+init_status(spamd_update_t)
+systemd_start_all_services(spamd_update_t)
```
zpytela: Instead of this, the spamassassin.service unit file should get a new private type so that sa-update is allowed to handle only this type.

zpytela (changes requested):

policy/modules/contrib/spamassassin.te
```diff
@@ -659,12 +660,29 @@ optional_policy(`
...
+optional_policy(`
+	init_status(spamd_update_t)
+')
+
+optional_policy(`
+	mta_read_config(spamd_update_t)
+')
+
+optional_policy(`
+	spamassassin_systemctl(spamd_update_t)
```
zpytela: Interfaces like this should be used only outside the module. I believe something like `spamd_unit_file_t:file read_file_perms` would work unless you have more denials.

policy/modules/contrib/spamassassin.te
```diff
@@ -659,12 +660,29 @@ optional_policy(`
...
+')
+
+optional_policy(`
+	systemd_exec_systemctl(spamd_update_t)
+')
+
+optional_policy(`
+	userdom_search_admin_dir(spamd_update_t)
```
zpytela, line 678: userdom_search_user_home_dirs is not needed?

zpytela (changes requested):

policy/modules/contrib/spamassassin.te
5umm3r15: this interface was only added to optional policy block and it is not related to the bz

policy/modules/contrib/spamassassin.te
5umm3r15: SELinux denied spamd_update_t to start and status /usr/lib/systemd/system/spamassassin.service

zpytela:
OK, merging, thank you.

zpytela (approved):

PR #1189, opened 2022-05-13
Move the chown call into make-srpm.sh

WOnder93:
This workaround for Git CVE-2022-24765 fix is also needed when
make-srpm.sh is run inside COPR, so move it there. It also makes a bit
more sense as that's [the only place] where the git operations happen.

### DIFF
--- a/.copr/make-srpm.sh
+++ b/.copr/make-srpm.sh
@@ -14,6 +14,10 @@ EXPANDER_URL=https://github.com/fedora-selinux/macro-expander
 
 rpm -q rpm-build git-core || dnf install -y rpm-build git-core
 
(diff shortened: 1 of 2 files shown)


zpytela:
Does it deprecate commit 6b4fa7ffafc89f2fdd1addfdb881c86a3b36dfd0?

WOnder93:
Hm, actually it might really cover both cases (running inside COPR and GH Actions)... Let me check.

WOnder93:
Yep, we can undo 6b4fa7ffafc89f2fdd1addfdb881c86a3b36dfd0 with this commit, so I updated the PR.

zpytela:
Merging, thank you.

PR #1191, opened 2022-05-16
Dontaudit domain map permission on directories

zpytela:
type=AVC msg=audit(1651700801.804:609): avc:  denied  { map } for  pid=5411 comm="stress-ng" path="/proc/10381" dev="proc" ino=67726 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=dir permissive=0

which can be triggered by "stress-ng --all 8 --timeout 60s --metrics-brief --tz".

Resolves: rhbz#2081864

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -172,6 +172,9 @@ files_append_inherited_tmp_files(domain)
 files_mmap_usr_files(domain)
 files_read_all_base_ro_files(domain)
 files_dontaudit_getattr_kernel_symbol_table(domain)
+files_dontaudit_map_all_dirs(domain)
+
+fs_dontaudit_map_all_dirs(domain)
 
 # All executables should be able to search the directory they are in
 corecmd_search_bin(domain)
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -1854,6 +1854,25 @@ interface(`files_dontaudit_search_all_dirs',`
 
 ########################################
 ## <summary>
+##	Do not audit attempts to map
+##	file_type directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`files_dontaudit_map_all_dirs',`
+	gen_require(`
+		attribute file_type;
+	')
+
+	dontaudit $1 file_type:dir map;
+')
+
+########################################
+## <summary>
 ##	Get the attributes of all filesystems
 ##	with the type of a file.
 ## </summary>
--- a/policy/modules/kernel/files.te
+++ b/policy/modules/kernel/files.te
@@ -249,7 +249,8 @@ fs_associate_tmpfs(tmpfsfile)
 
 # Create/access any file in a labeled filesystem;
 allow files_unconfined_type file_type:{ file chr_file } ~{ execmod entrypoint };
-allow files_unconfined_type file_type:{ dir lnk_file sock_file fifo_file blk_file } *;
+allow files_unconfined_type file_type:dir ~map;
+allow files_unconfined_type file_type:{ lnk_file sock_file fifo_file blk_file } *;
 allow files_unconfined_type file_type:service *;
 
 # Mount/unmount any filesystem with the context= option.
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -6540,6 +6540,25 @@ interface(`fs_dontaudit_getattr_all_dirs',`
 
 ########################################
 ## <summary>
+##	Dontaudit map of all directories
+##	with a filesystem type.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_dontaudit_map_all_dirs',`
+	gen_require(`
+		attribute filesystem_type;
+	')
+
+	dontaudit $1 filesystem_type:dir map;
+')
+
+########################################
+## <summary>
 ##	Search all directories with a filesystem type.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/kernel/filesystem.te
+++ b/policy/modules/kernel/filesystem.te
@@ -364,4 +364,5 @@ allow filesystem_unconfined_type filesystem_type:filesystem all_filesystem_perms
 # pseudo filesystem types that are applied to both the filesystem
 # and its files.
 allow filesystem_unconfined_type filesystem_type:{ file } ~entrypoint;
-allow filesystem_unconfined_type filesystem_type:{ dir lnk_file sock_file fifo_file chr_file blk_file } *;
+allow filesystem_unconfined_type filesystem_type:dir ~map;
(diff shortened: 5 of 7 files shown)


WOnder93:
@zpytela Were you able to reproduce this? Because I tried it last week and wasn't successful. `stress-ng` is something of a fuzz test for the kernel and this denial looks really strange - I don't think we should be adding it to the policy without understanding what is really going on.

zpytela:
Reproduced on a physical machine, anyway not much data available.
```

type=SYSCALL msg=audit(05/13/22 11:07:27.139:63912) : arch=x86_64 syscall=mmap success=no exit=EACCES(Permission denied) a0=0x0 a1=0x1000 a2=PROT_READ a3=MAP_PRIVATE items=0 ppid=1859 pid=4098006 auid=user uid=user gid=user euid=user suid=user fsuid=user egid=user sgid=user fsgid=user tty=pts6 ses=3 comm=stress-ng exe=/usr/bin/stress-ng subj=staff_u:staff_r:staff_t:s0-s0:c0.c1023 key=(null)
type=AVC msg=audit(05/13/22 11:07:27.139:63912) : avc:  denied  { map } for  pid=4098006 comm=stress-ng path=/proc/4099867 dev="proc" ino=68256165 scontext=staff_u:staff_r:staff_t:s0-s0:c0.c1023 tcontext=staff_u:staff_r:staff_t:s0-s0:c0.c1023 tclass=dir permissive=0
```

WOnder93:
So I have managed to reproduce and understand this. One of the `stress-ng`'s stressors, `pidfd`, randomly [opens a file descriptor](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L46-L63) either via `pidfd_open(<pid>, 0)` or `openat(AT_FDCWD, "/proc/<pid>", O_RDONLY|O_CLOEXEC|O_DIRECTORY)` and then [attempts to do `mmap(NULL, 4096, PROT_READ, MAP_PRIVATE, 4, 0)`](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L176-L180), which is illegal on both, but in case of the directory fd it hits this non-allowed access vector.

Since calling `mmap(2)` on a directory should always fail (with `ENODEV`) and there is no easy way to get it to fail before hitting the SELinux check, I think we should dontaudit this (globally) instead of allowing. So I suggest something like:
```
# mmap'ing a directory is not supported and will fail one way or another.
# Silence such denials to avoid triggering denials on odd attempts to do so
# (such as running stress-ng --pidfd ...).
dontaudit domain { domain files_type }:dir map;
```
(This would need to be split across the modules somehow - not sure about the best way.)

zpytela:
> So I have managed to reproduce and understand this. One of the `stress-ng`'s stressors, `pidfd`, randomly [opens a file descriptor](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L46-L63) either via `pidfd_open(<pid>, 0)` or `openat(AT_FDCWD, "/proc/<pid>", O_RDONLY|O_CLOEXEC|O_DIRECTORY)` and then [attempts to do `mmap(NULL, 4096, PROT_READ, MAP_PRIVATE, 4, 0)`](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L176-L180), which is illegal on both, but in case of the directory fd it hits this non-allowed access vector.
> 
> Since calling `mmap(2)` on a directory should always fail (with `ENODEV`) and there is no easy way to get it to fail before hitting the SELinux check, I think we should dontaudit this (globally) instead of allowing.

Interesting, thanks for digging into it. Still: if the operation is illegal, shouldn't rather the application be fixed not to do it?

I also suppose existing permissions should be removed, like https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/services/xserver.te#L513
(for https://bugzilla.redhat.com/show_bug.cgi?id=1725509) or general rules with `*` like
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/filesystem.te#L367

In case of stress-ng, I now think that even exposing the denials is not completely incorrect.

> So I suggest something like:
> 
> ```
> # mmap'ing a directory is not supported and will fail one way or another.
> # Silence such denials to avoid triggering denials on odd attempts to do so
> # (such as running stress-ng --pidfd ...).
> dontaudit domain { domain files_type }:dir map;
> ```
> 
> (This would need to be split across the modules somehow - not sure about the best way.)

zpytela:
@wonder93, I made some changes following your research outcome, please review.
I am still a bit reluctant though if the dontaudit rules should be in place at all, or if it should be dontaudited e. g. for unconfined_t only.
Another concern I have is why the access was requested (in the past, not many details to find) for xdm_t and x_userdomain.

WOnder93:
> > So I have managed to reproduce and understand this. One of the `stress-ng`'s stressors, `pidfd`, randomly [opens a file descriptor](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L46-L63) either via `pidfd_open(<pid>, 0)` or `openat(AT_FDCWD, "/proc/<pid>", O_RDONLY|O_CLOEXEC|O_DIRECTORY)` and then [attempts to do `mmap(NULL, 4096, PROT_READ, MAP_PRIVATE, 4, 0)`](https://github.com/ColinIanKing/stress-ng/blob/ac04cbd1002cffcf525b19c1c6ab119c2ee60917/stress-pidfd.c#L176-L180), which is illegal on both, but in case of the directory fd it hits this non-allowed access vector.
> > Since calling `mmap(2)` on a directory should always fail (with `ENODEV`) and there is no easy way to get it to fail before hitting the SELinux check, I think we should dontaudit this (globally) instead of allowing.
> 
> Interesting, thanks for digging into it. Still: if the operation is illegal, shouldn't rather the application be fixed not to do it?

In general it should, but in case of stress-ng the intent is to trigger all kinds of odd code paths in the kernel, trying to weed out any hidden bugs. So I don;t think we can expect it to "fix" it.

> @WOnder93, I made some changes following your research outcome, please review. I am still a bit reluctant though if the dontaudit rules should be in place at all, or if it should be dontaudited e. g. for unconfined_t only.

I think it's better to globally dontaudit it. SELinux is not intended to be a debugging tool, but rather a security control tool. It's not necessary (and from the user POV also not desirable) to generate audit noise for everything that looks like a programmer error. The application will get an error back from the syscall with or without SELinux and that should be sufficient (once the programmer sees what is being attempted, it's not hard to realize that it's an invalid operation).

IOW, we want to audit a denial when the process is trying to do something outside its defined security boundary, not necessarily when it tries to do some nonsense that we know would fail anyway.

> Another concern I have is why the access was requested (in the past, not many details to find) for xdm_t and x_userdomain.

Probably because the previous maintainer(s) often didn't think too much before allowing things ;)

WOnder93 (approved):

PR #1230, opened 2022-06-10
Allow nm-dispatcher custom plugin create and use unix dgram socket

zpytela:
type=AVC msg=audit(1654238161.81:4923): avc:  denied  { create } for  pid=696389 comm="logger" scontext=system_u:system_r:NetworkManager_dispatcher_custom_t:s0 tcontext=system_u:system_r:NetworkManager_dispatcher_custom_t:s0 tclass=unix_dgram_socket permissive=1

Resolves: rhbz#2093155

### DIFF
--- a/policy/modules/contrib/ddclient.if
+++ b/policy/modules/contrib/ddclient.if
@@ -100,3 +100,21 @@ interface(`ddclient_admin',`
 	files_list_tmp($1)
 	admin_pattern($1, ddclient_tmp_t)
 ')
+
+########################################
+## <summary>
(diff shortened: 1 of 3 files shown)


rakuco:
For the record, as mentioned in https://bodhi.fedoraproject.org/updates/FEDORA-2022-fd22b79a84, this does not fully fix https://bugzilla.redhat.com/show_bug.cgi?id=2092808 because 30-winbind is still failing to invoke `testparm` (the invocation returns an empty string instead of the expected value).

PR #1234, opened 2022-06-16
Allow init_t to rw insights_client unnamed pipe

5umm3r15:
Allow init_t read and write a insights_client unnamed pipe.


zpytela:
Merging, thank you.

PR #1235, opened 2022-06-16
Add rpm setattr db files macro

5umm3r15:
Add interface to set the attributes of RPM package database.


zpytela:
Merging, thank you.

PR #1252, opened 2022-06-22
Allow domain use userfaultfd over all domains

zpytela:
Until now, all processes were allowed to use userfaultfd as well other
anon_inodes to get a file descriptor from the same domain.
Since this commit the permissions are allowed between different domains.


WOnder93 (approved):

zpytela:
Merging and backporting to F36, hope that's sufficient.

PR #1274, opened 2022-07-13
fix(sedoctool.py): Fix syntax warning: "is not" with a literal

gyptazy:
## General
Running the make target `policy` throws a syntax warning on newer Py3 versions:
```
/home/vagrant/selinux-policy-targeted-fedora/selinux-policy/support/sedoctool.py:275: SyntaxWarning: "is not" with a literal. Did you mean "!="?
```

This PR fixes this warning when using Py3.8+ versions.


zpytela:
Merging, thank you.

PR #1291, opened 2022-07-26
Label /etc/insights-client/machine-id

5umm3r15:
Add label insights_client_etc_rw_t
for /etc/insights-client/machine-id


zpytela:
Merging, thank you.

PR #1292, opened 2022-07-26
Allow rhsmcertd to read insights config files

5umm3r15:
Rhsmcertd obtains system information, including insights-client ID.

Add interface that allows domain to read insights_client config files.
Allow rhsmcertd to read insights_client config files.


zpytela:
Merging, thank you.

PR #1295, opened 2022-07-27
Allow init work with its private ramfs files

zpytela:
The fs_ramfs_filetrans() interface was added.

Resolves: rhbz#2096857

### DIFF
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -188,6 +188,7 @@ interface(`files_stub_tmp',`
 ##		<li>files_spool_file()</li>
 ##		<li>files_tmp_file()</li>
 ##		<li>files_tmpfs_file()</li>
+##		<li>files_ramfs_file()</li>
 ##		<li>logging_log_file()</li>
 ##		<li>userdom_user_home_content()</li>
 ##	</ul>
@@ -589,6 +590,26 @@ interface(`files_tmp_file',`
 ## </param>
 #
 interface(`files_tmpfs_file',`
+	gen_require(`
+		attribute tmpfsfile;
+	')
+
+	files_type($1)
+	typeattribute $1 tmpfsfile;
+')
+
+########################################
+## <summary>
+##	Transform the type into a file, for use on a
+##	virtual memory filesystem (ramfs).
+## </summary>
+## <param name="type">
+##	<summary>
+##	The type to be transformed.
+##	</summary>
+## </param>
+#
+interface(`files_ramfs_file',`
 	gen_require(`
 		attribute tmpfsfile;
 	')
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -5716,6 +5716,41 @@ interface(`fs_tmpfs_filetrans',`
 
 ########################################
 ## <summary>
+##	Create an object in a ramfs filesystem, with a private
+##	type using a type transition.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="private type">
+##	<summary>
+##	The type of the object to be created.
+##	</summary>
+## </param>
+## <param name="object">
+##	<summary>
+##	The object class of the object being created.
+##	</summary>
+## </param>
+## <param name="name" optional="true">
+##	<summary>
+##	The name of the object being created.
+##	</summary>
+## </param>
+#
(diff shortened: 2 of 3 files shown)


zpytela:
@WOnder93 I'd appreciate if you could take a look. This may look like an overkill given that so far only systemd used the ramfs files.

WOnder93:
I'm thinking if it wouldn't be better (and actually easier) to just use the same labeling for tmpfs and ramfs. I.e. make `ramfs_t` an alias of `tmpfs_t` and replace the ramfs interfaces with the equivalent tmpfs ones. Both filesystems are very similar in principle and as the bug shows, they are apparently even interchangeable at least in some scenarios. Unifying labeling for the two would simplify things nicely without losing anything significant.

zpytela:
I considered this approach, but why a different label exists then? So I only reused tmpfsfile attribute. There are now about 20 other references to ramfs interfaces.

The way how systemd uses it it tries to mount the file as ramfs first and then tmpfs in case of a failure. Certainly the file on the filesystem is used the same way.

WOnder93:
Originally ramfs didn't support extended attributes, so it had to be treated differently from tmpfs. There is little security benefit to be had for labeling the two filesystems differently and now that ramfs has started to be used more intensively by systemd, it will be much simpler to merge them into a common label at the policy level. Then any tmpfs mount can be swapped for a ramfs one and it will work transparently, since the same label and policy rules will be used.

The `tmpfsfile` attribute is intended to group the various `*_tmpfs_t` types that can occur on a tmpfs mount. Adding ramfs_t to that attribute on its own won't make this systemd's assumption "ramfs is just like tmpfs so I can mount it wherever I would normally mount tmpfs" reflected in the policy. (And the fact that you had to add a new transition rule proves it.)

Since I'm the one bitching about it, I'll try to propose a PR, hopefully sometime this week...

jdoss:
Can we get this merged in the meantime while @WOnder93 gets the other PR sorted out? Fedora 37 can't use systemd-creds with SELinux enabled and that is a bummer. :(

zpytela:
@jdoss Do you see any actual problem? AFAIK systemd should fall back to using tmpfs if ramfs is not available.

jdoss:
Hey @zpytela, yeah it is an actual problem still. See https://bugzilla.redhat.com/show_bug.cgi?id=2096857#c12 for more info.

systemd-251.5-607.fc37.x86_64 and selinux-policy-37.12-2.fc37.noarch on Fedora 37

```
$ sudo setenforce 0
$ sudo systemd-run -P --wait -p LoadCredential=abc:/etc/hosts systemd-creds cat abc
Running as unit: run-u5979.service
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
Finished with result: success
Main processes terminated with: code=exited/status=0
Service runtime: 15ms 
CPU time consumed: 8ms
$ sudo setenforce 1
$ sudo systemd-run -P --wait -p LoadCredential=abc:/etc/hosts systemd-creds cat abc
Running as unit: run-u5995.service
Finished with result: exit-code
Main processes terminated with: code=exited/status=243
Service runtime: 4ms
CPU time consumed: 4ms
```

jdoss:
Here is an ausearch that shows the denial when trying to use ramfs. systemd is not falling back to tmpfs.

```
$ sudo ausearch -m avc -m user_avc -m selinux_err -i -ts today
----
type=AVC msg=audit(11/22/2022 10:26:44.604:16255) : avc:  denied  { read write open } for  pid=1308009 comm=(sd-mkdcreds) path=/dev/shm/.#cred3a4868acd3fd2e1b dev="ramfs" ino=6217355 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:ramfs_t:s0 tclass=file permissive=0
```

zpytela:
Learned today: There actually is a tmpfs fallback, but it only applies if the ramfs volume cannot be mounted.
We will resolve this issue either way in the next build.

jdoss:
Thank you @zpytela!!

WOnder93:
FWIW: I have now opened #1491, which implements my tmpfs+ramfs merge idea. Sorry for the delay :/

zpytela:
@jdoss, I've just merged the other PR and it will be a prt of the next build. I am sorry for not resolving it earlier as I misunderstood the fallback concept.

jdoss:
Thanks to both of you @zpytela @WOnder93! 🎉

PR #1311, opened 2022-08-05
fix(system/init.fc): Fix section description

gyptazy:
Fixes the corresponding section description in `system/init.fc`.


zpytela:
Merging, thank you.

PR #1316, opened 2022-08-05
Allow systemd_hostnamed label /run/systemd/* as hostnamed_etc_t

5umm3r15:
Allow systemd_hostnamed_t to create dirs and files in /run/systemd/* with label hostnamed_etc_t
Names of these files and dirs include hashes.

Fix: bz#1976684


policy/modules/system/systemd.te
```diff
@@ -851,6 +851,7 @@ allow systemd_hostnamed_t self:unix_dgram_socket create_socket_perms;
 manage_files_pattern(systemd_hostnamed_t, hostname_etc_t, hostname_etc_t)
 manage_lnk_files_pattern(systemd_hostnamed_t, hostname_etc_t, hostname_etc_t)
 files_etc_filetrans(systemd_hostnamed_t, hostname_etc_t, file)
+init_pid_filetrans(systemd_hostnamed_t, hostname_etc_t, { dir file })
```
zpytela: afaik it is only the default-hostname file which is created there

zpytela (commented):
Have you managed to run any related test with this change?

5umm3r15:
tested with all available tests, no problems.

zpytela:
Merging, thank you.

PR #1323, opened 2022-08-09
Allow systemd-modules-load write to /dev/kmsg and send mssg to syslogd

5umm3r15:
Allow systemd_modules_load_t, early boot service that loads kernel modules,
write to the kernel messages device and send a message to syslogd over a unix domain datagram socket.

Fix: bz#2088257


zpytela:
Merging, thank you.

PR #1332, opened 2022-08-11
Allow unconfined_t transition to targetclid_home_t

5umm3r15:
If the targetcli command is executed before starting the targetclid service,
then the /root/.targetcli/ directory (and all files in it) is mislabeled.
The start of targetclid service in such situation then leads to SELinux denials.

Resolves: BZ#2106360


zpytela:
I believe the same interface should be used for sysadm_t.

zpytela:
Merging, thank you.

PR #1341, opened 2022-08-22
Allow system_dbusd ioctl kernel with a unix stream sockets

5umm3r15:
Systemd creates and starts to listen on the socket before
SELinux policy is loaded and it inherits the kernel secid as its label.

Allow system_dbusd ioctl kernel with a unix stream sockets
Create interface to allow the specified domain to ioctl a kernel with a unix domain stream sockets.

Fix: bz#2085392


zpytela:
Merging, thank you.

PR #1342, opened 2022-08-22
Update tor_bind_all_unreserved_ports interface

5umm3r15:
When enabled boolean tor_bind_all_unreserved_ports,
allow tor bind UDP sockets to all ports > 1024.

Fix: bz#2089486


zpytela:
Merging, thank you.

PR #1343, opened 2022-08-22
Allow chronyd bind UDP sockets to ptp_event ports

5umm3r15:
As a new feature added in chrony-4.2, chronyd supports an experimental NTP-over-PTP protocol.
It needs it for timestamping PTP event messages containing NTP messages.

Fix: #bz2118631


zpytela:
Please instead of a generic allow rule label the port with ptp_event_port_t, udp protocol should suffice.

zpytela:
IANA only describes the port as PTP event messages.
Copying information from the chrony developer:
It needs it for timestamping PTP event messages containing NTP messages.
Please use it in the commit message(s).

5umm3r15:
Fixed, can you please review it? @zpytela

zpytela:
Merging, thank you.

PR #1346, opened 2022-08-23
Allow stalld get and set scheduling policy of all domains.

5umm3r15:
Allow stalld get and set scheduling policy of all domains.

Stalld needs to have full access to sched_getattr() or sched_setattr()
to any type of "program" (daemon, kernel threads, user commands... etc),
which includes common process started from a commandline with unconfined_t type.

Fixed: bz#2105038

### DIFF
--- a/policy/modules/contrib/stalld.te
+++ b/policy/modules/contrib/stalld.te
@@ -36,7 +36,9 @@ kernel_setsched(stalld_t)
 
 dev_read_sysfs(stalld_t)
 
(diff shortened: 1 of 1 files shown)


zpytela:
How I understand the problem stalld requires the access to any process, i. e. the domain attribute.

5umm3r15:
@zpytela I allowed it for all domains, can you review it please?

zpytela:
Merging, thank you.

PR #1372, opened 2022-09-07
blueman-mechanism can read ~/.local/lib/python*/site-packages directory

milosmalik:
If the ~/.local/lib/python*/site-packages/ directory exists in root's home directory then the blueman-mechanism service tries to read that directory during its start.

The blueman-mechanism program is written in Python and I believe that Python is trying to find locally installed python modules.

In order to avoid these SELinux denials, SELinux policy should allow the access.

Resolves: BZ#2027044

### DIFF
--- a/policy/modules/contrib/blueman.te
+++ b/policy/modules/contrib/blueman.te
@@ -93,7 +93,7 @@ optional_policy(`
 ')
 
 optional_policy(`
-	gnome_search_gconf(blueman_t)
+	gnome_search_gconf_data_dir(blueman_t)
 ')
 
 optional_policy(`


policy/modules/contrib/blueman.te
```diff
@@ -103,3 +103,6 @@ optional_policy(`
 optional_policy(`
 	xserver_read_state_xdm(blueman_t)
 ')
+
+gnome_search_gconf_data_dir(blueman_t)
```
zpytela: Looks this is a superset call of existing gnome_search_gconf(), so it can replace it. It anyway needs to be in an optional block (between 'd' and 'i' ones).

Also note this will resolve the problem only for non-root users, but I am fine with it.

zpytela:
Please see the comments inline.

zpytela (changes requested):

milosmalik:
I updated the PR based on your feedback.

zpytela (approved):

zpytela:
Merging, thank you.

wadebee:
I am attempting to install a brand new instance of Fedora 36 and was immediately met with this warning upon initial login. Could you please provide instructions for how to properly apply this patch?

PR #1373, opened 2022-09-07
pidof executed by abrt can readlink /proc/*/exe

milosmalik:
At least one of the ABRT addons calls `pidof abrtd` which leads to { sys_ptrace } SELinux denials in cap_userns class.

In order to support the full functionality of ABRT and its addons, I believe that SELinux policy should allow this access.

Resolves: BZ#2071586


zpytela:
The content is ok, but please move it up next to abrt_t capabilities.

milosmalik:
I updated the PR based on your feedback.

zpytela:
Merging, thank you.

PR #1374, opened 2022-09-07
Add the new 'cmd' permission to the 'io_uring' class

WOnder93:
This permission has been added in kernel version 6.0, but is planned to be backported to 5.19 stable as well (since the IORING_OP_URING_CMD support was added in kernel 5.19).

Initially allow it for unconfined domains on all files. The usage of this type of io_uring command should be very rare (if any), so I don't expect any major disruption from this change.


zpytela:
Merging, thank you.

PR #1428, opened 2022-10-05
Allow targetclid to manage tmp files 

5umm3r15:
Resolves: bz#2127408


zpytela:
In the bz only plain files were created, did you see also directories in some other scenarios?

5umm3r15:
I got only plain files, thanks.

zpytela:
Merging, thank you.

PR #1434, opened 2022-10-07
gLabel port 15354/tcp and 15354/udp with opendnssec

5umm3r15:
OpenDNSSEC is a policy-based zone signer
that automates the process of keeping track of DNSSEC keys and the signing of zones.
The default value of the listening port for Signer is 15354

Allow opendnssec to bind to 15354/tcp and 15354/udp.

Resolves: bz#2057501


5umm3r15:
Tested with our tests, no new AVC messages.

Copr build:
`dnf copr enable nknazeko/selinux-policy`
`dnf update selinux policy`

zpytela:
LGTM

PR #1439, opened 2022-10-11
Label ports 10161-10162 tcp/udp with snmp

5umm3r15:
Simple Network Management Protocol over TLS and DTLS uses 10161-10162 ports. https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=snmp

Resolves: bz#2133221


zpytela:
LGTM

PR #1452, opened 2022-11-02
Add watch_sb and watch interfaces

5umm3r15:
Add interafaces that allow watch_sb user temporary directories, tmpfs directories, boot and root directories.
Add interfacecs that allow watch filesystem.


5umm3r15:
https://bugzilla.redhat.com/show_bug.cgi?id=2139363

zpytela:
Note "using the context= mount option." does not apply here.

PR #1455, opened 2022-11-07
Add watch dosfs interface

5umm3r15:
Add interfacec that allow watch dosfs filesystem.

Resolves: rhbz#2139363


zpytela:
Merging, thank you.

PR #1457, opened 2022-11-09
Allow snapper to create snapshots of /home/ subvolume/partition

levi-jcbs:


zpytela:
@levi-jcbs the pr commit looks good, but can you gave me a usage example?

levi-jcbs:
@zpytela 

I previously faced the problem, that creating a snapper configuration of the home partition (or subvolume) didn't work:
`snapper -c home create-config /home`

So I had to add the SELinux rule myself:
`semanage fcontext -a -t snapperd_data_t '/home/\.snapshots(/.*)?'`

It would be better, if this is the default behaviour. Snapshotting the home subvolume/partition is pretty common.

In my opinion, the best thing would be to allow snapper to create `.snapshot` directories everywhere. If you all agree with that, I would do a PR changing the snapper rule that way.

levi-jcbs:
FYI currently snapper is only allowed to snapshot

- /
- /usr
- /var
- /etc

wich is quite arbitrarily. Users actually want to be able to take snapshots anywhere.

zpytela:
Merging, thank you and sorry for the delay.

PR #1466, opened 2022-11-16
Add the user_namespace security class

zpytela:


WOnder93:
This needs to also give the permission to `unconfined_domain_type` (like e.g. in 0846d11da6b70950ac446279d356ffa0177a3051).

zpytela:
> This needs to also give the permission to `unconfined_domain_type` (like e.g. in [0846d11](https://github.com/fedora-selinux/selinux-policy/commit/0846d11da6b70950ac446279d356ffa0177a3051)).

You are right, added.

PR #1468, opened 2022-11-17
systemd: allow sys_admin for gpt-generator

QuintinHill:
In Fedora 37 it seems systemd-gpt-auto-generator needs sys_admin capability. 
Hopefully fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2141998

Signed-off-by: Quintin Hill <stuff@quintin.me.uk>

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1163,7 +1163,7 @@ systemd_read_efivarfs(systemd_hwdb_t)
 # systemd_gpt_generator domain
 #
 
-allow systemd_gpt_generator_t self:capability sys_rawio;
+allow systemd_gpt_generator_t self:capability { sys_admin sys_rawio };
 allow systemd_gpt_generator_t self:netlink_kobject_uevent_socket create_socket_perms;
 
 dev_read_sysfs(systemd_gpt_generator_t)


zpytela:
@QuintinHill Please add more information to the commit message, like triggering condition, syscall, library call. The sys_admin capability is too powerful so needs proper justification.

QuintinHill:
@zpytela I've raised this fixing this another way in https://github.com/systemd/systemd/issues/24431.  But from that we at least learn that this is relevant only to Fedora 37 and not to Rawhide (since it is fixed already in the systemd version there).

QuintinHill:
@zpytela Actually https://github.com/systemd/systemd/commit/1b010ae7c96d953f11a9925d4c5aef29befc86d1 introduces use of the BLKPG_ADD_PARTITION ioctl which requires sys_admin (https://github.com/torvalds/linux/blob/master/block/ioctl.c#L23) to the library code used by systemd-gpt-auto-generator.   I was incorrect in my earlier comment that this is fixed in Rawhide that only removed use of the BLKPG_DEL_PARTITION ioctl that only appeared in an unreleased version of systemd.  I've reworded the commit message.

PR #1471, opened 2022-11-22
Add policy for rshim

5umm3r15:
rshim is the user-space rshim driver for BlueField SoC. It provides ways to access the rshim resources on the BlueField target via USB or PCIe from external host machine.

Resolves: rhbz#2080439


zpytela:
Merging, thank you.

PR #1475, opened 2022-11-23
Harden against usermode helper path overwriting

WOnder93:
The goal here is to restrict the `kernel_t` domain such that it is not allowed to execute arbitrary binary files as usermode helpers and such that it always transitions to a minimal separate domain when doing so. The motivation is to mitigate privilege escalation via overwriting paths to usermode helpers (such as `modprobe`). This method is nicely described e.g. in [this writeup](https://lkmidas.github.io/posts/20210223-linux-kernel-pwn-modprobe/),

This is achieved by:
1. Reverting a bad commit that grants the `execute_no_trans` when it's not logical or necessary.
2. Explicitly adding several rules to `kernel_t` that are required for normal operation.
3. Adding missing transitions for `kernel_t` executing known usermode helpers.
4. Drop `unconfined_domain_noaudit(kernel_t)`.
   Unconfined domains are inherently allowed to execute any file, even without a transition. Thus we need to either a) add a way to make a domain unconfined with the exception of executing files and use it for `kernel_t` or b) make `kernel_t` a confined domain and add any missing rules explicitly.
   Naturally, b) is more risky as there may be corner cases not currently allowed that will need to be addressed based on users' reports, but it is much simpler to do (and testing on my own machine identified only a few missing rules), so I opted for it over a).

See also the following proposed test that verifies this change (to the extend that is automatically testable):
https://src.fedoraproject.org/tests/selinux/pull-request/342

I have installed a policy with these changes applied on my workstation and didn't observe any regressions. It is of course possible that there are legitimate usage patterns that would expose a missing rule, but at least this provides confidence that the policy coverage is close to complete and that the changes shouldn't generate much extra work to close the gap. I volunteer to investigate and fix (open PRs) any denials resulting from this PR :)

NB: This mitigation is only effective when the attacker is not able to set arbitrary labels on files (i.e. `unconfined_t` can still partially circumvent it - e.g. by relabeling the malicious binary to `init_exec_t` and gain privileges of `init_t`, which is practically almost unconfined).


zpytela:
Merging, thank you.

PR #1491, opened 2022-12-01
Reuse tmpfs_t also for the ramfs filesystem

WOnder93:
systemd-creds apparently runs in a chroot where /dev/shm is backed by ramfs instead of the usual tmpfs. ramfs currently uses a separate type (ramfs_t), but both ramfs and tmpfs can in fact be used for the same things interchangeably and since commit a7697467e082 ("filesystem: add fs_use_trans for ramfs"), they also use the same superblock labeling scheme, and thus it would be natural to use a single type for both.

This commit implements this idea, dropping the ramfs_t type and instead using the tmpfs_t type in its place (with ramfs_t being an alias for tmpfs_t for backwards compatibility).

The ramfs-specific interfaces are deprecated and references to them are removed from the policy (as all of them seem to be related to rhgb or readahead, which are not currently shipped in Fedora and will likely be completely removed from the policy in the future).

I ran my work laptop with a policy including this patch for several days and noted no regressions or denials.

This is an alternative to #1295.


zpytela:
Merging, thank you.

PR #1495, opened 2022-12-06
Do not label /usr/bin/systemd-notify with systemd_notify_exec_t

zpytela:
The systemd-notify binary was not expected to use as a separate service, but rather a helper to notify PID 1 about some particular service state. This includes user services. As a result, particular label for the binary does not seem to be helpful.


zpytela:
It actually seems to help with allowing daemons execute systemd_notify.
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/unconfined.te#L38

PR #1500, opened 2022-12-08
Make `bootc` be `install_exec_t`

cgwalters:
For the same reasons ostree and rpm-ostree are.
xref https://github.com/containers/bootc/issues/24


zpytela:
Merging to rawhide, thank you.

PR #1506, opened 2022-12-12
Allow admin users to create user namespaces

WOnder93:
Admin users should definitely be able to create them. Without this patch, for example, the selinux-testsuite (when run as sysadm_t) detects user namespaces as unsupported and skips the cap_userns test.


zpytela:
Merging, thank you.

PR #1508, opened 2022-12-12
Allow kernel threads to use fds from all domains

WOnder93:
In specific scenrios, the kernel may need to access file handles opened by userspace domains. For example, selinux-testsuite fails on the filesystem tests unless kernel_t is allowed to use fsadm_t's fds.

Since we generally trust the kernel to handle any file handle correctly, we can allow it to use all domains' fds rather than trying to limit it only to specific domains.

Fixes: 1e8688ea6943 ("Don't make kernel_t an unconfined domain")
Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>


zpytela:
Merging, thank you.

PR #1509, opened 2022-12-13
Grant kernel_t certain permissions in the system class

WOnder93:
Grant those that are not systemd/service-related except module_load (which has special semantics and shouldn't be needed anyway).

This fixes a denial via plymouthd (started before the policy was loaded, thus ending up with kernel_t), but it makes sense to allow these also for the kernel just in case.

Fixes: 1e8688ea6943 ("Don't make kernel_t an unconfined domain")
Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>


zpytela:
Merging, thank you.

PR #1514, opened 2022-12-14
Allow syslogd_t to relabel syslogd_var_run_t

5umm3r15:
Resolves: bz#2152823


zpytela:
Resolved in https://github.com/fedora-selinux/selinux-policy/pull/1511 which is a superset.

PR #1517, opened 2022-12-15
Allow stalld to read /sys/kernel/security/lockdown file

5umm3r15:
The last version of stalld is checking if the system is in lockdown mode.

Resolves: rhbz#2140673

### DIFF
--- a/policy/modules/contrib/stalld.te
+++ b/policy/modules/contrib/stalld.te
@@ -43,6 +43,8 @@ domain_use_interactive_fds(stalld_t)
 
 files_read_etc_files(stalld_t)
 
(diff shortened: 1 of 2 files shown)


zpytela:
Does the service need more permissions than reading the reported file?

5umm3r15:
I created and used interface to allow reading security_t files.
Can you review it again please?

zpytela:
At least search for the same type and sysfs_t, too, is needed.

5umm3r15:
@zpytela Fixed

zpytela:
Merging, thank you.

PR #1526, opened 2022-12-17
Allow systemd-time-wait-sync.service to add a watch for /run/systemd/

Fale:
Fixes: rhbz#1971057


zpytela:
Fabio,

While the permission is correct, the type is from the init module so interface needs to be used.

zpytela:
I see it actually has been resolved, although for a bit different use case:
https://github.com/fedora-selinux/selinux-policy/pull/1505
Anyway thank you for your effort.

PR #1530, opened 2022-12-21
Allow stalld to read /sys/kernel/security/lockdown file

zpytela:
The last version of stalld is checking if the system is in lockdown mode.

Add interface to allow read security_t files

Resolves: rhbz#2140673


zpytela:
The build issue is unrelated (module only in f38).

PR #1532, opened 2022-12-21
Allow systemd-resolved watch tmpfs directories

zpytela:


zpytela:
The build failure is unrelated.

PR #1546, opened 2023-01-04
Allow systemd-journal list cgroup directories

zpytela:


WOnder93:
This seems to contain unrelated commit(s)?

zpytela:
Well, they are related in the sense they are reported by the same test suite. Commits are independent.

policy/modules/kernel/kernel.te
```diff
@@ -259,6 +259,7 @@ sid tcp_socket		gen_context(system_u:object_r:unlabeled_t,mls_systemhigh)
 
 allow kernel_t self:capability2 ~mac_override;
 allow kernel_t self:capability ~sys_module;
+allow kernel_t self:key write;
```
WOnder93: I would go one step further and allow the whole `manage_key_perms` set. There is not much benefit in being strict about kernel_t - on the contrary, it might save us some whack-a-mole work later.

WOnder93:
Ah, okay. I got confused by the PR title, which matches the subject of one of the commits.

policy/modules/kernel/kernel.te
zpytela: Done. I also updated the commit message.

WOnder93 (approved):

zpytela:
Thanks for the review and sorry for the confusion.

PR #1552, opened 2023-01-11
Allow upsd to send signal to itself

opoplawski:
This is required for systemctl reload nut-server

type=AVC msg=audit(1673477932.190:38451): avc:  denied  { kill } for  pid=22403 comm="upsd" capability=5  scontext=system_u:system_r:nut_upsd_t:s0 tcontext=system_u:system_r:nut_upsd_t:s0 tclass=capability permissive=0


zpytela:
Merging, thank you.

PR #1553, opened 2023-01-12
Allow wireguard to create udp sockets and read net_conf

5umm3r15:
Allow wireguard to read network config files and create its udp sockets

Resolves: rhbz#2129438
rhbz#2149452


zpytela:
LGTM

PR #1555, opened 2023-01-12
Allow irqbalance to run scripts

5umm3r15:
Resolves: rhbz#2151466


zpytela:
I think this should be discussed with irqbalance maintainers first what are the real scenarios and then decide if is should be allowed for irqbalance_t directly or have a transition to a dedicated domain, possibly implementing a similar solution to keepalived.

PR #1565, opened 2023-01-20
Add policy for nvme-stas

5umm3r15:
Add interface to allow domain to write the dbus pid sock_file.

Resolves: rhbz#2111414

### DIFF
--- a/policy/modules/contrib/nvme_stas.if
+++ b/policy/modules/contrib/nvme_stas.if
@@ -1,0 +1,60 @@
+## <summary>policy for nvme_stas</summary>
+
+########################################
+## <summary>
+##	Execute nvme_stas_exec_t in the nvme_stas domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed to transition.
+## </summary>
+## </param>
+#
+interface(`nvme_stas_domtrans',`
+	gen_require(`
+		type nvme_stas_t, nvme_stas_exec_t;
+	')
+
+	corecmd_search_bin($1)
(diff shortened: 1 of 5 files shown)


policy/modules/contrib/nvme_stas.if
```diff
@@ -0,0 +1,61 @@
+
```
zpytela: Seems to be an unnecessary blank line.

policy/modules/contrib/nvme_stas.te
```diff
@@ -0,0 +1,117 @@
...
+
+corecmd_exec_bin(nvme_stas_t)
+
+dev_read_sysfs(nvme_stas_t)
+domain_use_interactive_fds(nvme_stas_t)
+
+files_getattr_all_files(nvme_stas_t)
+files_read_etc_files(nvme_stas_t)
+
+nvme_stas_dbus_chat(nvme_stas_t)
```
zpytela: This domain really needs dbus chat with itself?

zpytela (changes requested):
Also an entry in modules.conf seems to be missing.

zpytela:
Since the original bz has been closed migrated, I'd rather also have jira reference in the resolves field.

5umm3r15:
thanks for the review, fixed

policy/modules/contrib/nvme_stas.te
zpytela: Interface from the same module should not be used unless there is a good reason; this time it seems just a simple allow rule is needed.

zpytela:
Merging, thank you.

PR #1570, opened 2023-01-24
Allow wg to send msg to kernel, write to syslog and dbus connections

5umm3r15:
Allow wireguard to write to the kernel messages device, send messages to kernel unix datagram sockets,
create connections to the system DBUS,
write to the syslog pid sock_files.

Resolves: rhbz#2149452


zpytela:
LGTM

PR #1574, opened 2023-01-26
Allow kernel_t to manage and execute all files

WOnder93:
This is needed to get early overlay mounts to work - see the comment and
this bug:
https://bugzilla.redhat.com/show_bug.cgi?id=2154428
    
Allowing to manage all files may not be strictly necessary, but not
allowing it may bite us later and there is not much point in confining
kernel's file operations beyond the helper execution anyway.

---

Note that this will make the current `sekinux-policy/kernel-confined-exec` test fail - https://src.fedoraproject.org/tests/selinux/pull-request/358 needs to be merged before bringing this into Fedora.

### DIFF
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -1800,6 +1800,32 @@ interface(`files_manage_all_files',`
 
 ########################################
 ## <summary>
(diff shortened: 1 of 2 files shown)


zpytela:
Thank you, I haven't found any problem so far.

PR #1577, opened 2023-01-27
Allow init read all non-security socket files

zpytela:
This permission is required when LoadCredential is set to point to a socket file for providing credentials to a service:

LoadCredential=id:/run/wg-credentials.socket

Systemd in its read_full_file_full() function checks exit code of xfopenat() and in case of -ENXIO connects to the AF_UNIX socket.

The files_read_non_security_sock_files() interface was added.

type=AVC msg=audit(1672210323.355:1362): avc:  denied  { read } for  pid=4490 comm="(sd-mkdcreds)" name="wg-credentials.socket" dev="tmpfs" ino=1033 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=sock_file permissive=0


umglurf:
Hi, I installed the rpm from the build and got one step further, but there is now a new block for connect
```
type=AVC msg=audit(1674992652.819:336): avc:  denied  { connectto } for  pid=1225 comm="(sd-mkdcreds)" path="/run/wg-credentials.socket" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=unix_stream_socket permissive=0
```

zpytela:
@umglurf, is this a custom service, not in Fedora?

umglurf:
> @umglurf, is this a custom service, not in Fedora?

Hi, that is correct, it's a service I've written, but I think this would be the same issue with services in Fedora. I can create a small demo service if that would be helpful?

zpytela:
That'd be nice. If a similar AVC appears, the following module should help:
```
cat local_init_connect_domain.cil
(allow init_t domain (unix_stream_socket (connectto))
semodule -i local_init_connect_domain.cil
```

umglurf:
> That'd be nice. 

I've attached a simple credential server that just always returns `foobar`, it requires python3-systemd, otherwise nothing but python3.
[test-socket.tar.gz](https://github.com/fedora-selinux/selinux-policy/files/10538150/test-socket.tar.gz)



> If a similar AVC appears, the following module should help:
> 
> ```
> cat local_init_connect_domain.cil
> (allow init_t domain (unix_stream_socket (connectto))
> semodule -i local_init_connect_domain.cil
> ```

That worked.

umglurf:
@zpytela any update on this PR?

PR #1580, opened 2023-01-30
Update to socket_class_set

zpytela:


zpytela:
I believe removing "socket" can be done in 2 steps, with removing all references to the class and deprecating interfaces coming later.

zpytela:
@WOnder93 Please review. A good thing to say about this pr is that the system boots.

WOnder93:
The commit message of the second commit ([Remove socket from socket_class_set](https://github.com/fedora-selinux/selinux-policy/pull/1580/commits/672b12cd54cdba521e2a16c723bc3cebdc4943fb)) should be changed, because `socket` is not being removed from the `socket_class_set` macro, just from the `unconfined_domain_type` allow rule.

Otherwise LGTM.

PR #1583, opened 2023-01-31
Add initial policy for the /usr/sbin/request-key helper

WOnder93:
The kernel is hard-coded to call /sbin/request-key to handle requests for instantiating keys that are not found in the existing keyrings.

Thus, we need to add a domain for this helper and set up a transition into that domain from kernel_t.

request-key dispatches the key requests to further helper programs based on the configuration in /etc/request-key.d/*.conf and /etc/request-key.conf. Currently, the only known used dispatcher is /usr/sbin/nfsidmap, which is set up by the nfs-utils package to handle requests for the 'id_resolver' key type. This patch adds the minimal policy for this helper that is needed for an NFS mount to succeed.

Policy for other request-key helper programs may need to be added in the future. An optional mechanism to allow any possible configuration (e.g. by setting up a transition over any file to unconfined_service_t) may be also desired. For now let's at least make the one known use case work.

### DIFF
--- a/policy/modules.conf
+++ b/policy/modules.conf
@@ -3078,3 +3078,10 @@ rhcd = module
 # wireguard
 #
 wireguard = module
+
+# Layer: contrib
+# Module: keyutils
+#
+#  keyutils - Linux Key Management Utilities
+#
+keyutils = module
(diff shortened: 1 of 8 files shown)


zpytela:
Merging, thank you.

cgwalters:
Possible fallout related to this in https://github.com/coreos/fedora-coreos-tracker/issues/1447#issuecomment-1482006010

PR #1594, opened 2023-02-07
Label systemd-journald feature LogNamespace

5umm3r15:
Systemd-journald introduced feature `LogNamespace` to isolate certain logs. These names space are created from a template service `systemd-journald@.service`. For example to activate a LogNamespace `testcase` we use: `# systemctl start systemd-journald@testcase.service`

LogNamespaces were created with wrong SELinux context init_var_run_t, instead of syslogd_var_run_t and devlog_t.

Resolves: rhbz#2124797


zpytela:
LGTM

PR #1595, opened 2023-02-08
Add label virt_qemu_ga_tmp_t to /tmp/qemu-ga/ and /var/tmp/qemu-ga/

5umm3r15:
The open on tmp files used to be granted by mistake to all domains and by fixing this qemu-ga lost the ability to open these files by default.

Resolves: rhbz#2152121


5umm3r15:
bfu@redhat.com zpytela@redhat.com

I labeled directories /var/tmp/qemu-ga and /tmp/qemu-ga.
After creating these directories, it is necessary to use restorecon, so they get right context.

5umm3r15:
Closing this MR after discussion with qemu-ga developers.

PR #1598, opened 2023-02-10
Add policy for bootupd

5umm3r15:
Bootupd is a small socket activated program that takes care of updating the bootloader.

resolves: rhbz#2044508


zpytela:
Merging, thank you.

cgwalters:
I didn't test this but I will re-express my fear that this is highly likely to break bootupd and I am not seeing significant value from a custom policy for it.

5umm3r15:
@cgwalters it should be tested together with coreos installer and rust-afterburn. I created [copr build](https://copr.fedorainfracloud.org/coprs/nknazeko/coreos_installer-afterburn-bootupd.selinux/) for these three selinux modules and sent it to developers for testing 4 months ago, and there were no SELinux denials.

PR #1607, opened 2023-02-15
Allow staff_t getattr init pid chr & blk files and read krb5

5umm3r15:
Add interface to get the attributes of block and character device nodes in the /run/systemd directory.

Resolves: rhbz#2112729


zpytela:
LGTM

PR #1613, opened 2023-02-21
contrib/virt: Optional policies for passt lifecycle handling in libvirt

sbrivio-rh:
```
These optional policies rely on new interfaces which are now exported
by passt-selinux0^20230310.g70c0765-1. Interface compatibility
doesn't need to be preserved as libvirt currently isn't able to start
passt (unless SELinux is in permissive mode), and was never able to.

In detail:

- passt_domtrans() implements the domain transition to passt_t and
  allows to execute passt(1)

- passt_kill() allows libvirtd to terminate the process, which is
  only needed in case the guest didn't start (otherwise, passt will
  terminate by itself)

- passt_socket_dir() allows passt to add the socket directory entry
  to the paths required for root (virt_var_run_t) and non-root
  (user_tmp_t) operation

- passt_socket_create() allows passt to create the socket file itself

- passt_socket_use() allows qemu (from its virt_domain) to open the
  interface to passt, which is a UNIX domain socket, with contexts
  supporting guest domain root operation (path in qemu_var_run_t
  context) and non-root (user_tmp_t)

- path_socket_delete() allows libvirtd to delete the socket file

- passt_pidfile_dir() allows passt to add its PID file directory
  entry (qemu_var_run_t for guest domains started by root,
  user_tmp_t otherwise)

- passt_pidfile_write() allows passt to write its PID file

- passt_pidfile_delete() allows libvirtd to delete the PID file

- passt_logfile_dir() allows passt to add the directory entry for
  its log file

- passt_logfile_use() allows passt to create, write and manipulate
  the log file itself. Note that passt can't delete its own log
  file, as it has no access to the filesystem after initalisation
  and sandboxing.
```

### DIFF
--- a/policy/modules/contrib/passt.if
+++ b/policy/modules/contrib/passt.if
@@ -1,0 +1,100 @@
+interface(`passt_domtrans',`
+	gen_require(`
+		type passt_t, passt_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, passt_exec_t, passt_t)
+')
+
+interface(`passt_socket_dir',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow passt_t $1:dir add_entry_dir_perms;
+')
+
+interface(`passt_socket_create',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow passt_t $1:sock_file create;
+')
+
+interface(`passt_socket_use',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 passt_t:unix_stream_socket connectto;
+	allow $1 $2:sock_file { read write };
+	allow passt_t $2:sock_file { read write };
+')
+
+interface(`passt_socket_delete',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 $2:sock_file unlink;
+')
+
+interface(`passt_logfile_dir',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow passt_t $1:dir add_entry_dir_perms;
+')
+
+interface(`passt_logfile_use',`
+	gen_require(`
+		type passt_t;
+	')
+
+	logging_log_file($1);
+	allow passt_t $1:file { create open read write };
+')
+
+interface(`passt_pidfile_dir',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow passt_t $1:dir add_entry_dir_perms;
+')
+
+interface(`passt_pidfile_write',`
+	gen_require(`
+		type passt_t;
+	')
+
+	files_pid_file($1);
+	allow passt_t $1:file { create open write };
+')
+
+interface(`passt_pidfile_read',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 $2:file { open read };
+')
+
+interface(`passt_pidfile_delete',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 $2:file unlink;
+')
+
+interface(`passt_kill',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 passt_t:process { signal sigkill };
+')
--- a/policy/modules/contrib/virt.te
+++ b/policy/modules/contrib/virt.te
@@ -719,6 +719,38 @@ optional_policy(`
 ')
 
 optional_policy(`
+	passt_socket_dir(user_tmp_t)
+	passt_socket_dir(virt_var_run_t)
+
+	passt_socket_create(user_tmp_t)
+	passt_socket_create(virt_var_run_t)
+
+	passt_socket_use(virt_domain, user_tmp_t)
+	passt_socket_use(virt_domain, virt_var_run_t)
+
+	passt_socket_delete(virtd_t, user_tmp_t)
+	passt_socket_delete(virtd_t, virt_var_run_t)
+
+	passt_pidfile_dir(user_tmp_t)
+	passt_pidfile_dir(qemu_var_run_t)
+
+	passt_pidfile_write(user_tmp_t)
+	passt_pidfile_write(virt_var_run_t)
+
+	passt_pidfile_delete(virtd_t, user_tmp_t)
+	passt_pidfile_delete(virtd_t, virt_var_run_t)
+
+	passt_logfile_dir(user_tmp_t)
+	passt_logfile_dir(qemu_var_run_t)
+
+	passt_logfile_use(user_tmp_t)
+	passt_logfile_use(virt_var_run_t)
+
+	passt_domtrans(virtd_t)
+	passt_kill(virtd_t)
+')
+
+optional_policy(`
 	firewalld_dbus_chat(virtd_t)
 ')
 


zpytela:
@sbrivio-rh there are some rules dealing with `user_tmp_t`, which files are they? In general, it should rather be a private type than a shared one.

I will continue with the PR review.

sbrivio-rh:
> @sbrivio-rh there are some rules dealing with user_tmp_t, which files are they? In general, it should rather be a private type than a shared one.

It's in case libvirtd(8) (or virtqemud(8)) asks to write the PID file or have passt(1) set up the socket in `/var/run/<UID>/libvirt/qemu/*`. I guess this could be made more specific in `virt.fc`, but at the same time we can't collide with `user_tmp_t`, so I couldn't really find a better solution. And anyway I guess it's out of scope for this changeset.

> I will continue with the PR review.

Thanks!

zpytela:
If you mean the `/var/run/user/%{USERID}` structure, there are already various private types defined for domains.
What makes it difficult is you cannot just use types from other modules.
Stay tuned.

policy/modules/contrib/virt.te
```diff
@@ -718,6 +718,20 @@ optional_policy(`
...
 
+optional_policy(`
+	passt_socket(virt_domain, user_tmp_t)
+	passt_socket(virt_domain, qemu_var_run_t)
+
+	passt_pidfile(virtd_t, user_tmp_t)
+	passt_pidfile(virtd_t, virt_var_run_t)
+	passt_pidfile(virtd_t, qemu_var_run_t)
+
+	passt_logfile(user_tmp_t)
```
zpytela: What is the purpose of this line?
user_tmp_t is a generic type for runtime and temporary files.

policy/modules/contrib/virt.te
```diff
@@ -718,6 +718,20 @@ optional_policy(`
 	dnsmasq_manage_pid_files(virtd_t)
 ')
 
+optional_policy(`
+	passt_socket(virt_domain, user_tmp_t)
+	passt_socket(virt_domain, qemu_var_run_t)
+
+	passt_pidfile(virtd_t, user_tmp_t)
+	passt_pidfile(virtd_t, virt_var_run_t)
+	passt_pidfile(virtd_t, qemu_var_run_t)
```
zpytela: These two lines effectively, among other things, allow `virtd_t` manage virt/qemu runtime files unconditionally.

policy/modules/contrib/virt.te
```diff
@@ -718,6 +718,20 @@ optional_policy(`
 	dnsmasq_manage_pid_files(virtd_t)
 ')
 
+optional_policy(`
+	passt_socket(virt_domain, user_tmp_t)
+	passt_socket(virt_domain, qemu_var_run_t)
+
+	passt_pidfile(virtd_t, user_tmp_t)
```
zpytela: Accessing a type outside this module.
Also see below.

policy/modules/contrib/passt.if
```diff
@@ -0,0 +1,67 @@
...
+
+interface(`passt_socket',`
+	gen_require(`
+		type passt_t;
+	')
+
+	allow $1 $2:sock_file write;
+	allow $1 passt_t:unix_stream_socket connectto;
+
+	allow passt_t $2:sock_file { create read write unlink };
```
zpytela: I don't understand what this interface is expected to support.

policy/modules/contrib/passt.if
```diff
@@ -0,0 +1,67 @@
...
+# Copyright (c) 2022 Red Hat GmbH
+# Author: Stefano Brivio <sbrivio@redhat.com>
+
+interface(`passt_read_data',`
+	gen_require(`
+		type passt_data_t;
+	')
+	allow $1 passt_t:dir { search add_name };
+	allow $1 passt_t:file { open read getattr };
+')
```
zpytela: Is `passt_t` as the target domain a typo?
It also allows add_name which does not sound like `read`.

policy/modules/contrib/passt.if
```diff
@@ -0,0 +1,67 @@
...
+	gen_require(`
+		type passt_t;
+	')
+
+	logging_log_file($1);
+	allow passt_t $1:dir { search write add_name };
+	allow passt_t $1:file { create open read write };
+')
+
+interface(`passt_pidfile',`
```
zpytela: The interface name is not self-explanatory; maybe it helped to split it to passt_create_pid_files etc.?

zpytela (changes requested):
Please also use object permission sets and patterns wherever possible, e.g. for managing files or stream connect. All interfaces should be wrapped by ifdefs, we have `ifndefy.py` in the rpm sources for that.

policy/modules/contrib/virt.te
sbrivio-rh: This is to let users configure passt to write a log file under `/var/run/<UID>/libvirt`, for current lack of a better type. It's a temporary file in some sense, it's expected to be used for debugging rather than for auditing.

policy/modules/contrib/virt.te
sbrivio-rh: I'm actually not very concerned by it, after all `virtd_t` also starts QEMU with arbitrary options, but I see it's not ideal. Anyway, lacking a specific type I don't see a better solution. What do you propose?

policy/modules/contrib/virt.te
sbrivio-rh: ...which one?

policy/modules/contrib/passt.if
sbrivio-rh: virt_domain (QEMU) needs to be able to write to the socket, and to connect to the passt daemon running as `passt_t`.

passt needs to create that socket, read and write from it, and unlink it.

policy/modules/contrib/passt.if
sbrivio-rh: This interface (`passt_read_data`) should be dropped, and is not used by any policy. I just imported the current interface here as I don't see a proper way to do so otherwise, and it seems to be uncommon for `*-selinux` packages to install their interfaces to `/usr/share/selinux/devel/include/contrib/`

policy/modules/contrib/passt.if
sbrivio-rh: Sure, it could be done. So you're suggesting that `allow $1 $2:file { open read unlink }` would be provided by a separate `passt_handle_pid_files`...?

andreabolognani:
Not a SELinux maintainer (or expert), just sharing some things that I've noticed while testing the proposed patches for passt support in libvirt.

First of all, the file added by this PR (`passt.if`) only partially covers the needs of passt. In particular, without the matching `passt.fc`, `/usr/bin/passt` doesn't get the `passt_exec_t` label assigned to it and things don't work.

Additionally, I've found that the lack of at least a `<summary>` entry at the top of the module results in an error at build time. There are also a bunch of warnings about missing XML for all the interfaces, but those at least don't prevent the build from completing.

The module should also be listed in `policy/modules.conf` to be picked up.

I have added all the changes mentioned above to my [f37-passt-fixes](https://github.com/andreabolognani/fedora-selinux-policy/commits/f37-passt-fixes) branch. It's based on the `f37` branch instead of `rawhide` because that's what I'm testing on.

I've actually tested this by building Fedora RPM packages, simply replacing the upstream tarball with one generated from my repo. In addition to that, I've also needed to list the module in `modules-targeted-contrib.conf` to get it to show up in the RPM.

After doing all of the above, things seem to be working great :) libvirt is able to start passt with the expected context.

One last thing: the module added as part of this PR is AGPL-3.0+ licensed, which I'm not sure is compatible with the license for this project (GPL-2.0). It might be necessary to relicense it.

I hope at least some of this was helpful :)

andreabolognani:
Forgot to mention a couple of things. I've only looked at the passt module, because that's what I'm interested in testing right now, but I assume the pasta module will need to also be dealt with in the same way.

Once the modules are merged here, the `passt-selinux` package in Fedora should cease to exist. We should probably declare package conflicts of some sort, to ensure that people who have `passt-selinux` installed today are smoothly migrated to the integrated policy. I'm not 100% sure how that would be handled in Fedora.

policy/modules/contrib/virt.te
andreabolognani: Not specifically related to the logfile, but I'm wondering... When running in `qemu:///system` mode, both the pidfile and the socket will be created under `/run/libvirt/qemu/passt/` (`qemu_var_run_t`). When running in `qemu:///session` mode, `virtqemud` will be running unconfined, so I would expect passt to also follow that. In other words, I think the `user_tmp_t` rules are unnecessary.

Now speaking of the logfile in particular :) The current XML schema allows the user to pick an arbitrary path where the file should be stored, but that's obviously problematic because, even with the rule you've added here, there are going to be a number of technically valid paths that are not going to work in practice. So I wonder, does it even make sense to make to allow the user to pick an arbitrary path as opposed to generating a predictable one?

Prior art for predictable locations: swtpm. Counter-example: serial console logging, where arbitrary paths are allowed.

sbrivio-rh:
> First of all, the file added by this PR (`passt.if`) only partially covers the needs of passt. In particular, without the matching `passt.fc`, `/usr/bin/passt` doesn't get the `passt_exec_t` label assigned to it and things don't work.

Well, yes, but that's something you get from `passt-selinux` (I tested these changes, even though I didn't try building the `selinux-policy` RPM as you described below), and packaging SELinux policy modules as policy module, with a subpackage as I did, is expressly described in [packaging guidelines](https://fedoraproject.org/wiki/SELinux_Policy_Modules_Packaging_Draft#Separate_Subpackage?).

The most relevant aspect for me here is:

> Policy modules allow an SELinux policy for an application to be developed alongside and packaged with the application itself, rather than having to integrate the policy with the monolithic base policy

Note that the policies introduced here are `optional`, and include `gen_require(`type passt_t;')`, so if `passt_t` is not present, they won't be expanded. See also https://selinuxproject.org/page/NB_RefPolicy#gen_require_Macro and https://selinuxproject.org/page/NB_RefPolicy#policy_module_Macro.

> Additionally, I've found that the lack of at least a `<summary>` entry at the top of the module results in an error at build time. There are also a bunch of warnings about missing XML for all the interfaces, but those at least don't prevent the build from completing.

Okay, thanks, I can steal this from your branch.

> The module should also be listed in `policy/modules.conf` to be picked up.

And this too.

> One last thing: the module added as part of this PR is AGPL-3.0+ licensed, which I'm not sure is compatible with the license for this project (GPL-2.0). It might be necessary to relicense it.

It's not compatible for relicensing, but it is indeed for joint distribution. In any case, I don't think it matters, because:

> Once the modules are merged here, the `passt-selinux` package in Fedora should cease to exist.

I don't think we should go this way, and as Fedora maintainer for `passt`, I strongly advise against it. Distributing `passt-selinux` allows the SELinux policy to closely follow the development of passt itself, and that's the reason why we have modules and interfaces after all.

I'll wait for Zdeněk to follow up on his comments and integrate the changes you provided meanwhile.

sbrivio-rh:
> Forgot to mention a couple of things. I've only looked at the passt module, because that's what I'm interested in testing right now, but I assume the pasta module will need to also be dealt with in the same way.

Not necessarily: the `container-selinux` repository also plays a role there, and things currently work with Podman. Stand-alone usage is, of course, already covered.

policy/modules/contrib/virt.te
sbrivio-rh: > When running in `qemu:///session` mode, `virtqemud` will be running unconfined

Not in my tests, and `/usr/sbin/virtqemud` is defined as `virtd_exec_t` by `virt.fc`.

> so I would expect passt to also follow that

Even if that was the case, we clearly don't want to let passt run unconfined.

> So I wonder, does it even make sense to make to allow the user to pick an arbitrary path as opposed to generating a predictable one?

Probably not, but here I'm reflecting the current status of the libvirt integration.

> Prior art for predictable locations: swtpm.

This looks preferable to me because it avoids tricking the user into set a path that can't actually be used.

lainestump: > > So I wonder, does it even make sense to make to allow the user to pick an arbitrary path as opposed to generating a predictable one?
> 
> Probably not, but here I'm reflecting the current status of the libvirt integration.

At the time I wrote the libvirt code, I was not aware that passt had selinux policies to limit where the logfile could be written :-/

> > Prior art for predictable locations: swtpm.
> 
> This looks preferable to me because it avoids tricking the user into set a path that can't actually be used.

Which aspect of swtpm are you talking about here? (I think we're straying a bit from discussing the job at hand in this PR, BTW...)

lainestump:
> Not a SELinux maintainer (or expert)

Likewise :-)

> Additionally, I've found that the lack of at least a `<summary>` entry at the top of the module results in an error at build time. There are also a bunch of warnings about missing XML for all the interfaces, but those at least don't prevent the build from completing.

I've added the two patches in this PR (with no modification) to the Fedora selinux-policy.spec file and "fedpkg local" completes with no error. Strange...

> 
> The module should also be listed in `policy/modules.conf` to be picked up.

I don't think this is true. It looks like modules.conf should only list those modules that are completely implemented within the selinux-policy package. For example, a couple years ago the cockpit module was moved out of selinux-policy into a sub-package of cockpit itself, and part of this process was to remove the reference to cockpit from modules.conf. (It also moved all cockpit.* files from selinux-policy's contrib directory over to the cockpit source tree, although it kept a copy of cockpit.if in selinux-policy - this seems to parallel what Stefano has done here - he put a copy of passt.if in selinux-policy contrib directory, apparently to eliminate "undefined" errors)


> I have added all the changes mentioned above to my [f37-passt-fixes](https://github.com/andreabolognani/fedora-selinux-policy/commits/f37-passt-fixes) branch. It's based on the `f37` branch instead of `rawhide` because that's what I'm testing on.
> 
> I've actually tested this by building Fedora RPM packages, simply replacing the upstream tarball with one generated from my repo. In addition to that, I've also needed to list the module in `modules-targeted-contrib.conf` to get it to show up in the RPM.

Which module are you talking about here? The passt module? That shouldn't be included in the selinux-policy package - it gets installed with the passt-selinux package, which is a subpackage of passt. Everything that's done in this PR is just to update the virt module to add policies related to libvirt running passt. (The only reason passt.if is included in this PR is because it's needed in order to resolve references to passt_* that were added to virt.te.)


> 
> After doing all of the above, things seem to be working great :) libvirt is able to start passt with the expected context.

What I've done is:

1) build libvirt with the patches I posted upstream yesterday
2) build/install the passt.pp module *from the passt source tree* (I did it manually because a new passt-selinux package wasn't yet available)
3) add the 2 patches of this PR to the F37 selinux-policy.spec file, run "fedpkg local", and dnf reinstall the resulting selinux-policy package.

That resulted in passt being exec'ed with the right label, and being able to create its pidfile, listen on ports > 1024, and write to a logfile (but only in /var/run/<UID>/blah.log (UID being 107 ("qemu") in the case of qemu:///system)

policy/modules/contrib/virt.te
andreabolognani: >> When running in `qemu:///session` mode, `virtqemud` will be running unconfined
>>
> Not in my tests, and `/usr/sbin/virtqemud` is defined as `virtd_exec_t` by `virt.fc`.
>
>> so I would expect passt to also follow that
>>
> Even if that was the case, we clearly don't want to let passt run unconfined.

Mh, here `virtqemud` (in `qemu:///session` mode, as autostarted by running `virsh -c qemu:///session list` from a terminal) definitely runs unconfined. But QEMU itself runs as `svirt_t` and passt gets `passt_t`, just like in `qemu:///system` mode. I'm not sure how that's happening, but it looks like indeed we need those `user_tmp_t` rules.

With that established, let's leave the discussion about logfile to another day and forum :)

andreabolognani:
> It looks like modules.conf should only list those modules that are completely implemented within the selinux-policy package. For example, a couple years ago the cockpit module was moved out of selinux-policy into a sub-package of cockpit itself, and part of this process was to remove the reference to cockpit from modules.conf. (It also moved all cockpit.* files from selinux-policy's contrib directory over to the cockpit source tree, although it kept a copy of cockpit.if in selinux-policy - this seems to parallel what Stefano has done here - he put a copy of passt.if in selinux-policy contrib directory, apparently to eliminate "undefined" errors)

My impression is that you can have it one of two ways: either the module is part of fedora-selinux-policy, or you ship it in a separate foo-policy package.

However, the fact that the policy for libvirt is part of fedora-selinux-policy means that the one for passt also has to: a hypothetical libvirt-selinux package could easily depend on passt-selinux, but fedora-selinux-policy obviously can't.

The cockpit example goes directly against this mental model, so I might just have gotten it all wrong O:-) In my defense, see the earlier disclaimer about not being a SELinux expert ;)

Note that the commit that dropped the integrated cockpit module also changed the various interfaces:

```diff
-interface(`cockpit_ws_domtrans',`
-       gen_require(`
-               type cockpit_ws_t, cockpit_ws_exec_t;
+ifndef(`cockpit_ws_domtrans',`
+       interface(`cockpit_ws_domtrans',`
+               gen_require(`
+                       type cockpit_ws_t, cockpit_ws_exec_t;
+               ')
+
+               corecmd_search_bin($1)
+               domtrans_pattern($1, cockpit_ws_exec_t, cockpit_ws_t)
        ')
-
-       corecmd_search_bin($1)
-       domtrans_pattern($1, cockpit_ws_exec_t, cockpit_ws_t)
 ')
```

I think the `ifndef()` dance might be intended to prevent the definitions in cockpit-selinux from being overwritten by those in fedora-selinux-policy? If so, we might want to do the same for passt.

I will now leave it to the maintainers. Sorry for the noise.

zpytela:
I'd like to make one thing clear: Goal of this commit is to add passt support to virt. For that, copies of its interfaces are needed when policy is being built. The rest of passt policy is provided by the passt SELinux module.

sbrivio-rh:
> However, the fact that the policy for libvirt is part of fedora-selinux-policy means that the one for passt also has to: a hypothetical libvirt-selinux package could easily depend on passt-selinux, but fedora-selinux-policy obviously can't.

No, it doesn't need to (otherwise "external" or "optional" interfaces would make little sense, right?). I just need to install `passt.if` from `passt-selinux` in the right place and fix things up. Other packages with external interfaces do that as well (see `snapd-selinux`), I didn't know about that. I'll get to this in a bit and update this merge request.

By the way, somewhat on the topic:

> Mh, here `virtqemud` (in `qemu:///session` mode, as autostarted by running `virsh -c qemu:///session list` from a terminal) definitely runs unconfined.

I see this as a sign that libvirt could benefit from maintaining its own policy package as well. Per se, that's "just" a bug, but the fact it went unnoticed makes me think...

Anyway, I'll update this in a couple of hours.

policy/modules/contrib/passt.if
zpytela: From long-term-maintaining perspective it is good to follow how it works in other modules. Certainly unless there is a good reason to make it different way.

policy/modules/contrib/virt.te
zpytela: user_tmp_t

policy/modules/contrib/virt.te
zpytela: Problem is I still don't understand which rules are needed and why. Here it just feels incorrect that from inside a different module there are allow rules for 2 types from virt. Maybe a boolean would help.

policy/modules/contrib/virt.te
sbrivio-rh: ...what should I use instead? When the guest domain is started by a regular user, the PID file needs to be written inside `/run/user/<UID>/libvirt/qemu/run/passt/` (e.g. `/run/user/1001/libvirt/qemu/run/passt/2-Fedora_37-net0-passt.pid`), and `user_tmp_t` covers that.

I tried to define a different file context for it, but it clashes with `user_tmp_t` as I load the policy module.

policy/modules/contrib/virt.te
sbrivio-rh: Trying to rephrase this:

> passt_pidfile() allows passt to write PID files, and libvirtd to read them, whether started as root (for which we need virt_var_run_t and qemu_var_run_t contexts) or not (user_tmp_t)

point by point, per rule:

- the first rule, `passt_pidfile(virtd_t, user_tmp_t)`, covers the case where the guest domain is started by an unprivileged user, and is needed to:
  - allow `virtd_t` to open, read, and unlink the PID file under ` /run/user/<UID>/libvirt/qemu/run/passt/`
  -  allow passt to check entries under ` /run/user/<UID>/libvirt/qemu/run/passt/`, and to add a new directory entry
  -  allow passt to write PID files under ` /run/user/<UID>/libvirt/qemu/run/passt/`

- the second rule, `passt_pidfile(virtd_t, virt_var_run_t)`, covers the case where the guest domain is started as root, and is needed to:
  - allow `virtd_t` (regardless of MLS) to open, read, and unlink the PID file under ` /var/run/libvirt/qemu/run/passt`
  -  allow passt to read directory entries under ` /run/user/<UID>/libvirt/`, and to add a new directory entry if needed
 
- the third rule, `passt_pidfile(virtd_t, qemu_var_run_t)`, also covers the case where the guest domain is started as root, and is needed to:
  - allow `virtd_t` to open, read, and unlink the PID file under ` /var/run/libvirt/qemu/run/passt`
  -  allow passt (with a specific, per-VM MLS category) to read directory entries under `/var/run/libvirt/qemu`, and to add a new directory entry
  -  allow passt (with a specific, per-VM MLS category) to write PID files under `/var/run/libvirt/qemu`

Indeed, the second rule isn't "entirely needed", and could be avoided by reworking types defined in `virt.fc`, which is in my opinion out of scope here.

andreabolognani:
> I'd like to make one thing clear: Goal of this commit is to add passt support to virt. For that, copies of its interfaces are needed when policy is being built. The rest of passt policy is provided by the passt SELinux module.

Okay, if this is the accepted way to do things then we're already on the right track :) Apologies making noise because of my misunderstanding of the best practices.

I've also verified that the `<summary>` and `modules.conf` changes are indeed unnecessary when only adding `passt.if` to this repo, and at least for the latter it would probably be actively wrong to include it.

> I see this as a sign that libvirt could benefit from maintaining its own policy package as well. Per se, that's "just" a bug, but the fact it went unnoticed makes me think...

AFAIK that has been discussed several times in the past, and I have no idea if there's anything genuinely blocking such a change other than lack of manpower and inertia. But I'm also not convinced that the issue I've described (assuming it actually turns out to be one) would have been noticed and addressed any earlier if libvirt maintained its own policy...

sbrivio-rh:
The new version plays along nicely with the new SELinux policy package `passt-selinux-0^20230310.g70c0765-1.fc39` (`sudo dnf upgrade --refresh --advisory=FEDORA-2023-a72a4c05e2`). I dropped the commit introducing `passt.if` from this merge request, split the interfaces in smaller parts and used patterns wherever possible.

sbrivio-rh:
...except that @lainestump noticed that, with this, the `fedora-selinux` RPM package doesn't build anymore. Finally I realised that the Makefile used there is *not* `/usr/share/selinux/devel/Makefile`, which is the one I use to build and test policies.

Adding back the interface file, commit 3a55e3a69bb3 explicitly supports this usage, calling those "standalone interface files".

Note that I didn't wrap any interface with ifdefs or ifndefs, as I don't see any such directive used elsewhere that seems to fit here.

berrange:
> I'd like to make one thing clear: Goal of this commit is to add passt support to virt. For that, copies of its interfaces are needed when policy is being built. The rest of passt policy is provided by the passt SELinux module.

This feels really awkward as a requirement, as it creates a bi-directional dependency between passt and selinux-policy packages.  If the passt-selinux package ever wants to change one of the .if files it has, then this change needs to be replicated in selinux-policy too, which somewhat defeats the purpose of having policy split off into the packages.



> >    I see this as a sign that libvirt could benefit from maintaining its own policy package as well. Per se, that's "just" a bug, but the fact it went unnoticed makes me think...
>
> AFAIK that has been discussed several times in the past, and I have no idea if there's anything genuinely blocking such a change other than lack of manpower and inertia. 

It was proposed, but when reviewing it, I discovered that the policy that was being proposed for libvirt.git was rather different from the policy that existed here in selinux-policy, which would have created compatibility problems. Part of the problem is that virt.te in here does too much, some not related to libvirt and should be split off into separate modules. I requested that virt.te in the selinux-policy be updated to match what was being proposed for libvirt.git, but that hasn't happened AFAIK.
 
This problem with .if files is something that was never raised at the time either. There are a very large number of modules which call virt policy interfaces, so it seems like we'd also be forced to keep a virt.if duplicated into selinux-policy which is rather unappealing. Many other things need to be split off out of selinux-policy package to avoid this problem it seems.

vmojzis:
> > AFAIK that has been discussed several times in the past, and I have no idea if there's anything genuinely blocking such a change other than lack of manpower and inertia.
> 
> It was proposed, but when reviewing it, I discovered that the policy that was being proposed for libvirt.git was rather different from the policy that existed here in selinux-policy, which would have created compatibility problems. Part of the problem is that virt.te in here does too much, some not related to libvirt and should be split off into separate modules. I requested that virt.te in the selinux-policy be updated to match what was being proposed for libvirt.git, but that hasn't happened AFAIK.

Yes, the policy was rewritten to accommodate the split daemon configuration so it is quite different from the one currently in selinux-policy-*.
There is no problem introducing the new libvirt policy via selinux-policy-* packages first. In fact I believe we agreed on just that in discussion about the policy on libvir-list (here is a PR for that https://github.com/fedora-selinux/selinux-policy/pull/492). But the policy needs to be properly tested by someone other than policy maintainers first.
I'm sorry if we did not communicate that clearly.
> 
> This problem with .if files is something that was never raised at the time either. There are a very large number of modules which call virt policy interfaces, so it seems like we'd also be forced to keep a virt.if duplicated into selinux-policy which is rather unappealing. Many other things need to be split off out of selinux-policy package to avoid this problem it seems.

Yes, a copy of virt.if needs to stay in selinux-policy package. There is no going around this without adding build dependencies for selinux-policy. To avoid any issues with duplicate definitions of interfaces, all interfaces in the copy of virt.if will be enclosed in "ifndef" statements (that way your versions of interfaces are used whenever available and the copy serves as a fallback).

andreabolognani:
It looks like passt support has been already taken care of with #1629 and #1631. Can this PR be closed now?

zpytela:
> It looks like passt support has been already taken care of with #1629 and #1631. Can this PR be closed now?

Correct, closing.

PR #1617, opened 2023-02-28
Confine gnome-initial-setup

zpytela:


zpytela:
container-selinux was updated yesterday and needs another update: https://github.com/containers/container-selinux/pull/206

PR #1618, opened 2023-03-01
Allow iptables read all tmpfiles

5umm3r15:
From the files_read_inherited_tmp_files interface was removed the open permission, that allowed iptables to open all tmp files.

On a Kubernetes there is a kubelet agent on each node. In order to create a Pod, kubelet is calling the CRI (e.g. cri-o) which in turn calls CNI plugins to configure the networking part. The CNI plugin, in turn, just calls `nft` to setup nftables rules. When doing so, it uses a temporary file to place there the JSON config and later loads it.

Resolves: rhbz#2172090


zpytela:
Is a private tmp type created by a CNI plugin an option?

PR #1629, opened 2023-03-24
Passt-virt support

zpytela:


5umm3r15:
LGTM

vmojzis:
LGTM, I only have 2 nitpicks:

- passt_stub(svirt_t) should probably be "passt_stub()", not to confuse anyone (although functionally there is no difference and we'll probably clean it up soon)  https://github.com/fedora-selinux/selinux-policy/pull/1629/files#diff-dd97425bec2d1c7b5a712e54ea068e5f480808ee9db6e093ce808f2de942df70R1067

- Is there a reason for nested optional_policy blocks as opposed to multiple separate ones? Again, in this case there is probably no functional difference since the rules all impact passt_t, but do we know if there is a difference in behavior when this approach is used with disparate interfaces? https://github.com/fedora-selinux/selinux-policy/pull/1629/files#diff-dd97425bec2d1c7b5a712e54ea068e5f480808ee9db6e093ce808f2de942df70R1057

zpytela:
Thanks for checking, you are right.
1. Checking the policy content, I found out most of the stub interfaces usage contain the domain.
2. One nested block should be dropped in the near future, the only purpose was to have it somehow visually separated; independent block would work as well. The other containing userdom_ depends on the passt usage and is not needed without it.

PR #1632, opened 2023-03-30
New geoclue

zpytela:


funnelfiasco:
Fixes the behavior in [RHBZ #2183190](https://bugzilla.redhat.com/show_bug.cgi?id=2183190) on my system.

PR #1648, opened 2023-04-14
Allow generic kernel helper to read inherited kernel pipes

WOnder93:
When /usr/bin/false is set as the coredump handler, it inherits a readable pipe from the kernel, which it should to be allowed to read, otherwise a denial is reported when the helper is being started.

Denial example:
type=AVC msg=audit(04/14/2023 09:11:03.033:1343) : avc:  denied  { read } for  pid=8582 comm=false path=pipe:[33516] dev="pipefs" ino=33516 scontext=system_u:system_r:kernel_generic_helper_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=fifo_file permissive=0


zpytela:
merging, thank you.

PR #1652, opened 2023-04-19
Keyutils policy fixes

WOnder93:
Extend the keyutils_request_t policy to cover denials triggered by the `testcases/kernel/syscalls/request_key/request_key03` test from [LTP](https://github.com/linux-test-project/ltp) on Fedora ELN (not sure why, but the same doesn't occur on Rawhide).

See also: https://gitlab.com/redhat/centos-stream/tests/kernel/kernel-tests/-/issues/1604


zpytela:
Merging, thank you.

PR #1662, opened 2023-04-26
Allow keepalived to manage its tmp files

5umm3r15:
Allow keeaplived to manage keepalived_tmp_t files

Resolves: rhbz#2179212


zpytela:
LGTM

PR #1672, opened 2023-05-04
Keyutils-cifs

zpytela:


WOnder93:
I would prefer to have the `cifs.upcall` stuff in a separate new policy module (called `cifs_utils`), since both `/usr/sbin/cifs.upcall` and `/etc/request-key.d/cifs.spnego.conf` (which registers it with the request-key dispatcher) are shipped by `cifs-utils`, which is independent from `keyutils`.

WOnder93:
Also, I wonder if using a separate new type for `/usr/sbin/key.dns_resolver` wouldn't be better. I suspect it will do a different set of things from `/usr/sbin/request-key` and by principle of least privilege it would be better to confine each separately. (But if it is shown that it does only a subset of what we already have for `keyutils_request_t`, then it can be fine to just add [the missing `execute_no_trans`](https://bugzilla.redhat.com/show_bug.cgi?id=2182643#c9) and do it this way.)

PR #1686, opened 2023-05-11
Allow unconfined_service_t to create .gnupg labeled as gpg_secret_t

wrabcak:
System is connected to Red Hat Insights via "/usr/bi/rhc connect" command executed using systemd service. This results that process rhc has label unconfined_service_t. The process creates directory /root/.gnupg. Following commint adds type transition defined as: "type_transition unconfined_service_t admin_home_t:dir gpg_secret_t .gnupg;"

To ensure the directory gets the correct SELinux label also for regular user, file transition rules were also added.

### DIFF
--- a/policy/modules/system/unconfined.te
+++ b/policy/modules/system/unconfined.te
@@ -45,3 +45,7 @@ optional_policy(`
 optional_policy(`
 	container_runtime_domtrans(unconfined_service_t)
 ')
+
+optional_policy(`
+    gpg_manage_admin_home_content(unconfined_service_t)
+')


zpytela:
@wrabcak, I don't quite understand which service ends in this context. /usr/bin/rhc is just a command and I can't find any service executing this command.

wrabcak:
There is a systemd one shot service for insights registration defined as: 
```
# /usr/lib/systemd/system/osbuild-first-boot.service
[Unit]
Description=OSBuild First Boot Service
ConditionPathExists=/etc/osbuild-first-boot
Wants=network-online.target
After=network-online.target

[Service]
Type=oneshot

ExecStart=/usr/bin/rm /etc/osbuild-first-boot
ExecStart=/usr/bin/rhc connect -o=XXX -a=hms-testkey --server subscription... 
```
Systemd is labeled as init_t and rhc binary is labeled as bin_t. So the rhc end up running as unconfined_service_t and rhc is executing gpg.

zpytela:
OK, still, this is a system service, so managing user home content should not be required.

wrabcak:
> OK, still, this is a system service, so managing user home content should not be required.

Fixed.

zpytela:
Merging, thank you.

PR #1687, opened 2023-05-15
Allow virsh name_connect virt_port_t

5umm3r15:
Allow virsh to tcp connection to the virt port.
Resolves: rhzb#2187290


zpytela:
LGTM

PR #1692, opened 2023-05-18
Label /usr/lib/systemd/system/proftpd.* & vsftpd.* with ftpd_unit_file_t

5umm3r15:
The files (/usr/lib/systemd/system/proftpd.* & /usr/lib/systemd/system/vsftpd.*) were labeled as 'iptables_unit_file_t' in the ftp.fc  rather than 'ftpd_unit_file_t',

Resolves: rhbz#2188173


zpytela:
Merging, hopefully not many other domains will be affected.

PR #1694, opened 2023-05-18
Allow wireguard to rw network sysctls

5umm3r15:
Allow wireguard to read and write to sysctl network files. 
Add interface and allow iptables to read wireguard to read fifo files.

Resolves: rhbz#2192154

### DIFF
--- a/policy/modules/system/iptables.te
+++ b/policy/modules/system/iptables.te
@@ -209,3 +209,8 @@ optional_policy(`
 optional_policy(`
 	udev_read_db(iptables_t)
 ')
+
+optional_policy(`
+	wireguard_read_fifo_files(iptables_t)
+')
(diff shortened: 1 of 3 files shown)


policy/modules/system/iptables.te
```diff
@@ -209,3 +209,9 @@ optional_policy(`
 optional_policy(`
 	udev_read_db(iptables_t)
 ')
+
+optional_policy(`
+	wireguard_read_fifo_files(iptables_t)
+')
+
+
```
zpytela: Please remove trailing blank lines.

5umm3r15: fixed, thank you

zpytela:
LGTM

PR #1696, opened 2023-05-22
Replace cifsutils_helper_domtrans() with keyutils_request_domtrans_to()

WOnder93:
rpc.te already uses this interface to "register" a keyutils helper, so make cifsutils.te do the same instead of creating another interface.


zpytela:
Well, I thought there would be another domain in the need of a domain transitions, but it is also possible this way. Thank you.

PR #1698, opened 2023-05-22
Allow kernel to manage its own BPF objects

WOnder93:
Kernel threads may end up calling __sys_bpf(), which does the usual BPF access checks, so make sure kernel_t is allowed to at least operate on its own BPF fds.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2186595


zpytela:
Merging, thank you.

PR #1706, opened 2023-05-25
Update the cifs-helper policy

zpytela:


WOnder93:
FYI: the build failure is caused by https://bugzilla.redhat.com/show_bug.cgi?id=2209989. The fix is already in buildroot, but it will take some time before it gets to the compose/mirrors as well.

zpytela:
> FYI: the build failure is caused by https://bugzilla.redhat.com/show_bug.cgi?id=2209989. The fix is already in buildroot, but it will take some time before it gets to the compose/mirrors as well.

Thank you, it took me some time to find out the problem is not in the pr, but I did not find where it actually is.

PR #1709, opened 2023-05-26
Allow firewalld rw ica_tmpfs_t files

5umm3r15:
When restarting the firewalld service on systems where the openssl-ibmca engine is configured systemwide, SELinux prevents firewalld to read and write ica tmpfs files

Resolves: rhbz#2207487


zpytela:
LGTM

PR #1710, opened 2023-05-26
[DO NOT MERGE] userdom_base_user_template: Define role corresponding to the new user

vmojzis:
The template creates a new SELinux user, but requires the corresponding role, meaning that the policy utilizing the interface needs to look as follos to work:

role <name>_r;
userdom_base_user_template(<name>)

This also breaks the policy generated by
sepolicy generate --term_user -n <username>


vmojzis:
Work in progress, do not merge.

PR #1718, opened 2023-05-30
Allow NetworkManager_dispatcher_dhclient_t to execute shells without a domain transition

naokitnk:
Need to allow NetworkManager_dispatcher_dhclient_t to execute shells without a domain transition since /usr/lib/NetworkManager/dispatcher.d/11-dhclient is a shell script but corecmd_exec_bin(networkmanager_dispatcher_plugin) calls files_exec_all_base_ro_files(networkmanager_dispatcher_plugin) only when mls is not enabled.


zpytela:
Merging, thank you.

PR #1725, opened 2023-06-01
Allow kdumpctl_t to execmem

naokitnk:
Need to explicitly allow kdumpctl_t to execmem as the domain is not unconfined when mls is used.

More specifically, "grep -P" introduced in kdump-lib.sh in the following commit for kexec-tools needs this execmem permission.

kdump-lib: attempt to fix BOOT_IMAGE detection
https://src.fedoraproject.org/rpms/kexec-tools/c/f9c32372d2c8d5e58024d2ddc0b70498c696b5d8


zpytela:
Please move the tunable block before the first optional_policy one.

naokitnk:
> Please move the tunable block before the first optional_policy one.

done

zpytela:
Merging, thank you.

PR #1729, opened 2023-06-09
Set default ports for keylime policy

Koncpa:
Keylime use several default ports, set keylime label for these default ports.

Signed-off-by: Patrik Koncity <pkoncity@redhat.com>

Resolves: RHEL-594


zpytela:
Merging, thank you.

PR #1730, opened 2023-06-09
Label /var/log/kdump.log with kdump_log_t

naokitnk:
Need to label /var/log/kdump.log with kdump_log_t type as kdumpctl_t domain is not unconfined when mls is used.

FYI, /var/log/kdump.log was introduced in kdumpctl in the following commit for kexec-tools.

kdumpctl: add the '-d' option to enable the kexec loading debugging messages
https://src.fedoraproject.org/rpms/kexec-tools/c/88a8b94de9994264f1520441656300d6658801da

### DIFF
--- a/policy/modules/contrib/kdump.te
+++ b/policy/modules/contrib/kdump.te
@@ -28,6 +28,9 @@ systemd_unit_file(kdump_unit_file_t)
 type kdump_lock_t;
 files_lock_file(kdump_lock_t)
 
+type kdump_log_t;
+logging_log_file(kdump_log_t)
+
 type kdumpctl_t;
 type kdumpctl_exec_t;
 init_daemon_domain(kdumpctl_t, kdumpctl_exec_t)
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/kdump.te
```diff
@@ -100,6 +103,9 @@ allow kdumpctl_t self:unix_stream_socket create_stream_socket_perms;
 manage_files_pattern(kdumpctl_t, kdump_lock_t, kdump_lock_t)
 files_lock_filetrans(kdumpctl_t, kdump_lock_t, file, "kdump")
 
+manage_files_pattern(kdumpctl_t, var_log_t, kdump_log_t)
```
zpytela: `var_log_t` is not provided by this module, can
```
manage_files_pattern(kdumpctl_t, kdump_log_t, kdump_log_t)
```
be used instead?

naokitnk: Correct. Updated it.

FYI, I thought it needs to be `var_log_t` since `kdumpctl_t` does need write permission on dirs with `var_log_t` type when `kdump.log` is created. But, I just realized that `filetrans_pattern($1, var_log_t, $2, $3, $4)` in `logging_log_filetrans()` does that job. So, we can have `kdump_log_t` in the second parameter of `manage_files_pattern()` as you pointed out.

zpytela:
Merging, thank you.

PR #1734, opened 2023-06-13
Allow rpm_t sys_admin capability

naokitnk:
Need to allow rpm_t sys_admin capability as the domain is not unconfined when mls is used.

FYI, rpm_t needs this capability to handle IMA signatures, which was recently introduced as described below. More specifically, setxattr() system call to set security.ima attribute needs this capability.

https://fedoraproject.org/wiki/Changes/Signed_RPM_Contents

### DIFF
--- a/policy/modules/contrib/rpm.te
+++ b/policy/modules/contrib/rpm.te
@@ -75,7 +75,7 @@ files_tmpfs_file(rpm_script_tmpfs_t)
 #
 
 allow rpm_t self:capability2 block_suspend;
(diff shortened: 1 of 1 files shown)


zpytela:
Please include the information into the commit message, in particular using the security.ima attribute seems to me quite important.

naokitnk:
> Please include the information into the commit message, in particular using the security.ima attribute seems to me quite important.

done

zpytela:
Merging, thank you.

PR #1663, opened 2023-04-26
Allow rpc.statd to search network sysctl dirs

5umm3r15:
Addresses the following AVC denial:

time->Wed Mar  1 08:23:45 2023
type=PROCTITLE msg=audit(1677677025.236:151): proctitle="/usr/sbin/rpc.statd" type=SYSCALL msg=audit(1677677025.236:151): arch=c000003e syscall=257 success=no exit=-13 a0=ffffff9c a1=7ffc8ff7d990 a2=80100 a3=0 items=0 ppid=1 pid=21104 auid=4294967295 uid=29 gid=29 euid=29 suid=29 fsuid=29 egid=29 sgid=29 fsgid=29 tty=(none) ses=4294967295 comm="rpc.statd" exe="/usr/sbin/rpc.statd" subj=system_u:system_r:rpcd_t:s0 key=(null) type=AVC msg=audit(1677677025.236:151): avc:  denied  { search } for  pid=21104 comm="rpc.statd" name="net" dev="proc" ino=34064 scontext=system_u:system_r:rpcd_t:s0 tcontext=system_u:object_r:sysctl_net_t:s0 tclass=dir permissive=0

Resolves: rhbz#2175516

PR #1740, opened 2023-06-14
Allow rpcd_t search sysctl_net_t dirs

5umm3r15:
Allow rpcd search network sysctl directories

Resolves: rhbz#2175516


zpytela:
Looks like a dup of https://github.com/fedora-selinux/selinux-policy/pull/1663

PR #1741, opened 2023-06-14
Allow collectd_t read proc_net link files

5umm3r15:
Allow collectd read generic symbolic links in /proc

Resolves: rhbz#2209650


zpytela:
LGTM

PR #1746, opened 2023-06-15
Allow nfsd_t to list exports_t dirs

naokitnk:
Need to allow nfsd_t to list exports_t dirs as nfs_export_all_ro boolean is off by default when mls is used.


zpytela:
LGTM

PR #1752, opened 2023-06-20
Label /var/run/tmpfiles.d/static-nodes.conf with kmod_var_run_t

naokitnk:
Need to label /var/run/tmpfiles.d/static-nodes.conf with kmod_var_run_t type as kmod_t domain is not unconfined when mls is used.

### DIFF
--- a/policy/modules/system/modutils.fc
+++ b/policy/modules/system/modutils.fc
@@ -35,3 +35,4 @@ ifdef(`distro_gentoo',`
 /usr/lib/modules/modprobe\.conf -- 	gen_context(system_u:object_r:modules_conf_t,s0)
 
 /var/run/tmpfiles.d/kmod.conf --	gen_context(system_u:object_r:kmod_var_run_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
Is the "owner" really kmod and not udev?

naokitnk:
> Is the "owner" really kmod and not udev?

As static-nodes.conf is an output of kmod generated in the following line of systemd's kmod-static-nodes.service, yes, I understand its owner is kmod.

https://github.com/systemd/systemd/blob/main/units/kmod-static-nodes.service.in#LL20C1-L20C93
`ExecStart={{KMOD}} static-nodes --format=tmpfiles --output=/run/tmpfiles.d/static-nodes.conf`

zpytela:
@naokitnk Thank you, apparently I went down the wrong path, please add the reference to the commit message.
Also please ensure formatting guidelines are followed, namely text width.

naokitnk:
> @naokitnk Thank you, apparently I went down the wrong path, please add the reference to the commit message. Also please ensure formatting guidelines are followed, namely text width.

done

zpytela:
Thank you, merging.

PR #1755, opened 2023-06-22
Allow fsadm_t to get attributes of cgroup filesystems

naokitnk:
Need to allow fsadm_t to get attributes of cgroup filesystems as the domain is not unconfined when mls is used.

Specifically, systemd-fsck running in fsadm_t domain needs this permission.


zpytela:
LGTM

PR #1756, opened 2023-06-23
Fix location for /run/nsd

5umm3r15:
Resolves: rhbz#2181600

### DIFF
--- a/policy/modules/contrib/nsd.fc
+++ b/policy/modules/contrib/nsd.fc
@@ -16,5 +16,7 @@
 /var/lib/nsd(/.*)?		gen_context(system_u:object_r:nsd_zone_t,s0)
 /var/run/nsd\.pid	--	gen_context(system_u:object_r:nsd_var_run_t,s0)
 /var/run/nsd\.ctl	-s	gen_context(system_u:object_r:nsd_var_run_t,s0)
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/nsd.fc
```diff
@@ -14,7 +14,7 @@
 /usr/sbin/nsd-control-setup		--	gen_context(system_u:object_r:nsd_exec_t,s0)
 
 /var/lib/nsd(/.*)?		gen_context(system_u:object_r:nsd_zone_t,s0)
-/var/run/nsd\.pid	--	gen_context(system_u:object_r:nsd_var_run_t,s0)
-/var/run/nsd\.ctl	-s	gen_context(system_u:object_r:nsd_var_run_t,s0)
+/run/nsd/nsd\.pid	--	gen_context(system_u:object_r:nsd_var_run_t,s0)
+/run/nsd/nsd\.ctl	-s	gen_context(system_u:object_r:nsd_var_run_t,s0)
```
zpytela: The legacy /var/run path needs to be used because of equivalency rules. I think the parent directory should be labeled, too, or perhaps by one rule. File transition for the directory is not needed?

zpytela (changes requested):

zpytela:
RHEL 9 version still uses files in /run, so we'd better keep them.

5umm3r15:
fixed, thank you

zpytela:
LGTM

PR #1758, opened 2023-06-23
Make systemd_tmpfiles_t MLS trusted for lowering the level of files

naokitnk:
Need to make systemd_tmpfiles_t MLS trusted for lowering the level of files as the domain needs to lower the levels of device files when mls is used.

Specifically, systemd_tmpfiles_t needs to lower the sensitivity level of /dev/fuse from s15 to s0.

Note that another way we could consider is to change the default sensitivity level of /dev/fuse from s0 to s15 so that the sensitivity level does not need to be lowered from s15 to s0 in the first place. However, the default sensitivity level of /dev/fuse was intentionally changed from s15 (mls_systemhigh) to s0 in the following commit before. So, I don't understand changing it back to s15 (mls_systemhigh) is a good idea.

/dev/fuse should be s0 not mls_high
https://github.com/fedora-selinux/selinux-policy/commit/105e85ac8eaed2fee7c164eb120064b9478d492f

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -692,6 +692,7 @@ files_map_system_db_files(systemd_tmpfiles_t)
 mls_file_read_all_levels(systemd_tmpfiles_t)
 mls_file_write_all_levels(systemd_tmpfiles_t)
 mls_file_upgrade(systemd_tmpfiles_t)
+mls_file_downgrade(systemd_tmpfiles_t)
 
 selinux_get_enforce_mode(systemd_tmpfiles_t)
 selinux_setcheckreqprot(systemd_tmpfiles_t)


naokitnk:

FYI, below is the corresponding error message in /var/log/messages this commit addresses.

> audit: SELINUX_ERR op=security_validate_transition seresult=denied oldcontext=system_u:object_r:fuse_device_t:s15:c0.c1023 newcontext=system_u:object_r:fuse_device_t:s0 taskcontext=system_u:system_r:systemd_tmpfiles_t:s0-s15:c0.c1023 tclass=chr_file

I understand this is started being observed with mls when systemd-tmpfiles service started using /usr/lib/tmpfiles.d/static-nodes-permissions.conf to set permissions and security contexts of device files including /dev/fuse because of the following commit:

tmpfiles: override permissions of static nodes that need this
https://github.com/systemd/systemd/commit/1f9290fcc48d44373174497b5500b028778db6ab

zpytela:
Please include the audit denial and the additional information in the commit.

naokitnk:
> Please include the audit denial and the additional information in the commit.

done

zpytela:
Thank you. I know I keep bothering you, but the commit message is important especially from long-term perspective in commits like this.

naokitnk:
> Thank you. I know I keep bothering you, but the commit message is important especially from long-term perspective in commits like this.

No problem at all. Could you merge this now that I have added the info to the commit message as you requested?

zpytela:
Sure, thank you.

PR #1772, opened 2023-07-03
Allow virt_qemu_ga_t create .ssh dir with correct label

5umm3r15:
Create .ssh directory in the user home directory with an correct label.

Add capabilities: chown dac_override dac_read_search fsetid fowner

Resolves: rhbz#2181402

### DIFF
--- a/policy/modules/contrib/virt.te
+++ b/policy/modules/contrib/virt.te
@@ -1836,7 +1836,10 @@ tunable_policy(`virt_rw_qemu_ga_data',`
 ')
 
 optional_policy(`
+    ssh_filetrans_home_content(virt_qemu_ga_t)
         tunable_policy(`virt_qemu_ga_manage_ssh',`
+                allow virt_qemu_ga_t self:capability { chown dac_override dac_read_search fowner fsetid };
+
 		ssh_manage_home_files(virt_qemu_ga_t)
         ')
 ')


zpytela:
Do I understand correctly the newly added permissions match the purpose of the `virt_qemu_ga_manage_ssh` boolean or is the scenario any different?

5umm3r15:
Yes, this is correct. 
.ssh directory was created with wrong label user_home_dir_t and should be fixed with  ssh_filetrans_home_content.

Also based on [Milos comment](https://bugzilla.redhat.com/show_bug.cgi?id=2181402#c9) capabilities were required.

zpytela:
So both the groups should be inside a tunable_policy block?

5umm3r15:
I had problem with inserting filetrans rule into tunable policy. There was error

zpytela:
I see. Would it make sense then (semantically) put the additional allow rules to either tunable or optional block which these rules belong to?

5umm3r15:
Can you please check it now?

zpytela:
Better, can you also move the allow rules to the optional block?

5umm3r15:
This way?
```
optional_policy(`
    allow virt_qemu_ga_t self:capability { chown dac_override dac_read_search fowner fsetid };
    ssh_filetrans_home_content(virt_qemu_ga_t)
        tunable_policy(`virt_qemu_ga_manage_ssh',`
                ssh_manage_home_files(virt_qemu_ga_t)
        ')
')
```

5umm3r15:
This way? Or to the tunable?

```
optional_policy(`
    allow virt_qemu_ga_t self:capability { chown dac_override dac_read_search fowner fsetid };
    ssh_filetrans_home_content(virt_qemu_ga_t)
        tunable_policy(`virt_qemu_ga_manage_ssh',`
                ssh_manage_home_files(virt_qemu_ga_t)
        ')
')
```

zpytela:
In which case are they needed? I believe it rather should be inside the tunable block, but I may not read the bz properly.

5umm3r15:
I wasn't able to test it properly, so fixes are based on the comments

zpytela:
OK, merging.

PR #1773, opened 2023-07-03
Permit kernel_t to change the user identity in object contexts

WOnder93:
This is needed to make overlayfs mounts created before initial policy load work correctly.

Resolves: rhbz#2215454


zpytela:
Merging, thank you.

PR #1775, opened 2023-07-10
Update cifs.upcall policy according to BZ 2220943

WOnder93:
Resolves: rhbz#2220943

### DIFF
--- a/policy/modules/contrib/cifsutils.te
+++ b/policy/modules/contrib/cifsutils.te
@@ -57,6 +57,7 @@ optional_policy(`
 
 optional_policy(`
 	sssd_stream_connect(cifs_helper_t)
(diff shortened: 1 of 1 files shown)


abbra:
The build failure seems to be related to dnf5 parallel installation in your CI. Any plans to fix that as it prevents from getting  most PRs in?

WOnder93:
It is not a bug on our side. I wasn't looking into it, as I expected that a fixed container image would eventually become available. But it seems [the default Docker registry may not always have the rawhide image up to date](https://discussion.fedoraproject.org/t/84680/4) and we're supposed to use `quay.io`. I opened #1848 to switch to `quay.io`, which indeed fixes the build.

zpytela:
Merging, thank you.

PR #1781, opened 2023-07-14
Allow logrotate_t to map generic files in /etc

naokitnk:
Need to allow logrotate_t to map generic files in /etc as the domain doesn't have map permission on files with types having base_ro_file_type attribute including etc_t when mls is used.

Specifically, logrotate_t needs map permission on /etc/logrotate.conf with etc_t type.

### DIFF
--- a/policy/modules/contrib/logrotate.te
+++ b/policy/modules/contrib/logrotate.te
@@ -132,6 +132,7 @@ domain_getattr_all_entry_files(logrotate_t)
 # Read /proc/PID directories for all domains.
 domain_read_all_domains_state(logrotate_t)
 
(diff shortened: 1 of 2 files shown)


naokitnk:

type=AVC msg=audit(1609240081.842:4033): avc:  denied  { map } for  pid=33609  comm="logrotate" path="/etc/logrotate.conf" dev="vda3" ino=4781910 scontext=system_u:system_r:logrotate_t:s0-s15:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0

zpytela:
Merging, thank you.

PR #1785, opened 2023-07-18
Allow systemd_network_generator_t notify systemd manager

zpytela:
Allow systemd-network-generator send system log messages

Logging support for the network generator was added by systemd commit
8b24bcdfa8 ("network-generator: Add missing log_setup()",
https://github.com/systemd/systemd/commit/8b24bcdfa8).

Resolves: rhbz#2223571

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1217,6 +1217,10 @@ init_named_pid_filetrans(systemd_network_generator_t, net_conf_t, dir, "network"
 sysnet_manage_config(systemd_network_generator_t)
 sysnet_manage_config_dirs(systemd_network_generator_t)
 
+optional_policy(`
+	logging_send_syslog_msg(systemd_network_generator_t)
+')
+
 #######################################
 #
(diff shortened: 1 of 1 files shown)


WOnder93:
The description doesn't seem to be correct... The denial in the BZ is for the journald socket and another one for /dev/kmsg (I assume this one only pops up as a fallback after the first one is denied). I believe the right interface for that is `logging_send_syslog_msg()`.

keszybz:
Yeah, the description doesn't seem to agree with the patch. systemd-network-generator does not use `Type=notify-reload`.

I think https://github.com/systemd/systemd/commit/8b24bcdfa8c5a4b32cf273d74dc5130626c44c22 is the culprit. Simply speaking, logging was broken, and now we started logging. It probably tries to log to the journal, but gets a denial, and falls back to kmsg.

zpytela:
Thank you both for the review, I apparently followed an incorrect lead written in oneof the reports and subsequently mismatched the problem with another one, still this was a draft PR where things can happen.

PR #1787, opened 2023-07-19
Dontaudit the execute permission on sock_file globally

WOnder93:
It is a nonsensical combination, so it doesn't matter if it's allowed or denied. Remove the few existing rules involving it and instead add a global dontaudit.

Related: rhbz#2216408

### DIFF
--- a/policy/modules/contrib/openshift.te
+++ b/policy/modules/contrib/openshift.te
@@ -219,7 +219,6 @@ allow openshift_domain openshift_initrc_t:tcp_socket getattr;
 
 dontaudit openshift_domain openshift_initrc_tmp_t:file append;
 dontaudit openshift_domain openshift_var_run_t:file append;
-dontaudit openshift_domain openshift_file_type:sock_file execute;
 
 kernel_dontaudit_search_network_state(openshift_domain)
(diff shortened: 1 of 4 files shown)


zpytela:
I seem not quite understand what's the point in dontauditing - if this is a wrong behaviour, shouldn't the maintainer rather be notified to change it?

On the other hand it is in a way similar to https://github.com/fedora-selinux/selinux-policy/commit/7595359c13

WOnder93:
Yes, similar arguments apply here (although in the other case it is guaranteed that the syscall would error out either way, while here access(2) could succeed). Global dontauditing is better than global allowing and simpler than ad-hoc allowing or dontauditing.

Let's not make SELinux a bug catching tool - we don't have capacity to investigate bugs for other components. When it's inevitable we of course do our best, but we shouldn't make it a feature to audit each kind of weird behavior, only the potentially unsafe subset. Plus, in this particular case any bug found is most likely benign and thus there is not that much value in finding and fixing it.

zpytela:
Certainly purpose of SELinux is different, but some changes make troubleshooting harder which is also the case of dontaudit rules.

zpytela:
Merging, thank you.

PR #1789, opened 2023-07-19
Fix the CI to work with DNF5

WOnder93:
The --srpm parameter is no longer supported by DNF5 (default in rawhide) and it doesn't seem to be needed even with the old DNF, so just remove it.


zpytela:
Merging, thank you.

PR #1796, opened 2023-07-25
Set the MLS range of fsdaemon_t to s0 - mls_systemhigh

naokitnk:
Need to set the MLS range of fsdaemon_t to s0 - mls_systemhigh to meet mls-specific constraints when the domain tries to write to sock_file /var/run/systemd/notify with init_var_run_t type having sensitivity level s0.

type=AVC msg=audit(1611583033.083:202): avc:  denied  { write } for  pid=1375  comm="smartd" name="notify" dev="tmpfs" ino=15758 scontext=system_u:system_r:fsdaemon_t:s15:c0.c1023 tcontext=system_u:object_r:init_var_run_t:s0 tclass=sock_file permissive=0

### DIFF
--- a/policy/modules/contrib/smartmon.te
+++ b/policy/modules/contrib/smartmon.te
@@ -30,7 +30,7 @@ type fsdaemon_tmp_t;
 files_tmp_file(fsdaemon_tmp_t)
 
 ifdef(`enable_mls',`
-	init_ranged_daemon_domain(fsdaemon_t, fsdaemon_exec_t, mls_systemhigh)
+	init_ranged_daemon_domain(fsdaemon_t, fsdaemon_exec_t, s0 - mls_systemhigh)
 ')
 
 ########################################


naokitnk:

FYI, below is the mlsconstrain, because of which we need this change:

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/mls#L90C1-L96C31

```
# the "single level" file "write" ops
mlsconstrain { file lnk_file fifo_file dir chr_file blk_file sock_file } { write create setattr relabelfrom append unlink link rename mounton }
	(( l1 eq l2 ) or
	 (( t1 == mlsfilewritetoclr ) and ( h1 dom l2 ) and ( l1 domby l2 )) or
	 (( t2 == mlsfilewriteinrange ) and ( l1 dom l2 ) and ( h1 domby h2 )) or
	 ( t1 == mlsfilewrite ) or
	 ( t2 == mlstrustedobject ))
```

In this case, l1 is s15 and l2 is s0. So, "l1 eq l2" and "l1 domby l2" are false. Similarly, h1 is s15 and h2 is s0. So, "h1 domby h2" is also false. Also, fsdaemon_t doesn't have mlsfilewritetoclr nor mlsfilewrite attributes, and init_var_run_t doesn't have mlsfilewriteinrange nor mlstrustedobject attributes. So, they don't meet the above constraint.

Note that adding mlsfilewrite attribute to fsdaemon_t or adding mlstrustedobject attribute to init_var_run_t would also meet the constraint. But I understand they are most likely overprovisioning and thus we shouldn't do so.

zpytela:
Merging, thank you.

PR #1798, opened 2023-07-26
files_dontaudit_getattr_tmpfs_files allowed the access and didn't dontaudit it

jsegitz:


zpytela:
Also an interesting finding, the commit dates back to 2010. The interface is used for mdadm_t, I'll merge it when I find some test which successfully runs.

zpytela:
Merging, sorry for the delay.

PR #1799, opened 2023-07-26
files_dontaudit_mounton_modules_object allowed the access and didn't dontaudit it

jsegitz:
dontaudit it


zpytela:
That's an interesting finding. The interface is used for init_t, so I'd like to wait with merging when f39 branches off f38 if you don't mind.

zpytela:
Merging, sorry for the delay.

PR #1810, opened 2023-08-01
Fix labeling for no-stub-resolv.conf

5umm3r15:
Restorecon command relabels /run/NetworkManager/no-stub-resolv.conf from system_u:object_r:NetworkManager_var_run_t:s0 to system_u:object_r:net_conf_t:s0.

Resolves: rhbz#2148390


zpytela:
LGTM

PR #1819, opened 2023-08-07
Allow pdns name_bind and name_connect all ports

5umm3r15:
Allow powerdns name bind all udp sockets and name connect all tcp sockets. allow powerdns read the kerberos key table.

Resolves: rhbz#2047945


zpytela:
LGTM

PR #1821, opened 2023-08-08
Remove nsplugin_role from mozilla.if

5umm3r15:
Nsplugin module was removed from SELinux policy,
but interface nsplugin_role was still included in mozilla_role interface, which caused syntax error, when using this interface.

Resolves: rhbz#2221251


zpytela:
LGTM

PR #1827, opened 2023-08-09
Add loongson support for selinux-policy

wangqi123a:
Add loongson support for selinux-policy


zpytela:
I am sorry for the delay in responding. So far, we have never had conditional rules based on a boolean, compile-time ifdefs were used instead. Why did you decide to use booleans?

PR #1830, opened 2023-08-10
Allow gpsd write user_tty_device_t chr_file

5umm3r15:
Allow gpsd read and write a user domain tty

Resolves: rhbz#2223305


zpytela:
LGTM

PR #1831, opened 2023-08-10
Update policy for fdo

5umm3r15:
FIDO device onboard is used for automated and secure IoT onboarding.

Resolves: rhbz#2229722

### DIFF
--- a/policy/modules/contrib/fdo.fc
+++ b/policy/modules/contrib/fdo.fc
@@ -1,3 +1,7 @@
+/boot/device-credentials			--      gen_context(system_u:object_r:fdo_device_credentials_t,s0)
+
+/etc/device-credentials				--	gen_context(system_u:object_r:fdo_device_credentials_t,s0)
+/etc/device_onboarding_performed		--	gen_context(system_u:object_r:fdo_device_credentials_t,s0)
 /etc/fdo(/.*)?						gen_context(system_u:object_r:fdo_conf_t,s0)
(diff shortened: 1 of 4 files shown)


policy/modules/contrib/fdo.fc
```diff
@@ -5,6 +5,9 @@
 /etc/fdo/aio/logs(/.*)?					gen_context(system_u:object_r:fdo_conf_rw_t,s0)
 /etc/fdo/aio/stores(/.*)?				gen_context(system_u:object_r:fdo_conf_rw_t,s0)
 
+/etc/fdo/stores/manufacturing_sessions(/.*)?		gen_context(system_u:object_r:fdo_conf_rw_t,s0)
+/etc/fdo/stores/rendezvous_sessions(/.*)?		gen_context(system_u:object_r:fdo_conf_rw_t,s0)
```
runcom, line 16: so, the fdo manufacturing-client that runs in the initrd, writes the file `/etc/device-credentials` but i can't see it anywhere here in the policy - can we add it and have another test package?

policy/modules/contrib/fdo.fc
```diff
@@ -1,10 +1,17 @@
+
+/etc/device-credentials				--	gen_context(system_u:object_r:fdo_conf_rw_t,s0)
```
runcom: I also think we need to add `/boot/device-credentials` as that's where this file comes from

5umm3r15: thank you

PR #1838, opened 2023-08-11
Allow init_t to manage the dhcp client state files

naokitnk:
Need to allow init_t to manage the dhcp client state files as the process of import-state service runs in that domain when mls is used while it runs in unconfined_service_t domain with targeted policy.

audit(1615330079.292:7): avc:  denied  { add_name } for  pid=1168 comm="cp" name="dhclient-c9935e25-13d2-42b8-adb3-63d628e36e48-ens3.lease" scontext=system_u:system_r:init_t:s0-s15:c0.c1023 tcontext=system_u:object_r:dhcpc_state_t:s0 tclass=dir permissive=0

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -771,6 +771,7 @@ ifdef(`distro_redhat',`
     ')
 
     optional_policy(`
+        sysnet_manage_dhcpc_state(init_t)
         sysnet_relabelfrom_dhcpc_state(init_t)
         sysnet_setattr_dhcp_state(init_t)
     ')


naokitnk:

zpytela:
No information is needed, I just considered enabling the dhcp module in mls as an alternative.

naokitnk:
OK, but I understand dhcp policy module is already enabled with mls as you can confirm in modules-mls-contrib.conf below:

https://src.fedoraproject.org/rpms/selinux-policy/blob/rawhide/f/modules-mls-contrib.conf

```
# Layer: services
# Module: dhcp
#
# Dynamic host configuration protocol (DHCP) server
# 
dhcp = module
```


Moreover, the cause here is rather in the source domain side. Specifically, I understand import-state service provided by initscripts package needs this permission to execute the following line in init_t domain with mls policy while it's allowed with targeted policy since the process runs in unconfined_service_t domain instead as I wrote:

https://github.com/fedora-sysv/initscripts/blob/main/usr/libexec/import-state#L27

`  find . -mindepth 1 -maxdepth 1 -not -type d -exec cp -av -t "$dest_dir" {} \; > /dev/null`

naokitnk:
Any updates?

PR #1848, opened 2023-08-22
Use quay.io as the container image source for CI

WOnder93:
The current rawhide image in Docker registry is out-of-date and fails to install depndencies due to DNF/DNF5 conflict. quay.io should be a more reliable source, as per:
https://discussion.fedoraproject.org/t/84680/4


zpytela:
Merging, thank you.

PR #1854, opened 2023-08-25
Add the unconfined_read_files() and unconfined_list_dirs() interfaces

zpytela:


WOnder93:
Where is the justification? Also, the PR only adds unconfined_read_files() currently.

zpytela:
Thanks forspotting this, it was a result of an add-remove chain. I also added some justification.

PR #1856, opened 2023-08-25
Update policy for qatlib

5umm3r15:
Allow qatlib read generic files in /proc,
set the attributes of /dev directories,
create hardware state information files,
execute insmod in the caller domain.

Allow qatlib to request the kernel to load a module, and load kernel module files,
read the configuration options used when loading modules, and read the dependencies of kernel modules.

Resolves: rhbz#2080443


zpytela:
LGTM

PR #1861, opened 2023-09-01
Allow sysadm_t to mmap modules_object_t files

naokitnk:
Need to allow sysadm_t to mmap modules_object_t files as root user runs in sysadm_t domain when mls is used while it runs in unconfined_t domain with targeted policy.


zpytela:
merging, thank you

PR #1866, opened 2023-09-08
Add additions to dnsmasq interface

georou:
Hi, please have a look at these suggested changes:

- Allow reading the dnsmasq.d directory that stores additional config files. Currently there's not enough permissions to allow a `read` of the directory.

- Add a new manage_config interface for dnsmasq config files and directory. Thoughts on this one? Only thought to add it for completion sake if someone needed it down the line. How would it affect audit2allow - ie would it recommend this over the write interface which is more restrictive.

- read_lnk_file_perms within dnsmasq_manage_config: I'm not 100% if this is needed. I've seen it sprinkled in other policies for a manage interface and wondered if it would be handy to have it here. Thoughts and guidance?

- Misc comment clean up of dnsmasq.if to make it easier to understand what the interface does.

Thank you.

### DIFF
--- a/policy/modules/contrib/dnsmasq.if
+++ b/policy/modules/contrib/dnsmasq.if
@@ -215,7 +215,6 @@ interface(`dnsmasq_manage_config',`
 	files_search_etc($1)
 	allow $1 dnsmasq_etc_t:dir manage_dir_perms;
 	allow $1 dnsmasq_etc_t:file manage_file_perms;
-	allow $1 dnsmasq_etc_t:lnk_file read_lnk_file_perms;
 ')
 
 ########################################
--- a/policy/modules/contrib/dnsmasq.if
+++ b/policy/modules/contrib/dnsmasq.if
@@ -193,7 +193,7 @@ interface(`dnsmasq_read_config',`
 
 	files_search_etc($1)
 	allow $1 dnsmasq_etc_t:dir list_dir_perms;
-	read_files_pattern($1, dnsmasq_etc_t, dnsmasq_etc_t)
+	allow $1 dnsmasq_etc_t:file read_file_perms;
 ')
 
 ########################################
@@ -235,7 +235,7 @@ interface(`dnsmasq_write_config',`
 
 	files_search_etc($1)
 	allow $1 dnsmasq_etc_t:dir list_dir_perms;
-	write_files_pattern($1, dnsmasq_etc_t, dnsmasq_etc_t)
+	allow $1 dnsmasq_etc_t:file write_file_perms;
 ')
 
 ########################################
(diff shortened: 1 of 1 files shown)


zpytela:
@georou looks good. Comments:
- Is managing directories required, or did you mean it as a support to managing files?
- Does dnsmasq use symlinks anywhere in theconfiguration?
- With auditallow rules, did you mean using `audit2allow -R`? It is just a tool, there is no need to always follow its suggestions
- I see only cobblerd_t is dealing with the configuration, do you know of any other use cases?

georou:
Hi,

- If I understand, yes, in support of files. My thinking is to have manage_config() there in case it's needed by an external app to manipulate directories/files. I'm not married to the idea, just thought it might be handy in the future for someone. I can remove if need be.
- I'm not 110% sure symlinks are used. If there hasn't been a PR about it over all these years, maybe we just go on the safe side of not including them. I've stripped them out in that case. Can always add them in the future if suddenly needed.
- Sorry, what did you mean specifically for clobberd?

Cheers.

georou:
Thank you for the time looking at this. 

I might close this and revisit this PR in the future. I may have jumped the gun on why this is needed.

georou:
Just also looking for some help on this. If my application is throwing a `deny dir: read` on this dnsmasq.d/ directory. Is the responsibility on RHEL's policy to add a list_dir_perms like I've done in this PR or should it be just loaded into my custom local policy for my application? Especially since this is an interface - to be used by a 3rd party. It's an interesting dilemma. 

I've had a look around at other interfaces, so for example: mysql_read_config() has list_dir_perms. So does; munin_read_config().

zpytela:
I am sorry for the delay, I am about to merge the pr now: do you think it is ready?
I mentioned cobbler as this is the only service dealing with dnsmasq config files so possibly affected by this pr.

> If my application is throwing a deny dir: read on this dnsmasq.d/ 
What kind of application, running in which domain, how was it started? We may not get to an universal answer.

georou:
No worries for the delay.

I see what you mean for Cobbler now. Good call. Since this is a an addition, it shouldn't affect it.

My application is for a custom policy for Pi-hole(the popular ad blocking app). They have the requirement to change/read dnsmasq files. That's why I'm hesitant in asking to make this change upstream for a 3rd party requirement. 

The directory in question: `dnsmasq.q/` does hold config files. So for example if your application wants to backup that folder, it will get denied. 

I've found another example that's similar to mysql_read_config() and munin_read_config(). stunnel does the same: https://github.com/fedora-selinux/selinux-policy/blob/06d381f0ad63600bc68c7b20acd1a6fb4ac01391/policy/modules/contrib/stunnel.if#L37

If you're already allowing your policy to read/write to config files, it might make sense to also allow reading the directory in case there's a file that isn't named in a standard way. That's how I see these examples. Please guide my thinking if I'm wrong.

Thoughts so far? If you think it's a valid PR, I'm happy with it so far.

zpytela:
Merging, thank you.

PR #1867, opened 2023-09-08
Change systemd-network-generator transition to include the file class

zpytela:
This is a follow-up commit to f1d705ff01 ("Change file transition for systemd-network-generator") as systemd-network-generator actually creates a file in /run/systemd before moving it to /run/systemd/network, see strace log:
renameat(AT_FDCWD</root>, "/run/systemd/.#network8e6806e244d4b87a", AT_FDCWD</root>, "/run/systemd/network/91-default.network") = 0

so the transition needs to be defined also for a file.

Resolves: rhbz#2230226

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1214,7 +1214,7 @@ optional_policy(`
 # systemd_network_generator domain
 #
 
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe (approved):

PR #1870, opened 2023-09-13
Add policy for nbdkit

5umm3r15:
Add interfaces to virt and fix policy for nbdkit

Resolves: rhbz#2182505

### DIFF
--- a/policy/modules/contrib/nbdkit.fc
+++ b/policy/modules/contrib/nbdkit.fc
@@ -1,0 +1,3 @@
+/usr/sbin/nbdkit	 --    gen_context(system_u:object_r:nbdkit_exec_t,s0)
+
+/usr/lib/systemd/system/nbdkit.*	gen_context(system_u:object_r:nbdkit_unit_file_t,s0)
--- a/policy/modules/contrib/nbdkit.te
+++ b/policy/modules/contrib/nbdkit.te
@@ -1,0 +1,100 @@
+policy_module(nbdkit, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+gen_require(`
+    type unconfined_t;
+')
+
+type nbdkit_t;
+type nbdkit_exec_t;
+application_domain(nbdkit_t, nbdkit_exec_t)
+mcs_constrained(nbdkit_t)
+role system_r types nbdkit_t;
+
+type nbdkit_home_t;
+userdom_user_home_content(nbdkit_home_t)
+
+type nbdkit_tmp_t;
+files_tmp_file(nbdkit_tmp_t)
+
+type nbdkit_unit_file_t;
+systemd_unit_file(nbdkit_unit_file_t)
+
+permissive nbdkit_t;
+
+########################################
+#
+# nbdkit local policy
+#
+allow nbdkit_t self:capability { setgid setuid };
+allow nbdkit_t self:fifo_file rw_fifo_file_perms;
+allow nbdkit_t self:netlink_route_socket rw_netlink_socket_perms;
+allow nbdkit_t self:process { fork setsockcreate signal_perms };
+allow nbdkit_t self:tcp_socket create_stream_socket_perms;
+allow nbdkit_t self:udp_socket create_socket_perms;
+
+manage_dirs_pattern(nbdkit_t, nbdkit_tmp_t, nbdkit_tmp_t)
+manage_files_pattern(nbdkit_t, nbdkit_tmp_t, nbdkit_tmp_t) 
+userdom_user_tmp_filetrans(nbdkit_t, nbdkit_tmp_t, { dir file })
+
+manage_dirs_pattern(nbdkit_t, nbdkit_home_t, nbdkit_home_t)
+manage_files_pattern(nbdkit_t, nbdkit_home_t, nbdkit_home_t) 
+userdom_user_home_dir_filetrans(nbdkit_t, nbdkit_home_t, { dir file })
+
+corenet_tcp_connect_http_port(nbdkit_t)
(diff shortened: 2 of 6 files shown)


policy/modules/contrib/nbdkit.fc
```diff
@@ -0,0 +1,5 @@
+HOME_DIR/tmp(/.*)?        gen_context(system_u:object_r:nbdkit_home_t,s0)
```
zpytela: This will clash with other files created in HOME_DIR/tmp, we currently have <<none> for all files. Probably better to adjust for particular filenames or remove completely.

policy/modules/contrib/nbdkit.te
```diff
@@ -0,0 +1,115 @@
...
+
+manage_dirs_pattern(nbdkit_t, nbdkit_tmp_t, nbdkit_tmp_t)
+manage_files_pattern(nbdkit_t, nbdkit_tmp_t, nbdkit_tmp_t) 
+userdom_user_tmp_filetrans(nbdkit_t, nbdkit_tmp_t, { dir file })
+
+manage_dirs_pattern(nbdkit_t, nbdkit_home_t, nbdkit_home_t)
+manage_files_pattern(nbdkit_t, nbdkit_home_t, nbdkit_home_t) 
+userdom_user_home_dir_filetrans(nbdkit_t, nbdkit_home_t, { dir file })
+
+allow unconfined_t nbdkit_exec_t:file entrypoint;
```
zpytela: This can probably be solved in a way similar to container_runtime_entrypoint/virt_sandbox_entrypoint in unconfineduser.te

policy/modules/contrib/nbdkit.te
```diff
@@ -0,0 +1,115 @@
...
+
+permissive nbdkit_t;
+
+########################################
+#
+# nbdkit local policy
+#
+allow nbdkit_t self:capability { setgid setuid };
+allow nbdkit_t self:fifo_file rw_fifo_file_perms;
+allow nbdkit_t self:netlink_route_socket { bind create getattr nlmsg_read read write };
```
zpytela: Please use appropriate permissions sets for all sockets.

policy/modules/contrib/nbdkit.fc
```diff
@@ -0,0 +1,5 @@
...
+
+/usr/sbin/nbdkit	 --    gen_context(system_u:object_r:nbdkit_exec_t,s0)
+
+/usr/lib/systemd/system/nbdkit.*	gen_context(system_u:object_r:nbdkit_unit_file_t,s0)
```
zpytela, line 3: I cannot find any services like this in Fedora.

policy/modules/contrib/nbdkit.fc
5umm3r15: Thanks, I understand. 

It was based on this path: 
$HOME/tmp/known-hosts

Should I label known-hosts dir?

martinpitt:
Ugh, cockpit f39/rawhide tests are a trainwreck:
 - [rawhide](https://artifacts.dev.testing-farm.io/62ffb2a6-91c9-468b-9a79-0352e3460658/) installs an (old) cockpit-system-303-2.fc40.noarch from Fedora, instead of the cockpit main builds COPR. This is once again that TF tag-repository priority inflation that breaks the show. We need a hack similar to https://github.com/cockpit-project/cockpit-podman/commit/5ab52df9dd43461eae90aa7103fee7513809cf1e in cockpit to fix that. I'll send a PR.
 - [F39](https://artifacts.dev.testing-farm.io/3ea6e248-1a9c-429c-bdda-ee12cf53d6d5/) has the correct cockpit-system RPM from the COPR, but I  have a feeling it's the wrong cockpit-storaged. That installation is unfortunately not shown in the logs.

So either something changed recently in Testing Farm again, or this has always been lurking and some combination of Fedora composes and COPR breaks this. I'll send some debugging improvement and a workaround for that broken TF behaviour, then we can retry these.

Sorry for the mess!

martinpitt:
@thrix: https://gitlab.com/testing-farm/infrastructure/-/blob/testing-farm/ranch/public/citool-config/guest-setup/pre-artifact-installation/templates/tag.repo.j2?ref_type=heads claims that it would exclude cockpit packages, but (1) apparently it doesn't, and (2) this is also wrong: we *do* want the tag repository, it should just not trump COPRs -- we need "highest version wins", not "tag repo is special" (and I believe that is true for most packages). Can you please drop the `exclude=`?

martinpitt:
https://github.com/cockpit-project/cockpit/pull/19547 landed, so please either retry them or force-push this PR, or ignore the two failures.

policy/modules/contrib/nbdkit.fc
zpytela: That should work. If such an entry really is required - it applies only when restorecon is run and the files have incorrect label.

thrix:
@martinpitt messaged you on Slack, I would like to look at it, but seems I am missing what is wrong

policy/modules/contrib/nbdkit.fc
5umm3r15: It is based on the comment [20](https://bugzilla.redhat.com/show_bug.cgi?id=2182505#c20)

zpytela:
LGTM, thank you.

martinpitt:
I talked to Miro, we have a plan for TF to fix this globally, and I just landed the workaround in cockpit. So please click re-run on [the failed rawhide test](https://github.com/fedora-selinux/selinux-policy/pull/1870/checks?check_run_id=18230901432) or do a comment with `/packit retest-failed`. Thanks and sorry for the noise!

PR #1871, opened 2023-09-13
Allow cupsd_t to use bpf capability

JurajMarcin:
Gutenprint uses libusb which in turn uses libudev in such way that it tries to attach a network filter using the setsocketopt syscall on initialization. However, as gutenprint is started by CUPS deamon with cupsd_t context, the action is denied by the current policy.

This patch adds bpf permission to an already existing allow statement for cupsd_t type under which the gutenprint process is running to fix the issue.

Resolves: https://issues.redhat.com/browse/RHEL-3633


5umm3r15:
LGTM

PR #1873, opened 2023-09-20
Dontaudit rhsmcertd write memory device

5umm3r15:
Do not audit attempts to write to raw memory devices

lscpu uses  O_RDONLY to read /dev/mem,
but on ppc64 it uses IBM's librtas.so that mmap /dev/mem using O_RDRW for open().

Resolves: RHEL-1547


zpytela:
LGTM

PR #1886, opened 2023-10-02
Allow kdumpctl_t to list all directories with a filesystem type

naokitnk:
Need to allow kdumpctl_t to list all directories with a filesystem type as the domain is not unconfined when mls is used.

type=AVC msg=audit(1572468079.662:192): avc:  denied  { read } for  pid=2205 comm="find" name="/" dev="efivarfs" ino=11307 scontext=system_u:system_r:kdumpctl_t:s0-s15:c0.c1023 tcontext=system_u:object_r:efivarfs_t:s0 tclass=dir permissive=0


zpytela:
LGTM

PR #1889, opened 2023-10-04
Named anon_inode io_uring usage

zpytela:


zpytela:
@WOnder93 plese review, this is a follow-up to https://github.com/fedora-selinux/selinux-policy/pull/1351/commits/ecb0ee0e5

policy/modules/kernel/kernel.if
```diff
@@ -4619,3 +4619,20 @@ interface(`kernel_io_uring_use',`
...
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`kernel_io_uring_rw',`
+	gen_require(`
+		type io_uring_t;
+	')
+	allow $1 io_uring_t:anon_inode common_anon_inode_perms;
```
WOnder93: After 65b9e0bdceb7e6adbe308f9a591b103cba6986ef, we shouldn't use `common_anon_inode_perms` any more (in favor of listing the permissions verbatim in the few relevant interfaces). In fact, the macro should have been deprected in that commit, but I didn't think of that.

Any io_uring fd use should be covered by `kernel_io_uring_use()` already and it allows most of the perms from `common_anon_inode_perms` except `ioctl` (which I think shouldn't be needed, but I'm not sure). Is this motivated by seeing an actual `ioctl` AVC?

zpytela: These permissions were needed:
(allow named_t io_uring_t (anon_inode (create map read write)))
(allow ndc_t io_uring_t (anon_inode (create map read write)))

zpytela: You are right, I completely overlooked the actual content of the existing interface.

WOnder93 (approved):
This version looks good :+1:

PR #1890, opened 2023-10-04
Confine systemd fstab,sysv,rc-local

5umm3r15:
Confine systemd generators fstab-generator,
rc-local-generator and systemd_sysv_generator_t

Resolves: SELINUX-3172

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -2,8 +2,8 @@ HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_home
 HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit_file_t,s0)
 /root/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_home_t,s0)
 
-/etc/.*hostname.*			--		gen_context(system_u:object_r:hostname_etc_t,s0)
-/etc/machine-info		--		gen_context(system_u:object_r:hostname_etc_t,s0)
+/etc/.*hostname.*	--	gen_context(system_u:object_r:hostname_etc_t,s0)
+/etc/machine-info	--	gen_context(system_u:object_r:hostname_etc_t,s0)
 /etc/udev/.*hwdb.*	--	gen_context(system_u:object_r:systemd_hwdb_etc_t,s0)
 
 /bin/systemd-notify				--		gen_context(system_u:object_r:systemd_notify_exec_t,s0)
@@ -68,7 +68,10 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /usr/lib/systemd/systemd-coredump	--	gen_context(system_u:object_r:systemd_coredump_exec_t,s0)
 /usr/lib/systemd/systemd-modules-load	--	gen_context(system_u:object_r:systemd_modules_load_exec_t,s0)
 /usr/lib/systemd/systemd-network-generator	--	gen_context(system_u:object_r:systemd_network_generator_exec_t,s0)
+/usr/lib/systemd/system-generators/systemd-fstab-generator	--	gen_context(system_u:object_r:systemd_fstab_generator_exec_t,s0)
 /usr/lib/systemd/system-generators/systemd-gpt-auto-generator	--	gen_context(system_u:object_r:systemd_gpt_generator_exec_t,s0)
+/usr/lib/systemd/system-generators/systemd-rc-local-generator	--	gen_context(system_u:object_r:systemd_rc_local_generator_exec_t,s0)
+/usr/lib/systemd/system-generators/systemd-sysv-generator	--	gen_context(system_u:object_r:systemd_sysv_generator_exec_t,s0)
 /usr/lib/systemd/systemd-resolve(d|-host)			gen_context(system_u:object_r:systemd_resolved_exec_t,s0)
 /usr/lib/systemd/systemd-importd		--	gen_context(system_u:object_r:systemd_importd_exec_t,s0)
 /usr/lib/systemd/systemd-journal-upload		--	gen_context(system_u:object_r:systemd_journal_upload_exec_t,s0)
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -29,6 +29,39 @@ template(`systemd_domain_template',`
 
 ######################################
 ## <summary>
+##  Creates types and rules for 
+##  systemd generators.
+## </summary>
+## <param name="prefix">
+##  <summary>
+##  Prefix for the domain.
+##  </summary>
+## </param>
+#
+template(`systemd_generator_template',`
+	gen_require(`
+		attribute systemd_generator;
+	')
+
+	type $1_t, systemd_generator;
+	type $1_exec_t;
+	init_daemon_domain($1_t, $1_exec_t)
+	init_nnp_daemon_domain($1_t)
+
+	kernel_read_system_state($1_t)
+	
+	dev_write_kmsg($1_t)
+
+	auth_use_nsswitch($1_t)
+	selinux_get_enforce_mode($1_t)
+
+	systemd_manage_unit_dirs($1_t)
+	systemd_create_unit_file_dirs($1_t)
+	systemd_create_unit_file_lnk($1_t)
+')
+
+######################################
+## <summary>
 ##      Create a domain for processes which are started 
 ##      exuting systemctl.
 ## </summary>
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -23,6 +23,7 @@ gen_tunable(systemd_socket_proxyd_connect_any, false)
 
 attribute systemd_unit_file_type;
 attribute systemd_domain;
+attribute systemd_generator;
 attribute systemctl_domain;
 attribute systemd_mount_directory;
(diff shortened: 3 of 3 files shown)


policy/modules/system/systemd.fc
```diff
@@ -2,8 +2,9 @@ HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_home
 HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit_file_t,s0)
 /root/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_home_t,s0)
 
-/etc/.*hostname.*			--		gen_context(system_u:object_r:hostname_etc_t,s0)
-/etc/machine-info		--		gen_context(system_u:object_r:hostname_etc_t,s0)
+/etc/.*hostname.*	--	gen_context(system_u:object_r:hostname_etc_t,s0)
+/etc/fstab		--	gen_context(system_u:object_r:systemd_fstab_etc_t,s0)
```
zpytela: Systemd does not own fstab, so we should avoid using a private label for the file which is used by numerous services.

policy/modules/system/systemd.fc
```diff
@@ -33,6 +34,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /usr/lib/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit_file_t,s0)
 /usr/lib/systemd/system/systemd-machined\.service	--	gen_context(system_u:object_r:systemd_machined_unit_file_t,s0)
 /usr/lib/systemd/system/systemd-networkd\.service     gen_context(system_u:object_r:systemd_networkd_unit_file_t,s0)
+/usr/lib/systemd/system/systemd-remount-fs\.service	gen_context(system_u:object_r:systemd_remount_fs_unit_file_t,s0)
```
zpytela: Why does this unit file need context different to default?

policy/modules/system/systemd.if
```diff
@@ -44,6 +44,23 @@ interface(`systemd_stub_unit_file',`
         ')
 ')
 
+########################################
+## <summary>
+##      Common permissions for systemd generators. (INTERNAL)
```
zpytela: If it is not meant for other services to use, it'd be better to have it as a template.
Perhaps a new template for generators can be used instead of systemd_domain_template.

policy/modules/system/systemd.te
```diff
@@ -1211,6 +1226,40 @@ optional_policy(`
 	udev_read_pid_files(systemd_gpt_generator_t)
 ')
 
+#######################################
+#
+# systemd_fstab_generator_t 
+#
+allow systemd_fstab_generator_t self:capability dac_override;
```
zpytela: I suppose this is only because /sys has 0555 permissions, at least I haven't found any other.

policy/modules/system/systemd.te
```diff
@@ -1211,6 +1226,40 @@ optional_policy(`
...
+
+read_files_pattern(systemd_fstab_generator_t, systemd_fstab_etc_t, systemd_fstab_etc_t)
+
+dev_write_sysfs_dirs(systemd_fstab_generator_t)
```
zpytela, line 1241: Have you saved the denial with all details? Does it really need write to /sys?

zpytela (commented):
Apart of the minor glitches looks find, the updated test runs without reporting an error.

policy/modules/system/systemd.fc
5umm3r15: you are right, fixed, thanks

policy/modules/system/systemd.if
5umm3r15: changed for template

policy/modules/system/systemd.te
5umm3r15: removed

policy/modules/system/systemd.te
5umm3r15: this is the denial: 
type=PROCTITLE msg=audit(10/04/2023 05:40:32.985:1156) : proctitle=/usr/lib/systemd/system-generators/systemd-fstab-generator /run/systemd/generator /run/systemd/generator.early /run/systemd/gene
type=PATH msg=audit(10/04/2023 05:40:32.985:1156) : item=0 name=/proc/self/fd/5 inode=1 dev=00:16 mode=dir,555 ouid=root ogid=root rdev=00:00 obj=system_u:object_r:sysfs_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(10/04/2023 05:40:32.985:1156) : cwd=/
type=SYSCALL msg=audit(10/04/2023 05:40:32.985:1156) : arch=x86_64 syscall=access success=no exit=EACCES(Permission denied) a0=0x7fff94b1a1e0 a1=W_OK a2=0x0 a3=0x0 items=1 ppid=13943 pid=13951 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=unset comm=systemd-fstab-g exe=/usr/lib/systemd/system-generators/systemd-fstab-generator subj=system_u:system_r:systemd_fstab_generator_t:s0 key=(null)
type=AVC msg=audit(10/04/2023 05:40:32.985:1156) : avc:  denied  { write } for  pid=13951 comm=systemd-fstab-g name=/ dev="sysfs" ino=1 scontext=system_u:system_r:systemd_fstab_generator_t:s0 tcontext=system_u:object_r:sysfs_t:s0 tclass=dir permissive=0

policy/modules/system/systemd.fc
5umm3r15: It was by mistake, it is removed now

zpytela:
Merging, thank you.

daandemeyer:
@zpytela Will this be backported to f39? mkosi can't properly relabel fedora rawhide images at the moment because when running unprivileged the kernel doesn't allow setfiles to set file contexts that are not in the host policy and these new contexts at the moment only seem to be available in rawhide.

zpytela:
@DaanDeMeyer As a new feature it was planned for rawhide only. If there is interest and the policy appears to be complete, we can consider it. Certainly there are plans to confine all generators, but this cannot be expected soon.
I probably don't understand why you ask.

daandemeyer:
@zpytela As part of building linux system images with mkosi, we relabel all files with the SELinux "setfiles" tool before packaging things up. Since this PR, I've started seeing the following on my fedora 39 system when building a rawhide image:

```
‣  Relabeling files using targeted policy
setfiles: Could not set context for /home/daandemeyer/.cache/mkosi-workspacei6d6m2vr/root/usr/lib/systemd/system-generators/systemd-fstab-generator:  Invalid argument
setfiles: Could not set context for /home/daandemeyer/.cache/mkosi-workspacei6d6m2vr/root/usr/lib/systemd/system-generators/systemd-rc-local-generator:  Invalid argument
setfiles: Could not set context for /home/daandemeyer/.cache/mkosi-workspacei6d6m2vr/root/usr/lib/systemd/system-generators/systemd-sysv-generator:  Invalid argument
```

So it seems it's only possible to build images if the corresponding contexts are also available on the host policy (which isn't the case for these three contexts). So it'd be nice to make these contexts available on F39, even if unused, so that the labels can be applied when building images.

zpytela:
Certainly there are some options, but it requires some more thoughts.
I also need to get familiar with mkosi. On a standard system, installed files get the correct context automatically.
I can also suggest another way - can you use `setfiles -c path` with rawhide policy?

daandemeyer:
@zpytela We already use setfiles with -c to point to the rawhide policy from the image. 

I think the main difference compared to all the other tools that use setfiles with -c is that mkosi runs unprivileged in a user namespace. When I run mkosi with sudo I don't get the setfiles errors.

This would indicate that only root can set arbitrary labels on files that are not in the host policy.

zpytela:
It is possible, there should be a SELINUX_ERR entry in audit log. So you cannot use sudo or some other way to work around?

daandemeyer:
> It is possible, there should be a SELINUX_ERR entry in audit log. So you cannot use sudo or some other way to work around?

Well, much of the appeal of mkosi comes from the fact that you can run it without needing root privileges. I've currently hacked in a workaround to allow ignoring the result of `setfiles` but that still means I get AVCs when I boot the image because the files are not labeled correctly.

zpytela:
@DaanDeMeyer I'd like to be of any help, but full backporting of new features is not planned and I don't think it is desired. Certainly it is possible to define just the types as you suggest, but such changes happen quite often and is going to happen during F40 devel cycle further on.
We had a small team discussion, we came up with the following suggestions:
- use sudo just for the one command
- define a small local policy which brings the types
- use the systemd module from rawhide
Loading the new policy directly also seems to be an option, but with less predictable results. Please let us know if some of these options are suitable for you.

daandemeyer:
@zpytela None of these are really an option unfortunately. But I understand that backporting these doesn't work for you either. Ideally unprivileged users would be allowed to set arbitrary labels that are not in the host policy. I'm already allowed to set labels that are in the host policy, so I'm unsure what the danger would be in setting labels that are not in the host policy. For now we'll simply to tell users that unprivileged image building with mkosi will have to be done from a host with all the required labels available.

biubiuzy:
This change causes problems with systemd, Why confine systemd generators?
See: https://github.com/systemd/systemd/issues/35723

PR #1893, opened 2023-10-06
Allow kernel_t to manage and relabel all files

WOnder93:
Extend the abitlity to manage all files also to character & block device files and also allow relabeling any file.

This is required for early boot overlay mounts to fully work, but may be needed for other legitimate oprations as well.

See also: https://github.com/ostreedev/ostree/pull/3062

### DIFF
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -1765,8 +1765,12 @@ interface(`files_relabel_all_files',`
 	relabel_chr_files_pattern($1, { file_type $2 }, { file_type $2 })
 
 	# satisfy the assertions:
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM, but I still believe unconfined_domain(kernel_t) should be the default.

zpytela:
Other file classes (lnk, sock, pipe) are not needed to be managed?

alexlarsson:
> Other file classes (lnk, sock, pipe) are not needed to be managed?

These are already allowed via files_manage_all_files(kernel_t) in kernel.te.

alexlarsson:
lgtm

cgwalters (approved):

PR #1900, opened 2023-10-11
Allow unconfined_domain_type use IORING_OP_URING_CMD on self

5umm3r15:

type=AVC msg=audit(1696506406.082:97818): avc:  denied  { cmd } for  pid=541430 comm="io_uring_passth" path="/dev/ng2n1" dev="devtmpfs" ino=745 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:fixed_disk_device_t:s0 tclass=io_uring permissive=0

Resolves: RHEL-11792

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -259,7 +259,7 @@ optional_policy(`
 
 # allow special io_uring features
 allow unconfined_domain_type domain:io_uring override_creds;
-allow unconfined_domain_type self:io_uring sqpoll;
+allow unconfined_domain_type self:io_uring { cmd };
 dev_io_uring_cmd_on_all_dev_nodes(unconfined_domain_type)
 files_io_uring_cmd_on_all_files(unconfined_domain_type)
 


WOnder93:
In the AVC the target type doesn't match the source type, so something is wrong here. Isn't the AVC already covered by the `dev_io_uring_cmd_on_all_dev_nodes(unconfined_domain_type)` line?

5umm3r15:
@WOnder93 thank you, you are right, in the commit message was wrong denial.

The fix is for audit2allow output:
allow unconfined_t self:io_uring cmd;

WOnder93:
Yeah, I asked in RHEL-11792 for the full denial, because it looks suspicious.

policy/modules/kernel/domain.te
```diff
@@ -259,7 +259,7 @@ optional_policy(`
 
 # allow special io_uring features
 allow unconfined_domain_type domain:io_uring override_creds;
-allow unconfined_domain_type self:io_uring sqpoll;
+allow unconfined_domain_type self:io_uring { cmd sqpoll };
```
WOnder93: So, based on my understanding the target context here can be any socket context, which is usually from the `domain` attribute. So in order to cover potential sockets inherited from other domains, we should allow `allow unconfined_domain_type domain:io_uring { cmd }` instead, i.e. add the `cmd` permission to the rule at line 261 instead of 262.

5umm3r15:
Thanks @WOnder93, fixed 
:)

WOnder93:
I think you misunderstood - I was proposing the following as the final result:
```
allow unconfined_domain_type domain:io_uring { cmd override_creds };
allow unconfined_domain_type self:io_uring sqpoll;
```

zpytela:
Thanks for your effort, closing in favor of https://github.com/fedora-selinux/selinux-policy/pull/1937

5umm3r15:
@zpytela thanks for update and sorry for not delivering fix on time

zpytela:
No problem @5umm3r15, looking forward to your future contributions.

PR #1902, opened 2023-10-16
Allow sysadm_t to connect to iscsid using a unix domain stream socket

naokitnk:
Need to allow sysadm_t to connect to iscsid using a unix domain stream socket as root user runs in sysadm_t domain when mls is used while it runs in unconfined_t domain with targeted policy.

### DIFF
--- a/policy/modules/roles/sysadm.te
+++ b/policy/modules/roles/sysadm.te
@@ -366,6 +366,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	iscsi_stream_connect(sysadm_t)
+')
+
+optional_policy(`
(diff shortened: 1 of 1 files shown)


zpytela:
@naokitnk For which command or scenario is this required?

naokitnk:
> @naokitnk For which command or scenario is this required?

When root user manually executes `iscsiadm` command with mls policy.

zpytela:
Can you please add the command name to the commit message?

naokitnk:
> Can you please add the command name to the commit message?

done

zpytela:
Merging, thank you.

PR #1913, opened 2023-10-19
Remove the lockdown-class rules from the policy

zpytela:
A comprehensive fix for all the problems caused by the lockdown SELinux class was rejected by Linus and for the lack of a better option, the consensus upstream was to just remove the class entirely and stop checking anything in the lockdown hook.

The lockdown class was removed from the upstream kernel in the "selinux: remove the SELinux lockdown implementation" commit and is not present any longer in Fedora 36 and newer.

    Related: rhbz#2017848

    Resolves: rhbz#2145266

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -3232,7 +3232,6 @@ interface(`dev_read_raw_memory',`
 
 	read_chr_files_pattern($1, device_t, memory_device_t)
 	allow $1 memory_device_t:chr_file map;
-	allow $1 self:lockdown integrity;
 
 	allow $1 self:capability sys_rawio;
 	typeattribute $1 memory_raw_read;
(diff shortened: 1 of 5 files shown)


martinpitt:
Sorry about the cockpit failure, that's a rare daytime related problem. I'll look into it ASAP, please ignore.

martinpitt:
(Or retry, I can't do that myself)

martinpitt:
FTR, I finally understand how rawhide is broken. Fix in https://github.com/cockpit-project/cockpit/pull/19551 , now just blocked by COPR being broken due to a *drumroll* lapsed SSL cert.

martinpitt:
Yes, yes -- I landed the fix in cockpit; if one of the repo owners could click the "re-run" link in https://github.com/fedora-selinux/selinux-policy/pull/1913/checks?check_run_id=18217347717 or do the above `/packit retest-failed`, it should go green.

martinpitt:
Argh, my bad, so sorry. Needs https://github.com/cockpit-project/cockpit/pull/19552

PR #1924, opened 2023-11-02
F39 backports

zpytela:


martinpitt:
This was merged totally red..

> make: *** No rule to make target 'coreos_installer.if', needed by 'tmp/all_interfaces.conf'.  Stop.

zpytela:
> This was merged totally red..
> 
> > make: *** No rule to make target 'coreos_installer.if', needed by 'tmp/all_interfaces.conf'.  Stop.

Correct. There seems to be a mixup of rawhide and f39 sources, coreos_installer is available in F40+.

PR #1930, opened 2023-11-08
Allow graphical applications work in Wayland

zpytela:
Resolves: rhbz#2248488

PR #1931, opened 2023-11-09
Label /var/run/auditd.state as auditd_var_run_t

JurajMarcin:
/var/run/auditd.state is used by auditd to dump its internal state upon user request. When created, it has correct type. However, when relabeled, its type changes to var_run_t and auditd cannot access it anymore.

Resolves: RHEL-14374

### DIFF
--- a/policy/modules/system/logging.fc
+++ b/policy/modules/system/logging.fc
@@ -76,6 +76,7 @@ ifdef(`distro_redhat',`
 /var/run/audit_events	-s	gen_context(system_u:object_r:auditd_var_run_t,mls_systemhigh)
 /var/run/audispd_events	-s	gen_context(system_u:object_r:audisp_var_run_t,mls_systemhigh)
 /var/run/auditd\.pid	--	gen_context(system_u:object_r:auditd_var_run_t,mls_systemhigh)
+/var/run/auditd\.state	--	gen_context(system_u:object_r:auditd_var_run_t,mls_systemhigh)
 /var/run/auditd_sock	-s	gen_context(system_u:object_r:auditd_var_run_t,mls_systemhigh)
 /var/run/klogd\.pid	--	gen_context(system_u:object_r:klogd_var_run_t,s0)
 /var/run/log		-s	gen_context(system_u:object_r:devlog_t,s0)


martinpitt:
The [f39 failure](https://artifacts.dev.testing-farm.io/9c3f3a32-4398-451d-a0cd-75ec915787b6/):
```
  Running scriptlet: selinux-policy-targeted-39.1-1.fc39.noarch            3/32 
Re-declaration of type virt_bridgehelper_t
Previous declaration of type at /var/lib/selinux/targeted/tmp/modules/100/virt_supplementary/cil:5
Bad type declaration at /var/lib/selinux/targeted/tmp/modules/100/virt_supplementary/cil:5
Failed to build AST
/usr/sbin/semodule:  Failed!
```
... is exactly the same as in #1930. #1927 succeeded 6 days ago, and #1929 even just two days ago. So something apparently broke in F39 proper very recently, with the libvirt 0.9.9 update? As this already fails to install packages, we can't just work around that by skipping a test.

zpytela:
> The [f39 failure](https://artifacts.dev.testing-farm.io/9c3f3a32-4398-451d-a0cd-75ec915787b6/):
> 
> ```
>   Running scriptlet: selinux-policy-targeted-39.1-1.fc39.noarch            3/32 
> Re-declaration of type virt_bridgehelper_t
> Previous declaration of type at /var/lib/selinux/targeted/tmp/modules/100/virt_supplementary/cil:5
> Bad type declaration at /var/lib/selinux/targeted/tmp/modules/100/virt_supplementary/cil:5
> Failed to build AST
> /usr/sbin/semodule:  Failed!
> ```
> 
> ... is exactly the same as in #1930. #1927 succeeded 6 days ago, and #1929 even just two days ago. So something apparently broke in F39 proper very recently, with the libvirt 0.9.9 update? As this already fails to install packages, we can't just work around that by skipping a test.

Latest rawhide commits may not work on older releases.

martinpitt:
> Latest rawhide commits may not work on older releases.

Good point! I suppose we should limit PRs targetting :rawhide to actual rawhide, and set up testing against the f39 branch?

martinpitt:
I  sent PR #1932 for that, this will stop the redness for rawhide branch. Looking into f39 now.

zpytela:
Merging, thank you, Juraj.

zpytela:
It *could* possibly useful to indicate the commit can be easily backported, but I guess it would have been needed to do different way.

PR #1950, opened 2023-12-01
Label /dev/vas with vas_device_t

zpytela:
Power9 processor introduced Virtual Accelerator Switchboard (VAS) which allows both userspace and kernel communicate to co-processor (hardware accelerator) referred to as the Nest Accelerator (NX). The NX unit comprises of one or more hardware engines or co-processor types such as 842 compression, GZIP compression and encryption. On power9, userspace applications will have access to only GZIP Compression engine which supports ZLIB and GZIP compression algorithms in the hardware. https://docs.kernel.org/arch/powerpc/vas-api.html

### DIFF
--- a/policy/modules/kernel/devices.fc
+++ b/policy/modules/kernel/devices.fc
@@ -152,6 +152,7 @@
 ifdef(`distro_suse', `
 /dev/usbscanner		-c	gen_context(system_u:object_r:scanner_device_t,s0)
 ')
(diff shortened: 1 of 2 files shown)


martinpitt:
Not your fault, sorry, it's a new regression on our side or in Rawhide proper. Retried, and reported to https://github.com/cockpit-project/cockpit/issues/19683 .

**Update**: ~Retry failed, please ignore this for now. We'll investigate it today.~ We landed the fix, retry worked.

PR #1953, opened 2023-12-04
Allow postfix_master_t map postfix data files

zpytela:
Nov 03 16:25:28 audit[294288]: AVC avc:  denied  { map } for  pid=294288 comm="postscreen" path="/var/lib/postfix/postscreen_cache.lmdb" dev="dm-1" ino=25200449 scontext=system_u:system_r:postfix_master_t:s0 tcontext=system_u:object_r:postfix_data_t:s0 tclass=file permissive=0

Resolves: rhbz#2247848


martinpitt:
Same "swap" issue. We just landed the adjustment to current rawhide, should succeed now. Retrying. Sorry for the noise!

PR #1962, opened 2023-12-12
Update qatlib policy

zpytela:
The commit addresses the following AVC denials:
type=PROCTITLE msg=audit(12.12.2023 09:16:26.851:299) : proctitle=/usr/sbin/qatmgr --policy=0 type=SYSCALL msg=audit(12.12.2023 09:16:26.851:299) : arch=x86_64 syscall=ioctl success=yes exit=0 a0=0x5 a1=0x3b69 a2=0x3 a3=0x70000000004 items=0 ppid=1 pid=63688 auid=unset uid=root gid=qat euid=root suid=root fsuid=root egid=qat sgid=qat fsgid=qat tty=(none) ses=unset comm=qatmgr exe=/usr/sbin/qatmgr subj=system_u:system_r:qatlib_t:s0 key=(null) type=AVC msg=audit(12.12.2023 09:16:26.851:299) : avc:  denied  { ioctl } for  pid=63688 comm=qatmgr path=/dev/vfio/465 dev="devtmpfs" ino=2030 ioctlcmd=0x3b69 scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:object_r:vfio_device_t:s0 tclass=chr_file permissive=1

type=PROCTITLE msg=audit(12.12.2023 09:16:31.459:301) : proctitle=/usr/sbin/qatmgr --policy=0
type=PATH msg=audit(12.12.2023 09:16:31.459:301) : item=1 name=(null) inode=3808 dev=00:18 mode=socket,660 ouid=root ogid=qat rdev=00:00 obj=system_u:object_r:var_run_t:s0 nametype=CREATE cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
type=PATH msg=audit(12.12.2023 09:16:31.459:301) : item=0 name=(null) inode=3633 dev=00:18 mode=dir,755 ouid=root ogid=qat rdev=00:00 obj=system_u:object_r:var_run_t:s0 nametype=PARENT cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0
[...]

PR #1963, opened 2023-12-12
Sudo admindomains

zpytela:


martinpitt:
Same as in #1962, something in rawhide broke RAID repair. Please ignore here.

PR #1964, opened 2023-12-12
Dontaudit virt-qemu-ga getattr all mountpoint

zpytela:


zpytela:
There already is
```
files_list_all_mountpoints(virt_qemu_ga_t)
files_write_all_mountpoints(virt_qemu_ga_t)
```
in rawhide, so it needs to be resolved individually in older releases.

PR #1979, opened 2023-12-18
Update systemd generators

zpytela:


martinpitt:
@zpytela Please ignore or retry the cockpit failure. That got fixed yesterday in https://github.com/cockpit-project/cockpit/pull/19853 Thanks!

PR #1995, opened 2024-01-13
Set correct label on /etc/pki/pki-tomcat/kra

yrro:
Allows certmonger to modify files within the kra subsystem directory as it can already do within the ca subsystem directory.

https://bugzilla.redhat.com/show_bug.cgi?id=2166441


zpytela:
Merging, thank you.

PR #2001, opened 2024-01-20
nova: Fix duplicate declarations

kajinamit:
lvm_domtrans is called twice with the exactly same usage.


martinpitt:
@zpytela The storage stack in rawhide exploded over the weekend *sigh*. We are investigating it today, in the meantime please ignore the cockpit failure.

zpytela:
LGTM

PR #2005, opened 2024-01-22
Allow smbd_t to watch user_home_dir_t if samba_enable_home_dirs is on

JurajMarcin:
When samba is allowed to access home directory, it sometimes needs to add watch on the whole home directory to respond to watch requests from clients (for example from Windows).

type=AVC msg=audit(1705933921.682:202): avc:  denied  { watch } for pid=11956 comm="smbd-notifyd" path="/home/test" dev="vda4" ino=912095 scontext=system_u:system_r:smbd_t:s0 tcontext=unconfined_u:object_r:user_home_dir_t:s0 tclass=dir permissive=0

Resolves: RHEL-14735


zpytela:
LGTM

PR #2008, opened 2024-01-24
Allow conntrackd_t to use bpf capability2

JurajMarcin:
When conntrackd filters packets using kernelspace filter, it needs the capability to do so efficiently.

Addresses following AVC denials:
type=AVC msg=audit(01/22/2024 12:46:49.999:248) : avc:  denied  { bpf } for  pid=1927 comm=conntrackd capability=bpf  scontext=system_u:system_r:conntrackd_t:s0 tcontext=system_u:system_r:conntrackd_t:s0 tclass=capability2 permissive=0

Resolves: RHEL-22277


zpytela:
LGTM

PR #2029, opened 2024-02-08
Fixes for domain_fd_use=0

WOnder93:
All of this needs to be allowed when the `domain_fd_use` boolean is turned off.

BTW, we should consider disabling the boolean by default instead of hiding these issues.


zpytela:
Merging, thank you. I will also add your note to the existing list of enhancements.

PR #2037, opened 2024-02-15
systemd: allow systemd_notify_t to send data to kernel_t datagram sockets

rmetrich:
This is required because of systemd's notify socket is created while in the initramfs, hence as `kernel_t`.
Once SELinux permits relabeling socket objects created before the policy is loaded, this should be removed and _systemd_ fixed to relabel the socket appropriately.
Tracked by [systemd PR](https://github.com/systemd/systemd/pull/31336).


zpytela:
Can you please include reproducer and/or the avc denials into the commit message?

zpytela:
Merging, thank you.

PR #2049, opened 2024-02-22
Allow userdomain get attributes of files on an nsfs filesystem

zpytela:
The commit addresses the following AVC denial:
type=PROCTITLE msg=audit(22.2.2024 23:42:03.003:742) : proctitle=pstree type=PATH msg=audit(22.2.2024 23:42:03.003:742) : item=0 name=/proc/4139/ns/cgroup inode=4026531835 dev=00:04 mode=file,444 ouid=root ogid=root rdev=00:00 obj=system_u:object_r:nsfs_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0 type=SYSCALL msg=audit(22.2.2024 23:42:03.003:742) : arch=x86_64 syscall=newfstatat success=yes exit=0 a0=AT_FDCWD a1=0x7ffc200aed00 a2=0x7ffc200aed40 a3=0x0 items=1 ppid=7788 pid=55219 auid=username uid=username gid=username euid=username suid=username fsuid=username egid=username sgid=username fsgid=username tty=pts13 ses=7 comm=pstree exe=/usr/bin/pstree subj=staff_u:staff_r:staff_t:s0-s0:c0.c1023 key=(null) type=AVC msg=audit(22.2.2024 23:42:03.003:742) : avc:  denied  { getattr } for  pid=55219 comm=pstree path=cgroup:[4026531835] dev="nsfs" ino=4026531835 scontext=staff_u:staff_r:staff_t:s0-s0:c0.c1023 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=1

PR #2050, opened 2024-02-23
policy: support pidfs

brauner:
pidfds are ported to a tiny in-kernel filesystem that is not mountable in userspace. This is comparable to sockfs, pipefs, nsfs, or anon_inodefs to name a few examples.

Before pidfs it wasn't possible for selinux to manage them because they didn't go through LSM hooks. We can now start doing that. But if we do it right away we'd get permission errors.

pidfds are used in systemd, dbus, LXC, polkit etc. and they currently start failing because selinux denies pidfs:

Feb 23 12:09:58 fed1 audit[353]: AVC avc:  denied  { read write open } for  pid=353 comm="systemd-userdbd" path="pidfd:[709]" dev="pidfs" ino=709 scontext=system_u:system_r:systemd_userdbd_t:>

### DIFF
--- a/policy/modules/kernel/filesystem.te
+++ b/policy/modules/kernel/filesystem.te
@@ -169,6 +169,11 @@ type oprofilefs_t;
 fs_type(oprofilefs_t)
 genfscon oprofilefs / gen_context(system_u:object_r:oprofilefs_t,s0)
 
+type pidfs_t;
+fs_type(pidfs_t)
+files_mountpoint(pidfs_t)
+genfscon pidfs_t / gen_context(system_u:object_r:pidfs_t,s0)
+
 type pstore_t alias pstorefs_t;
 fs_type(pstore_t)
 files_mountpoint(pstore_t)


brauner:
I lack the Selinux knowledge to fill in any potentially missing bits. So help would be very much appreciated!

brauner:
Fwiw, I also filed https://bugzilla.redhat.com/show_bug.cgi?id=2265630

brauner:
Closing this since I now filed a bugzilla entry.

martinpitt:
Rawhide failure is the same issue as in https://github.com/fedora-selinux/selinux-policy/pull/2049#issuecomment-1960826431

@brauner You said "closing this", but you actually didn't. Also, this change looks legit?

brauner:
> Rawhide failure is the same issue as in [#2049 (comment)](https://github.com/fedora-selinux/selinux-policy/pull/2049#issuecomment-1960826431)
> 
> @brauner You said "closing this", but you actually didn't. Also, this change looks legit?

Yeah, sorry. I didn't want to keep a review open in case that would be obviously wrong. If this looks ok the I'm happy if this is merged ofc.

WOnder93:
I believe we should do the same as [refpolicy](https://github.com/SELinuxProject/refpolicy/pull/762), i.e. use `fs_use_task` for this filesystem.

WOnder93:
I have a concern, though... if we do use `fs_use_task`, then from the policy POV the pidfs/pidfd operations will be indistinguishable from procfs file access (i.e. the access vectors will be just `<source domain> <target domain> file <perms>` with nothing identifying them as pidfs/pidfd operations.

So I think a better approach would be to add some pidfd-specific SELinux security class (with an appropriate set of permissions) and LSM hooks in the kernel and use a single label for the pidfs pseudofiles (allowing to grant/deny access to pidfs also at the filesystem level, although with less granularity; or it could even be universally allowed, leaving access control up to the new pidfd-specific SELinux permissions).

In fact, reading up on pidfd_open(2), it seems that labeling pidfds by the creating process doesn't actually gain much anyway, since the label would not refer to the process identified by the original PID, but merely to the process that has created the fd. So I retract my comment above - the best we can do with the current kernel situation is indeed to label everything `pidfs_t`, as the PR already does.

brauner:
Sounds good to me.

zpytela:
Merging, thank you.

PR #2063, opened 2024-03-15
dontaudit execmem for modemmanager

jsegitz:
Calls to g_regex_match_full can use a JIT to improve performance, but that requests execmem. Perfomance isn't priority here, so we don't allow this. Fixes
https://bugzilla.redhat.com/show_bug.cgi?id=2149946 https://bugzilla.suse.com/show_bug.cgi?id=1219363

### DIFF
--- a/policy/modules/contrib/modemmanager.te
+++ b/policy/modules/contrib/modemmanager.te
@@ -20,7 +20,7 @@ systemd_unit_file(modemmanager_unit_file_t)
 # Local policy
 #
 
-dontaudit modemmanager_t self:process { setpgid };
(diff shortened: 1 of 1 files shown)


zpytela:
Merging, thank you.

dreua:
Can we also have a backport of this on f39? It has the same issue and will be supported for another 6 month or so.

zpytela:
> Can we also have a backport of this on f39? It has the same issue and will be supported for another 6 month or so.

Slated to the next build: https://github.com/fedora-selinux/selinux-policy/pull/2053

PR #2066, opened 2024-03-18
Sudo dmidecode

zpytela:


zpytela:
Resolved in https://github.com/fedora-selinux/selinux-policy/pull/2118

PR #2083, opened 2024-04-15
Update the README.md file with the c9s branch information

zpytela:
The c9s branch was created in fedora-selinux/selinux-policy to allow contributors work on selinux policy updates for Centos 9 stream seamlessly.


zpytela:
container specfile reverts some incompatible changes for older systems, but this does not apply on branch level, so this branch needs to be updated to use for builds a centos stream system instead of fedora ones.

PR #2087, opened 2024-04-15
Fix bad Python regexp escapes

keszybz:
Python 3.12 started warning about such escapes. Use r"" to suppress the warning.


zpytela:
Merging, thank you.

PR #2109, opened 2024-05-03
Declare ports 80/udp and 443/udp as http for HTTP/3 (QUIC)

kaechele:
type=AVC msg=audit(1714750445.928:337): avc:  denied  { name_bind } for  pid=9687 comm="nginx" src=443 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:reserved_port_t:s0 tclass=udp_socket permissive=1

Resolves: rhbz#2278979


pomac303:
So, shouldn't 8080 and 8443 also be added since they are in the list?

kaechele:
No, because some browsers do not support HTTP/3 over ports that are unprivileged.

See here: https://groups.google.com/a/chromium.org/g/chromium-discuss/c/GRT5quemqBE/m/KILc5IeTBgAJ

pomac303:
Oh! interesting, the example on nginx.org uses 8443! Oh and i would find it very odd to do quic/http3 on port 80 udp - must be too late, since my comment was inspired by the change but also made me wonder "why 8080?!" ;)

pomac303:
(this one: https://nginx.org/en/docs/http/ngx_http_v3_module.html -- good thing i never even tested it like that)

kaechele:
> Oh and i would find it very odd to do quic/http3 on port 80 udp

I generally tend to agree and would be OK with removing that port from this proposal.
However, that would leave the user with no alternative port for testing without fiddling with the SELinux settings in some way though. But I don't how strong of a rationale that is.

Opinions loosely held on my part here ;-)

The only reason I added it was because I saw it being used like that in some examples online. But given the link you posted I don't think that argument holds water at this point ;-)

pomac303:
So the only thing that makes the change actually mean anything is adding the header - before that you can test with sites online or curl  (using --http3-only not currently packaged in fedora 40)

zpytela:
Resolved by https://github.com/fedora-selinux/selinux-policy/pull/2307

PR #2114, opened 2024-05-10
Label systemd configuration files with systemd_conf_t

zpytela:
The systemd_conf_t type was added as default file context for files with the .conf suffix and .conf.d directories in /etc/systemd, /run/systemd, and /usr/lib/systemd.
The /usr/local/lib/systemd directory is covered by file equivalency. The systemd_domain attribute was allowed read access to these files.

Refer to https://github.com/systemd/systemd/blob/main/NEWS CHANGES WITH 256-rc1:
General Changes and New Features:

        * Various programs will now attempt to load the main configuration file
          from locations below /usr/lib/, /usr/local/lib/, and /run/, not just
          below /etc/. For example, systemd-logind will look for
          /etc/systemd/logind.conf, /run/systemd/logind.conf,
          /usr/local/lib/systemd/logind.conf, and /usr/lib/systemd/logind.conf,
          and use the first file that is found.  This means that the search
          logic for the main config file and for drop-ins is now the same.

Resolves: rhbz#2279923

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -100,6 +100,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/systemd/inhibit(/.*)?	gen_context(system_u:object_r:systemd_logind_inhibit_var_run_t,s0)
 /run/systemd/ask-password-block(/.*)?	gen_context(system_u:object_r:systemd_passwd_var_run_t,s0)
 /run/systemd/ask-password(/.*)?	gen_context(system_u:object_r:systemd_passwd_var_run_t,s0)
+/run/systemd/machine(/.*)?	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/machines(/.*)?	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/machines.lock	--	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/resolve(/.*)?	gen_context(system_u:object_r:systemd_resolved_var_run_t,s0)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -4,6 +4,10 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 
 /etc/.*hostname.*	--	gen_context(system_u:object_r:hostname_etc_t,s0)
 /etc/machine-info	--	gen_context(system_u:object_r:hostname_etc_t,s0)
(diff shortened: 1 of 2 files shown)


martinpitt:
Interesting -- this breaks `systemctl disable --now dnf-automatic dnf-automatic-install`, which feels a little too specific to this PR to be a coincidence. @mvollmer @jelly can you please take over? I'm on my way out to PTO, sorry.

zpytela:
This PR was expected to have unpredictable impact, but I don't see any relation to dnf neither I can reproduce it. What I can confirm is some mess regarding dnf4 and dnf5 plugins in rawhide.

If you find any particular problem, please point me to it. I am not going to merge this PR right away.

AdamWill:
@martinpitt we landed the move from dnf 4 to dnf 5 by default in rawhide yesterday. it seems more likely that that would cause the problem than this PR?

martinpitt:
@AdamWill Ah! That would be it then, dnf5-automatic doesn't have the -install service/timer variant any more. Thanks!

policy/modules/system/systemd.fc
```diff
@@ -4,6 +4,8 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 
 /etc/.*hostname.*	--	gen_context(system_u:object_r:hostname_etc_t,s0)
 /etc/machine-info	--	gen_context(system_u:object_r:hostname_etc_t,s0)
+/etc/systemd/.*\.conf	--	gen_context(system_u:object_r:systemd_conf_t,s0)
+/etc/systemd/.*\.conf\.d(/.*)?	gen_context(system_u:object_r:systemd_conf_t,s0)
```
keszybz: I guess this could be changed to `/etc/systemd/.*\.conf\.d(/.*\.conf)?`, but it doesn't matter much.

keszybz (approved):
Looks all reasonable.

policy/modules/system/systemd.fc
zpytela: Can be, although it may be another punch to clarity. Can there be only plain files?
```
/etc/systemd/.+\.conf	--	gen_context(system_u:object_r:systemd_conf_t,s0)
/etc/systemd/.+\.conf\.d	-d	gen_context(system_u:object_r:systemd_conf_t,s0)
/etc/systemd/.+\.conf\.d/.+\.conf	--	gen_context(system_u:object_r:systemd_conf_t,s0)
```

keszybz: This version is nice.

> Can there be only plain files?

Only `*.conf` is read. They don't have to be "plain", a symlink would be accepted too by the systemd code.

zpytela: OK, updated. The level of complexity now looks like an overkill.
I will merge it later tomorrow. thanks for the review.

PR #2132, opened 2024-05-20
Create basic hg policy

maddymeows:
Experimenting with hg I found out hgweb did not work ootb, so I tried to write a policy. First time I ever actually wrote a policy, please tell me if there's anything wrong.

This makes hgweb work for anonymous viewing. Writing generally isn't done over http (opposed to ssh) thus not implemented. Very very very loosely based on the git policy.

Validated with CGI and WSGI on Apache.

dev_read_sysfs is required as it reads /sys/devices/system/cpu/possible.
When running as CGI, it also tried reading through httpd config and cgi-bin for whatever reason, so added those permissions as well. WSGI seems fine without it but I think it's running in the incorrect context (read on for why).

When browsing hgweb, I also get this denial, which I don't know what to do about:
```
avc: denied { getattr } for pid=99829 comm="httpd" path="pid:[4026531836]" dev="nsfs" ino=4026531836 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=0
```
scontext changes to be hg_script_t when running as CGI instead of WSGI. The server runs completely fine without it and I don't see any errors in the app logs, but I would prefer to get rid of this denial somehow.

That denial also made me question if the WSGI script is running correctly, as it's showing as httpd_t. Adding fs_getattr_nsfs_files(httpd_t) seems inappropriate.

How logging is handled feels dodgy, but I couldn't find a better way myself. When hgweb throws an uncaught exception it goes up several chains of logging which just ended up confusing me.

Repos by default would have to be placed in /var/lib/hg or /var/www/hg which is unusual. To push/pull/etc there you'd need to clone from ssh://host//var/lib/hg/reponame, with the double slash for an absolute path.

There's no real established convention on public content in the home directory like with public_html and public_git so I refrained from adding a default file context in home dir. But after adding a custom fcontext via semanage browsing user content works fine without the need for any other modifications.

Please let me know how I can improve this.

### DIFF
--- a/policy/modules/contrib/hg.te
+++ b/policy/modules/contrib/hg.te
@@ -1,0 +1,62 @@
+policy_module(hg, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+## <desc>
+##	<p>
+##	Determine whether Mercurial CGI
+##	can access cifs file systems.
+##	</p>
+## </desc>
+gen_tunable(hg_cgi_use_cifs, false)
+
+## <desc>
+##	<p>
+##	Determine whether Mercurial CGI
+##	can access nfs file systems.
+##	</p>
+## </desc>
+gen_tunable(hg_cgi_use_nfs, false)
+
+########################################
+#
+# CGI policy
+#
+
+optional_policy(`
+	apache_content_template(hg)
+	apache_content_alias_template(hg, hg)
+
+	list_dirs_pattern(hg_script_t, hg_content_t, hg_content_t)
+	read_files_pattern(hg_script_t, hg_content_t, hg_content_t)
+
+	auth_use_nsswitch(hg_script_t)
+
+	dev_read_sysfs(hg_script_t)
+
+	apache_search_config(hg_script_t)
+	apache_list_sys_scripts(hg_script_t)
+
+	kernel_dgram_send(hg_script_t)
+	optional_policy(`
+		abrt_stream_connect(hg_script_t)
+		logging_write_syslog_pid_socket(hg_script_t)
+	')
+
+	tunable_policy(`hg_cgi_use_cifs',`
+		fs_getattr_cifs(hg_script_t)
+		fs_read_cifs_files(hg_script_t)
+	',`
+		fs_dontaudit_read_cifs_files(hg_script_t)
+	')
+
+	tunable_policy(`hg_cgi_use_nfs',`
+		fs_getattr_nfs(hg_script_t)
+		fs_read_nfs_files(hg_script_t)
+	',`
+		fs_dontaudit_read_nfs_files(hg_script_t)
+	')
+')
--- a/policy/modules.conf
+++ b/policy/modules.conf
@@ -3155,3 +3155,10 @@ nvme_stas = module
 # coreos_installer
 #
 coreos_installer = module
+
+# Layer: contrib
+# Module: hg
+#
+#  hg - Mercurial source control management
+#
+hg = module
--- a/policy/modules/contrib/apache.if
+++ b/policy/modules/contrib/apache.if
@@ -1498,6 +1498,24 @@ interface(`apache_search_sys_scripts',`
 
 ########################################
 ## <summary>
+##	List apache system CGI directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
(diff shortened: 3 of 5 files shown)


policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+
+## <desc>
+##	<p>
+##	Determine whether Mercurial CGI
+##	can access nfs file systems.
+##	</p>
+## </desc>
+gen_tunable(hg_cgi_use_nfs, false)
+
+apache_content_template(hg)
```
zpytela: Other module's interfaces/templates should be called from an optional block.

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+list_dirs_pattern(hg_script_t, hg_content_t, hg_content_t)
+read_files_pattern(hg_script_t, hg_content_t, hg_content_t)
+
+auth_use_nsswitch(hg_script_t)
+
+gen_require(`
+	type httpd_config_t;
+	type httpd_sys_script_exec_t;
+')
+list_dirs_pattern(hg_script_t, httpd_config_t, httpd_sys_script_exec_t)
```
zpytela: Types from other modules should not be referred to directly, but using existing interfaces or together with creating new ones.

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+
+dev_read_sysfs(hg_script_t)
+
+kernel_dgram_send(hg_script_t)
```
zpytela: The recommended order of calling interfaces is 1. kernel 2. others from base 3. modules, alphabetical inside

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+abrt_stream_connect(hg_script_t)
```
zpytela: should be in an optional block

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+optional_policy(`
+	logging_write_syslog_pid_socket(hg_script_t)
+')
+
+tunable_policy(`hg_cgi_use_cifs',`
+	fs_getattr_cifs(hg_script_t)
+	fs_list_cifs(hg_script_t)
+	fs_read_cifs_files(hg_script_t)
```
zpytela: this seems to be a superset of the previous one

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+',`
+	fs_dontaudit_read_cifs_files(hg_script_t)
+')
+
+tunable_policy(`hg_cgi_use_nfs',`
+	fs_getattr_nfs(hg_script_t)
+	fs_list_nfs(hg_script_t)
+	fs_read_nfs_files(hg_script_t)
```
zpytela: same here

policy/modules/contrib/hg.te
```diff
@@ -0,0 +1,65 @@
...
+## <desc>
+##	<p>
+##	Determine whether Mercurial CGI
+##	can access nfs file systems.
+##	</p>
+## </desc>
+gen_tunable(hg_cgi_use_nfs, false)
+
+apache_content_template(hg)
+apache_content_alias_template(hg, hg)
```
zpytela: Are you sure you need also this template?

zpytela (commented):
I am quite surprised the policy is so brief, have you tried actually use it?

maddymeows:
> I am quite surprised the policy is so brief, have you tried actually use it?

I am using it on my server, though via selinux-policy-devel package. But after disabling it now I can't reproduce any of the errors I got when initially deploying it. So I'm unsure if it's still worth creating a policy for this.

I addressed most of the complaints anyhow

policy/modules/contrib/hg.te
maddymeows: Every other policy that uses apache_content_template followed with it, and it wasn't happy without it at the time I created the policy

policy/modules/contrib/hg.te
maddymeows: I don't quite understand this comment, for this block using 3 templates all seem to be in that order.

martinpitt:

PR #2136, opened 2024-05-27
Allow fstab generator dac_read_search capability

zpytela:
The dac_override capability has already been allowed, so it makes sense to allow dac_read_search, too, as the latter one is actually the first one checked by kernel.

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1273,7 +1273,7 @@ optional_policy(`
 #
 # systemd_fstab_generator_t 
 #
-allow systemd_fstab_generator_t self:capability dac_override;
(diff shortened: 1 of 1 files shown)


martinpitt:

martinpitt:
What an odyssey -- we [finally found the root cause](https://github.com/amazonlinux/amazon-ec2-utils/issues/37) and worked around it. @zpytela I restarted the failing tests in the recent PRs.

martinpitt:
ah dang -- needs #2138 first, as I also reorganized the test plans. After that lands, I'll re-run all the failed PRs.

zpytela:
Addressed by https://github.com/fedora-selinux/selinux-policy/pull/2136

PR #2137, opened 2024-05-27
Allow additional services read and write z90crypt device

zpytela:
This is a follow-up to the 01507d2fe7c6 ("Allow various services read and write z90crypt device) commit. In this batch, the following services were allowed r/w access to the crypt device:

- systemd-coredump

Resolves: rhbz#2283367


martinpitt:

PR #2152, opened 2024-06-06
Allow systemd (PID 1) manage systemd conf files

zpytela:
Denials are triggered when "systemctl set-property --runtime" is used.

type=AVC msg=audit(06/03/2024 05:40:08.117:543) : avc:  denied  { create } for  pid=1 comm=systemd name=.#50-CPUQuota.conf02880077955bce25 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:systemd_conf_t:s0 tclass=file permissive=0

Resolves: rhbz#2284157

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -569,6 +569,7 @@ optional_policy(`
     systemd_allow_mount_dir(init_t)
     systemd_allow_create_mount_dir(init_t)
     systemd_hostnamed_delete_config(init_t)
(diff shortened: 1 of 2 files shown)


martinpitt:
@zpytela something in rawhide broke ABRT and changed tuned. These are high on our list to investigate. Unfortunately there are a lot of other OS regressions on that list too, so it'll take a bit.. So please ignore here, the test looks fine.

zpytela:
Thanks, Martin.

PR #2183, opened 2024-06-18
Remove virt_use_glusterd tunable

zpytela:
Remove the declaration of the virt_use_glusterd
(Allow confined virtual guests to use glusterd) tunable and all references to it.
glusterd-selinux does not seem to refer to it either.

Resolves: RHEL-18975


zpytela:
Not removing, the content is not empty.

PR #2209, opened 2024-07-02
Dontaudit domain write cgroup files

zpytela:
This rule is added to prevent from reporting bugs since there currently is not a clear way how to address the problem how to allow domains write to
/sys/fs/cgroup/system.slice/servicename.service/memory.pressure, but not give the write access completely to the cgroup filesystem.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2294402

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -184,6 +184,7 @@ files_dontaudit_map_all_dirs(domain)
 files_dontaudit_execute_all_sockets(domain)
 
 fs_dontaudit_map_all_dirs(domain)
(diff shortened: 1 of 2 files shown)


martinpitt:
Unrelated cockpit failure. `reportd` got removed from Rawhide, so trying to install it now fails. I sent https://github.com/cockpit-project/cockpit/pull/20696 to adapt. After that lands, I'll re-run all the recent failures here.

martinpitt:
@zpytela The above PR landed, and I restarted all four recent affected selinux PRs. Sorry for the noise..

Issue #2214, opened 2024-07-03
Fedora 40:  It is not possible to register system using rhsm.service


jirihnidek:
When user tries to register system using client tool communicating over D-Bus with rhsm.service (e.g. `rhc`), then registration is terminated with following error in `rhsm.log`:

```
2024-07-03 12:42:28,228 [DEBUG] rhsm-service:575312:MainThread @client_info.py:75 - D-Bus sender: :1.2504 (cmd-line: ./rhc)
2024-07-03 12:42:28,228 [DEBUG] rhsm-service:575312:MainThread @i18n.py:166 - Trying to use locale: cs_CZ.UTF-8
2024-07-03 12:42:28,229 [INFO] rhsm-service:575312:MainThread @i18n.py:169 - Could not import locale for cs_CZ.UTF-8: [Errno 2] No translation file found for domain: 'rhsm'
2024-07-03 12:42:28,229 [INFO] rhsm-service:575312:MainThread @i18n.py:139 - Could not import locale either for cs_CS.UTF-8: [Errno 2] No translation file found for domain: 'rhsm'
2024-07-03 12:42:28,229 [DEBUG] rhsm-service:575312:MainThread @register.py:63 - Trying to create new domain socket server.
2024-07-03 12:42:28,229 [DEBUG] rhsm-service:575312:MainThread @server.py:329 - Adding sender :1.2504 to the set of senders
2024-07-03 12:42:28,230 [ERROR] rhsm-service:575312:MainThread @util.py:72 - org.freedesktop.DBus.Error.AccessDenied: Failed to bind socket "/run/dbus-He8ox0m8QT": Operace zamítnuta
Traceback (most recent call last):
  File "/usr/lib64/python3.12/site-packages/rhsmlib/dbus/util.py", line 70, in dbus_handle_exceptions
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/rhsmlib/dbus/objects/register.py", line 117, in Start
    address: str = self.impl.start(sender)
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/rhsmlib/dbus/objects/register.py", line 69, in start
    address: str = self.server.run()
                   ^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/rhsmlib/dbus/server.py", line 386, in run
[...]

PR #2215, opened 2024-07-03
Allow to create and delete socket files created by rhsm.service

jirihnidek:
* rhsm.service need to create socket file, when client tries to register system. SELinux blocked this.
* Issue is described in: https://github.com/fedora-selinux/selinux-policy/issues/2214
* Card ID: CCT-561

### DIFF
--- a/policy/modules/contrib/rhsmcertd.te
+++ b/policy/modules/contrib/rhsmcertd.te
@@ -47,6 +47,7 @@ dontaudit rhsmcertd_t self:process setfscreate;
 
 allow rhsmcertd_t self:fifo_file rw_fifo_file_perms;
 allow rhsmcertd_t self:unix_stream_socket create_stream_socket_perms;
+allow rhsmcertd_t rhsmcertd_var_run_t:sock_file manage_sock_file_perms;
 
 manage_dirs_pattern(rhsmcertd_t, rhsmcertd_config_t, rhsmcertd_config_t)
 manage_files_pattern(rhsmcertd_t, rhsmcertd_config_t, rhsmcertd_config_t)


policy/modules/contrib/rhsmcertd.te
```diff
@@ -47,6 +47,8 @@ dontaudit rhsmcertd_t self:process setfscreate;
 
 allow rhsmcertd_t self:fifo_file rw_fifo_file_perms;
 allow rhsmcertd_t self:unix_stream_socket create_stream_socket_perms;
+allow rhsmcertd_t rhsmcertd_var_run_t:sock_file create;
+allow rhsmcertd_t rhsmcertd_var_run_t:sock_file { setattr unlink };
```
jirihnidek: I created this patch in two steps at it is described in the issue. Could it be simplified to: `allow rhsmcertd_t rhsmcertd_var_run_t:sock_file { create setattr unlink };`?

zpytela: Yes, the result is the same. A different approach is to change it to manage_sock_file_perms or even using a pattern (see line 72): are you sure no other permission will be needed in the future? 

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt#L236
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/file_patterns.spt#L409

zpytela: I actually think this is a bug introduced by the commiter of 5b4cfb2d6dee9906ead5020303050518da3342c5

jirihnidek: We need only this ATM. We don't have plan to extend this part of D-Bus API. I will update my PR tomorrow.

jirihnidek: @zpytela Hi, updated PR as you suggested.

zpytela: LGTM
What I meant with my previous comments is that if I had noticed that, I'd have included manage_sock_files_pattern() in the https://github.com/fedora-selinux/selinux-policy/commit/5b4cfb2d6dee9906ead5020303050518da3342c5 commit.

zpytela:
Merging, thank you.

jirihnidek:
@zpytela Thanks for review and merging. Should I create any backport PR for Fedora or RHEL?

zpytela:
f40 build should be ready this week, for rhel we need a jira

PR #2225, opened 2024-07-10
Drop publicfile module to fix recent httpd avcs

keszybz:
This is not tested, because I have no idea how to do the Fedora CI IPA tests without creating a bodhi update…
The patch is based on speculation based on the AVC denial.

Should fixes the issue reported in
https://bodhi.fedoraproject.org/updates/FEDORA-2024-64134f8805#comment-3600824:

Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org systemd[1]: Starting httpd.service - The Apache HTTP Server...
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org ipa-httpd-kdcproxy[9981]: ipa: INFO: KDC proxy enabled
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org ipa-httpd-kdcproxy[9981]: ipa-httpd-kdcproxy: INFO     KDC proxy enabled
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org (httpd)[9984]: httpd.service: Referenced but unset environment variable evaluates to an empty string: OPTIONS
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org httpd[9984]: httpd: Could not open configuration file /etc/httpd/conf/httpd.conf: Permission denied
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org systemd[1]: httpd.service: Main process exited, code=exited, status=1/FAILURE
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org systemd[1]: httpd.service: Failed with result 'exit-code'.
Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org systemd[1]: Failed to start httpd.service - The Apache HTTP Server.

Jun 29 14:07:33 ipa001.test.openqa.fedoraproject.org audit[9984]: AVC avc:  denied  { search } for  pid=9984 comm="httpd" name="httpd" dev="dm-0" ino=353970 scontext=system_u:system_r:publicfile_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0

PR #2226, opened 2024-07-10
Do not label /usr/bin/httpd with publicfile_exec_t

zpytela:
After the bin-sbin merge, an entry for /usr/bin/httpd in publicfile.fc started to clash with another one in apache.fc.


zpytela:
Closing in favor of https://github.com/fedora-selinux/selinux-policy/pull/2225

PR #2234, opened 2024-07-15
Allow systemd_cryptsetup_generator_t to write kmsg

ca-hu:
Addresses:
```
type=AVC msg=audit(1721045161.675:334): avc:  denied  { write } for  pid=1662 comm="systemd-cryptse" name="kmsg" dev="devtmpfs" ino=10 scontext=system_u:system_r:systemd_cryptsetup_generator_t:s0 tcontext=system_u:object_r:kmsg_device_t:s0 tclass=chr_file permissive=0
```

Reproducer:
https://openqa.opensuse.org/tests/4341371#downloads
encrypt during install, then `systemctl daemon-reload`

it might be better to allow this for all generators? idk
see: https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html
`Since syslog(3) is not available (see above), log messages have to be written to /dev/kmsg instead.`

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1275,6 +1275,7 @@ allow systemd_generator systemd_unit_file_type:lnk_file getattr_lnk_file_perms;
 fs_getattr_cgroup(systemd_generator)
 fs_search_cgroup_dirs(systemd_generator)
 kernel_read_proc_files(systemd_generator)
+dev_write_kmsg(systemd_generator)
 
 ### Rules for individual systemd generator domains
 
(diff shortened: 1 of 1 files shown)


zpytela:
@ca-hu You are right, please apply the change to `systemd_generator`; note the interfaces order should be alphabetical, but kernel goes first.

ca-hu:
hope i understand the order correctly, if not let me know :)

zpytela:
purrfect

zpytela:
Merging, thank you.

PR #2239, opened 2024-07-15
Confine coreos generators

zpytela:
Confine coreos-boot-mount-generator, coreos-installer-generator, coreos-liveiso-autologin-generator, coreos-sulogin-force-generator system generators.

### DIFF
--- a/policy/modules/contrib/coreos_installer.fc
+++ b/policy/modules/contrib/coreos_installer.fc
@@ -3,7 +3,10 @@
 /usr/libexec/coreos-installer-disable-device-auto-activation	--	gen_context(system_u:object_r:coreos_installer_exec_t,s0)
 /usr/libexec/coreos-installer-service				--	gen_context(system_u:object_r:coreos_installer_exec_t,s0)
 
-/usr/lib/systemd/system-generators/coreos-installer-generator	--	gen_context(system_u:object_r:coreos_installer_exec_t,s0)
+/usr/lib/systemd/system-generators/coreos-boot-mount-generator	--	gen_context(system_u:object_r:coreos_boot_mount_generator_exec_t,s0)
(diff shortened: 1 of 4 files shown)


jlebon:
I'm a bit concerned with the tight coupling between these policies and the implementation details of the generators. Are we confining all the generators in Fedora at this level of granularity? Keeping up with all the churn there sounds like a lot of work. Would be good to have a Fedora Change for this too to raise visibility.

I worry about having to block on poking more holes in the policy whenever we add/remove/modify generators. I think ideally, the policy would live alongside the generators in https://github.com/coreos/fedora-coreos-config, but we are also not SELinux experts.

The benefits of confining generators are a little fuzzy to me given how early and privileged they are, though I'm OK with it in principle. I think what would help _a lot_ though is better CI integration between the two projects.

E.g. we could hook up CoreOS CI to this repo to make it easier to test changes and ensure minimal breakage. Does that interest you?

In Bodhi, we currently run CoreOS CI tests on every selinux-policy package update. They are currently not blocking but in future we'd like to propose that they'd be so (along with other packages).

zpytela:
This is a draft to collect denials. The resulting policy is expected be different.

zpytela:
> E.g. we could hook up CoreOS CI to this repo to make it easier to test changes and ensure minimal breakage. Does that interest you?
> In Bodhi, we currently run CoreOS CI tests on every selinux-policy package update. They are currently not blocking but in future we'd like to propose that they'd be so (along with other packages).

If it was aligned with current fmf workflow, we can give it a go.

jlebon:
> If it was aligned with current fmf workflow, we can give it a go.

It currently isn't, though that's definitely on my radar.

cgwalters:
Agreed, the value of isolating individual generators at a fine granularity seems really unlikely to be worth it.

PR #2251, opened 2024-07-19
Afterburn run and sshhome

zpytela:


policy/modules/contrib/afterburn.fc
```diff
@@ -1,3 +1,5 @@
+/run/metadata(/.*)?	gen_context(system_u:object_r:afterburn_runtime_t,s0)
```
aleskandro, line 1: will this allow `afterburn` to create the `/run/metadata` dir if not existent yet?

zpytela: It's actually L29. I can change it to contain the name.

Is it intentional to use /run/metadata as a name without any reference to the afterburn package?

aleskandro: > It's actually L29. I can change it to contain the name.

Ok, thanks for confirming.

> Is it intentional to use /run/metadata as a name without any reference to the afterburn package?

I'm not sure about this, but I think we should not consider other folders for now as the content in /run/metadata is consumed by several other software (e.g., some open shift components).

zpytela: It is okay as long as no other package decides to use the same path.

PR #2255, opened 2024-07-23
Update policy for coreos generators

zpytela:


martinpitt:
I hacked around the shadow-utils b0rkage in https://github.com/cockpit-project/cockpit/pull/20802 . Re-trying tests.

PR #2145, opened 2024-06-03
modules/contrib/bootupd: Remove module

travier:
Bootupd is no longer using a systemd service and socket unit.

It is now called either directly from the command line by an administrator or in the furture as part of the boot process in a oneshot unit.

See: https://github.com/coreos/bootupd/issues/551

PR #2257, opened 2024-07-23
Revert to permissive for coreos-installer & bootupd

travier:
Temporarily revert to permissive for those domains so that we have time to fix the issues that we missed earlier.

---

Revert "Remove permissive domain for bootupd_t"

This reverts commit 0cbc7da8130fd7cf030ab61f68a3eb449a8d6391.

---

Revert "Remove permissive domain for coreos_installer_t"

This reverts commit cd99e90ef9fc952ac9925a9f42befc9bd14ea5e4.

### DIFF
--- a/policy/modules/contrib/bootupd.te
+++ b/policy/modules/contrib/bootupd.te
@@ -15,6 +15,8 @@ systemd_unit_file(bootupd_unit_file_t)
 type bootupd_var_run_t;
 files_pid_file(bootupd_var_run_t)
 
+permissive bootupd_t;
+
 ########################################
 #
 # bootupd local policy
--- a/policy/modules/contrib/coreos_installer.te
+++ b/policy/modules/contrib/coreos_installer.te
@@ -39,6 +39,8 @@ files_type(coreos_sulogin_force_generator_unit_file_t)
 type coreos_installer_var_run_t;
 files_pid_file(coreos_installer_var_run_t)
 
+permissive coreos_installer_t;
+
 ########################################
 #
 # coreos_installer local policy


HuijingHei:
I have no objection for this, maybe should exclude `afterburn` as the issue https://bugzilla.redhat.com/show_bug.cgi?id=2254975 was already fixed, and there is no new bug for it.

But there are existing bugs for bootupd and coreos-installer:
- bootupd issue: https://bugzilla.redhat.com/show_bug.cgi?id=2300306
- coreos-installer issue: https://issues.redhat.com/browse/RHEL-38614

HuijingHei (approved):
LGTM

cgwalters:
For the record I still generally don't think it's a valuable effort to confine bootupd or coreos-installer.

jlebon:
@zpytela Can you take a look at this? Otherwise, we'll likely have to set them back to permissive downstream in the CoreOS configs until the associated bugs are fixed.

(Note also since this bug was opened, branching happened, so this needs to be fixed in both rawhide and f41.)

jlebon:
For coreos-installer, https://bugzilla.redhat.com/show_bug.cgi?id=2305385 is the one for Fedora.

travier:
> For the record I still generally don't think it's a valuable effort to confine bootupd or coreos-installer.

I agree as well that we should not confine those tools. I tried to argue that in:
- https://github.com/fedora-selinux/selinux-policy/pull/2145
- https://src.fedoraproject.org/rpms/selinux-policy/pull-request/431

@zpytela Could you point us to the exact rule mentioned in https://github.com/fedora-selinux/selinux-policy/pull/2145#issuecomment-2163122778 so that we could explain why this does not apply here / is not needed?

travier:
https://github.com/fedora-selinux/selinux-policy/pull/2336 should help with bootupd, but we still have https://github.com/coreos/fedora-coreos-tracker/issues/1771 blocking us on Fedora CoreOS so we still might need to make it permissive for F41.

zpytela:
The coreos installer generators are permissive since 2024-07-15, can I see some fresh report?

dustymabe:
> The coreos installer generators are permissive since 2024-07-15, can I see some fresh report?

https://bugzilla.redhat.com/show_bug.cgi?id=2305385#c6

travier:
We turned `bootupd_t` back to permissive for all Atomic Desktops (https://pagure.io/workstation-ostree-config/blob/main/f/bootupd.yaml#_34) until we can get https://github.com/fedora-selinux/selinux-policy/issues/2362 fixed.

We had to do it for Fedora CoreOS as well: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/selinux-workaround.yaml

It would be good to merge this one here until we get all the issues resolved again.

PR #2263, opened 2024-07-25
Sync modules.conf with Fedora targeted modules

bachradsusi:
https://src.fedoraproject.org/rpms/selinux-policy/blob/1f5817de9171791d37d4596e33d2f643e7c9d5a0/f/modules-targeted.conf

### DIFF
--- a/policy/modules.conf
+++ b/policy/modules.conf
@@ -434,13 +434,6 @@ aiccu = module
 # 
 aide = module
  
-# Layer: services
-# Module: ajaxterm
-#
-# Web Based Terminal
-# 
-ajaxterm = module
-
 # Layer: admin
 # Module: alsa
 #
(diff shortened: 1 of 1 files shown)


zpytela:
Merging, thank you.

XhmikosR:
@bachradsusi @zpytela will this be backported to the version Fedora 40 is using?

zpytela:
@XhmikosR Not planned, do you have any specific need for that?

XhmikosR:
I believe this PR fixes the common issue with GRD not working with SELinux
enforcing mode, see https://bugzilla.redhat.com/show_bug.cgi?id=2271661

Would be nice if it was included in Fedora 40 too. :)

bachradsusi:
This is just a sync of modules.conf from Fedora to github. It's not related to https://bugzilla.redhat.com/show_bug.cgi?id=2271661

https://bugzilla.redhat.com/show_bug.cgi?id=2271661#c16 is wrong. The right PR which should fix the bug is https://github.com/fedora-selinux/selinux-policy/pull/2269

XhmikosR:
Ah, sorry, I came here due the wrong Bugzilla link to this PR.

I'll follow #2269 then, thanks for the reply!

PR #2301, opened 2024-08-07
Add policy for /usr/libexec/samba/samba-bgqd

zpytela:
samba-bgqd is an helper program added in samba v4.20.2 to be spawned by smbd or spoolssd to perform jobs like updating the printer list or other management tasks asynchronously on demand. It is not intended to be called by users or administrators.

### DIFF
--- a/policy/modules/contrib/samba.fc
+++ b/policy/modules/contrib/samba.fc
@@ -19,6 +19,7 @@
 /usr/lib/systemd/system/winbind.*   --  gen_context(system_u:object_r:samba_unit_file_t,s0)
 
 /usr/libexec/samba/rpcd_lsad	--	gen_context(system_u:object_r:winbind_rpcd_exec_t,s0)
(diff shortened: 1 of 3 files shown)


cryptomilk:
By default samba-bgqd is started on demand by samba-dcerpcd in Samba.

samba-bgqd is a deamon enumerating cups printers and storing the gathered information in a cache. The issue is that it takes ~1 second to get the information about a printer. There are print servers out there with 20 or 60 printers. So enumerating 60 printers takes ~60sec. The problem is that if samba-bgqd is started on demand and enumerating printers is taking too long, the process will get killed before it is ready.

Providing a way to start samba-bgqd as a systemd service and always having it running in the background solves this. The problem is sometimes that we are a very very small team for a very big software project. And something there are a few thing we don't test thoroughly.

I'm fine with creating a policy for RHEL 9.6 for this. It is not urgent to fix the issue, but we have to.

PR #2307, opened 2024-08-08
Add port 80/udp and 443/udp to http_port_t defination

lch:
According to RFC9110[1] the port 80/udp and 443/udp has been assigned to http service in IANA registery[2]. Adding those ports to the SELinux port defination should allow the http/3 services to listen on the ports.

[1]: https://www.rfc-editor.org/rfc/rfc9110.html#name-port-registration
[2]: https://www.iana.org/assignments/service-names-port-numbers/


zpytela:
@lch Please fix the defination typo 2x and add missing commas.

lch:
Thanks. I have fixed them.

zpytela:
Merging, thank you.

PR #2316, opened 2024-08-20
Add missing sys_chroot capability to groupadd policy

pmatilai:
Just like useradd, groupadd supports chroot operation, but for that to work it needs the sys_chroot capability, just like the useradd policy has (see commit 0651cc950eae4fef8e631a61b36aa3f38e72ea12).

This is one of the roadblocks in making RPM sysusers feature fully functional: https://github.com/rpm-software-management/rpm/issues/3227

### DIFF
--- a/policy/modules/admin/usermanage.te
+++ b/policy/modules/admin/usermanage.te
@@ -212,7 +212,7 @@ optional_policy(`
 # Groupadd local policy
 #
 
-allow groupadd_t self:capability { dac_read_search dac_override chown fowner kill setuid sys_resource audit_write };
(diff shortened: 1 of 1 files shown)


martinpitt:
This F41 cockpit failure is due to a coreutils/gnulib regression, see https://bugzilla.redhat.com/show_bug.cgi?id=2307847 . We created a naughty pattern ("failure due to OS regression") for that in https://github.com/cockpit-project/bots/pull/6790 . This landed now, so the next run will be green again. You can retry or ignore here.

**Update**: I forgot that I can retry myself in this project, kicked.

zpytela:
Merging, thank you.

PR #2318, opened 2024-08-23
Allow rasdaemon write access to sysfs

ca-hu:
Error message in rasdaemon:
Aug 23 09:38:48 localhost rasdaemon[17117]: rasdaemon: Kernel does not support page offline interface

Fixes:
----
time->Fri Aug 23 09:38:48 2024
type=AVC msg=audit(1724398728.627:998): avc:  denied  { write } for  pid=17117 comm="rasdaemon" name="soft_offline_page" dev="sysfs" ino=46 scontext=system_u:system_r:rasdaemon_t:s0 tcontext=system_u:object_r:sysfs_t:s0 tclass=file permissive=0

see: https://bugzilla.suse.com/show_bug.cgi?id=1229587

### DIFF
--- a/policy/modules/contrib/rasdaemon.te
+++ b/policy/modules/contrib/rasdaemon.te
@@ -31,9 +31,9 @@ kernel_read_system_state(rasdaemon_t)
 kernel_manage_debugfs(rasdaemon_t)
 
 dev_read_raw_memory(rasdaemon_t)
-dev_read_sysfs(rasdaemon_t)
 dev_read_urand(rasdaemon_t)
 dev_rw_cpu_microcode(rasdaemon_t)
+dev_rw_sysfs(rasdaemon_t)
 
 corecmd_exec_bin(rasdaemon_t)
 


martinpitt:
These two failures are unrelated to this PR, we started to see them in cockpit PRs as well. Something else in F41/rawhide regressed.  Please ignore for now, will investigate on Monday.

martinpitt:
This F41/rawhide cockpit failure is due to a coreutils/gnulib regression, see https://bugzilla.redhat.com/show_bug.cgi?id=2307847 . We created a naughty pattern ("failure due to OS regression") for that in https://github.com/cockpit-project/bots/pull/6790 . This and https://github.com/cockpit-project/bots/pull/6793 landed now, so the next run will be green again. You can retry or ignore here.

**Update**: I forgot that I can retry myself in this project, kicked.

zpytela:
@ca-hu does the service need to write to one particular file?

ca-hu:
ah sorry, forgot to reply, yes i think it was /sys/devices/system/memory/soft_offline_page:
https://github.com/mchehab/rasdaemon/blob/master/ras-page-isolation.c#L67

zpytela:
If it is one or two files only, it'd be worth using genfscon. However, it's not in the policy for sysfs yet like it is for proc so it would require some effort.

ca-hu:
i think this one can be closed, the same permission was allowed in:
https://github.com/fedora-selinux/selinux-policy/commit/db38c8ba20adbda59b389111b9103274d1603a76

PR #2324, opened 2024-08-27
Use better escape method for "interface"

bachradsusi:
Fixes commit c7e8af00ef03 ("Escape "interface" as a file name in a virt filetrans pattern")

For a context see https://github.com/SELinuxProject/selint/pull/291 and https://lore.kernel.org/selinux/20240827113150.1843304-1-lautrbach@redhat.com/T/#u


zpytela:
LGTM

PR #2329, opened 2024-08-28
Allow systemd-ssh-generator to load net-pf-40

ca-hu:
see:
https://www.freedesktop.org/software/systemd/man/devel/systemd-ssh-generator.html "systemd-ssh-generator binds a socket-activated SSH server to local AF_VSOCK"

and modinfo suggests net-pf-40 to be the kernel modules for virtual sockets

Fixes:
> Aug 22 05:17:20 localhost kernel: audit: type=1400 audit(1724303839.663:5): avc:  denied  { module_request } for  pid=593 comm="systemd-ssh-gen" kmod="net-pf-40" scontext=system_u:system_r:systemd_ssh_generator_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=system permissive=0

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1366,6 +1366,7 @@ allow systemd_ssh_generator_t self:vsock_socket create;
 allow systemd_ssh_generator_t vsock_device_t:chr_file { read_chr_file_perms };
 
 dev_read_sysfs(systemd_ssh_generator_t)
+kernel_request_load_module(systemd_ssh_generator_t)
 
 optional_policy(`
 	ssh_domtrans(systemd_ssh_generator_t)


ca-hu:
see https://bugzilla.suse.com/show_bug.cgi?id=1229766

zpytela:
@ca-hu I have 2 questions:

- Do you use policy for F40 or was domain_kernel_load_modules turned off? In F41+ the permission is allowed by default for domain.

- Do you have any comprehensive test for ssh-generator? So far, we seem to add permissions as they a reported for one use case, it would help if some real user of this feature could check what else is needed, especially when vsock is involved.

policy/modules/system/systemd.te
```diff
@@ -1366,6 +1366,7 @@ allow systemd_ssh_generator_t self:vsock_socket create;
 allow systemd_ssh_generator_t vsock_device_t:chr_file { read_chr_file_perms };
 
 dev_read_sysfs(systemd_ssh_generator_t)
+kernel_request_load_module(systemd_ssh_generator_t)
```
zpytela, line 1369: Please note kernel is an exempt from alphabetical order and goes first.
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/systemd.te#L608

ca-hu:
> Do you use policy for F40 or was domain_kernel_load_modules turned off? In F41+ the permission is allowed by default for domain.

Ah okay, no we just saw this downstream in the opensuse policy, I think we can close this then. Thanks!

PR #2330, opened 2024-08-28
Allow init and systemd-logind to inherit fds from sshd

WOnder93:
This is required by systemd since commit 76f2191d8eb5 ("logind: introduce CreateSessionWithPIDFD()") when domain_fd_use is turned off.

Otherwise trying to SSH into the system will hang for two minutes until the timeout triggers a fallback and the SSH session is finally created.

---

I hit this when running selinux-testsuite, which flips `domain_fd_use` on and off. Please backport this also to F40 & F41 (F39 doesn't need it, as there is still an older systemd version).

### DIFF
--- a/policy/modules/services/ssh.if
+++ b/policy/modules/services/ssh.if
@@ -482,6 +482,24 @@ interface(`ssh_signull',`
 
 ########################################
 ## <summary>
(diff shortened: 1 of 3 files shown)


zpytela:
LGTM

PR #2340, opened 2024-09-03
Allow sysadm_t to create PF_KEY sockets

WOnder93:
This is needed to run selinux-testsuite as sysadm_t starting with: https://github.com/SELinuxProject/selinux-testsuite/commit/a9e631f0f1d5b11756a62679e8da073b3cc85b13


zpytela:
LGTM

PR #2342, opened 2024-09-04
Sync users with Fedora targeted users

bachradsusi:


zpytela:
I know it is out of scope of this PR, but I wonder what is the reason for associating staff_u with system_r, and why staff_u should be allowed to transition to unconfined_r.

bachradsusi:
To allow staff_u to use sudo to get unconfined_t - https://danwalsh.livejournal.com/66587.html

zpytela:
Merging, thank you. I hope I will get to the two issues later.

PR #2343, opened 2024-09-04
F40 backports

zpytela:


WOnder93:
Could you please also cherry-pick #2330 and #2340?

PR #2345, opened 2024-09-06
Dist git config

bachradsusi:
This is https://github.com/fedora-selinux/selinux-policy/pull/686 updated, rebased and slightly changed.


bachradsusi:
Related dist git PR https://src.fedoraproject.org/rpms/selinux-policy/pull-request/467

PR #2351, opened 2024-09-15
Define context for `.mylogin.cnf`

nisbet-hubbard:
`.mylogin.cnf` is created by `mysql_config_editor`. The file location is the current user's home directory: https://dev.mysql.com/doc/refman/9.0/en/mysql-config-editor.html.

Also opened upstream as https://github.com/SELinuxProject/refpolicy/pull/808.

### DIFF
--- a/policy/modules/contrib/mysql.fc
+++ b/policy/modules/contrib/mysql.fc
@@ -4,7 +4,9 @@
 # /HOME
 #
 HOME_DIR/\.my\.cnf -- gen_context(system_u:object_r:mysqld_home_t, s0)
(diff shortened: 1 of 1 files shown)


zpytela:
@nisbet-hubbard The fc entry is fine, but is there any service which actually changes the type?

nisbet-hubbard:
Oh yes, when you first create `.mylogin.cnf` using `mysql_config_editor`, the fc it gets is `admin_home_t`.

zpytela:
LGTM, but it needs to go to https://github.com/devexp-db/mysql-selinux

PR #2352, opened 2024-09-16
Fix defautlt and dist config files

bachradsusi:
- add gnome_remote_desktop to targeted modules.conf
- use symlinks to dist/targeted for default modules.conf and users files
- fix typo in users files


zpytela:
LGTM

Issue #2312, opened 2024-08-13
F40: Creation of debugfs entries in QAT driver blocked after starting qat service in QATlib


gcabiddu:
In Fedora 40 with any kernel version it has been noticed that most of the debugfs entries for the QAT driver are not present after starting the qat service.

The debugfs folder reports only the key dev_cfg which is created at the startup of the driver.

This is caused by SELinux blocking the creation of debugfs entries in the driver. If SELinux is in permissive mode or disabled, this issue does not occur.

Expected:
```
root@fedora:/sys/kernel/debug/qat_4xxx_0000:e8:00.0# ls
cnv_errors  dev_cfg  fw_counters  heartbeat  pm_status  telemetry  transport
```

Actual:
```
root@fedora:/sys/kernel/debug/qat_4xxx_0000:e8:00.0# ls
dev_cfg
```

zpytela:
@gcabiddu Can you share AVC denials?

gcabiddu:
@zpytela Is this sufficient?

```
# cat audit.log

type=AVC msg=audit(1723218438.213:294): avc:  denied  { search } for  pid=3379 comm="lspci" name=".cache" dev="nvme0n1p3" ino=151671 scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:object_r:cache_home_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1723218438.812:295): avc:  denied  { read } for  pid=3562 comm="chown" name="userdb" dev="tmpfs" ino=41 scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:object_r:systemd_userdbd_runtime_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1723218438.812:296): avc:  denied  { open } for  pid=3562 comm="chown" path="/run/systemd/userdb" dev="tmpfs" ino=41 scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:object_r:systemd_userdbd_runtime_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1723218438.812:297): avc:  denied  { getattr } for  pid=3562 comm="chown" path="/run/systemd/userdb" dev="tmpfs" ino=41 scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:object_r:systemd_userdbd_runtime_t:s0 tclass=dir permissive=1
[...]

PR #2355, opened 2024-09-17
Allow qatlib connect to systemd-machined over a unix socket

zpytela:
type=AVC msg=audit(1725307496.795:489): avc:  denied  { connectto } for  pid=5096 comm="chown" path="/run/systemd/userdb/io.systemd.Machine" scontext=system_u:system_r:qatlib_t:s0 tcontext=system_u:system_r:systemd_machined_t:s0 tclass=unix_stream_socket permissive=0


gcabiddu:
Confirmed that this resolves #2312

PR #2360, opened 2024-09-24
Synchronize with 'mysql-selinux' upstream

FaramosCZ:
Together with https://github.com/devexp-db/mysql-selinux/pull/7 it completely synchronizes the content mysql.* files in these two repositories.

### DIFF
--- a/policy/modules/contrib/mysql.fc
+++ b/policy/modules/contrib/mysql.fc
@@ -48,7 +48,7 @@ HOME_DIR/\.my\.cnf -- gen_context(system_u:object_r:mysqld_home_t, s0)
 # /var
 #
 /var/lib/mysql(-files|-keyring)?(/.*)?		gen_context(system_u:object_r:mysqld_db_t,s0)
-/var/lib/mysql/mysql\.sock -s	gen_context(system_u:object_r:mysqld_var_run_t,s0)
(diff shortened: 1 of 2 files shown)


zpytela:
Notes only partially related to this commit, but since we are here:

1. Rules for different domains should be separated, i. e. do not mix rules for mysqld_t and mysqld_safe_t.
2. Order of interfaces call should be kept, i. e. kernel is the first after manage rules and then alphabetically. Interfaces in optional blocks should be sorted alphabetically, too.
3. Is there a reason for having mariadb-backup?
4. Non-base interfaces should be in an optional block (sysnet, logging, auth, userdom, usermanage).

Anyway, we currently consider removing sources of modules like mysql and keep only interface files, mysql.if seems to be synced.

FaramosCZ:
I've prepared a https://github.com/devexp-db/mysql-selinux/pull/8 based on your feedback.
Please review them.
Once merged, I'll update this PR accordingly.

Regarding the `mariadb-backup`, the original [commit that introduced it ](https://github.com/devexp-db/mysql-selinux/commit/d4b14cddbb43234dace42bdcd51af57a27a005c6) states:
```
mariadb-backup needs to access raw files so give it the same context as the server.
```
Which IMO makes sense, it offers multiple backup methods, with some making physical backups of the raw DB files, so it needs to be permitted to work with them.

zpytela:
Regarding mariadb-backup: note the context only comes to play if it is executed as a systemd service.

PR #2372, opened 2024-10-03
update systemd-homed policy

zpytela:
Together with that, label /var/cache/systemd with systemd_cache_t.

Resolves: rhbz#2036108


zpytela:
@richiedaze FYI - it seems people started to use the feature.

richiedaze:
@zpytela 

Upgrading rawhide, updates systemd to version 256, which introduced new features to systemd-homed. 

I have adjusted in these new features on my testing branch, but I am still testing these features before submitting a pull request.

PR #2420, opened 2024-11-05
Label all semanage store files in /etc as semanage_store_t

bachradsusi:
/etc/selinux/targeted/tmp needs to have the same file context as /etc/selinux/targeted/active

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2323878


zpytela:
Merging, thank you.

PR #2444, opened 2024-11-22
Make bootupd_t permissive

zpytela:
The updated package does not use a service unit nor a socket to be started by systemd.

### DIFF
--- a/policy/modules/contrib/bootupd.te
+++ b/policy/modules/contrib/bootupd.te
@@ -8,6 +8,7 @@ policy_module(bootupd, 1.0.0)
 type bootupd_t;
 type bootupd_exec_t;
 init_daemon_domain(bootupd_t, bootupd_exec_t)
+permissive bootupd_t;
 
 type bootupd_unit_file_t;
 systemd_unit_file(bootupd_unit_file_t)


dustymabe:
https://github.com/coreos/bootupd/pull/663 is what dropped the use of systemd in the bootupd package.

travier:
Thanks. For reference:
- https://github.com/fedora-selinux/selinux-policy/pull/2145
- https://src.fedoraproject.org/rpms/selinux-policy/pull-request/431
- https://github.com/fedora-selinux/selinux-policy/pull/2257

travier:
I'm working on https://github.com/coreos/bootupd/pull/790 and we will need either this PR or more rules.

cgwalters:
Yes just to reiterate I don't think SELinux confinement for bootupd itself is providing much value now. Back when it was socket activated, it made some sense because while we restricted to root access, we don't necessarily want all code that happens to run as uid 0 to be able to initiate bootloader changes.

But because there's no socket there's no "privilege changes" going on; bootupd is already pretty careful to only be touching things that it should be touching and so we're left only with "accidental damage" protection from the policy, but that same policy is just cramping development.

cgwalters (approved):

dustymabe:
> Yes just to reiterate I don't think SELinux confinement for bootupd itself is providing much value now.

Yep. This should merge (I assume since @zpytela opened the PR). Just for history/context, we talked with @zpytela and the need (compliance) for policies for things only apply if they have systemd services.. even if they aren't daemons. Since https://github.com/coreos/bootupd/pull/663 dropped the use of those then we should be able to drop the policy for it (unless we add back something similar in the future).

HuijingHei:
Should we remove the bootupd module because bootupd is not a service anymore and we do not need the module?

dustymabe:
> Should we remove the bootupd module because bootupd is not a service anymore and we do not need the module?

Yes. My understanding is that this PR was an initial step in that direction. @zpytela should we merge this and open a new PR for dropping the module completely?

zpytela:
As long as there is a service, there needs to be SELinux support for it.

PR #2474, opened 2024-12-11
allow kdm to create /root/.kde/ with correct label

milosmalik:
When the kdm service is started, it wants to create the .kde directory under /root/, but SELinux denies that action. When the /root/.kde directory exists, the kdm service wants to create a symlink in it, but SELinux denies that action too. The intended symlink should point this way:
 * /root/.kde/cache-machine-<FQDN> --> /var/tmp/kdecache-root

The fix has 2 parts. First, SELinux policy should label the newly created /root/.kde directory correctly (xdm_home_t). Second, SELinux policy should allow the kdm initiated process to create a symlink in that directory.

Resolves: bz#2275868


zpytela:
Merging, thank you.

PR #2526, opened 2025-01-17
Allow virt_domain to use pulseaudio - conditional

vmojzis:
Introduce a new conditional virt_use_xserver which allows virtual guests to use pulseaudio.

The boolean is disabled by default, because it is unlikely for qemu:///system mode of libvirt to need this functionality (this is intended for qemu:///session mode of libvirt).

Resolves: RHEL-62763


zpytela:
LGTM

However, all tunable_policy() blocks should better be ordered before unconditional ones even though optional_policy() is used here as well.

PR #2528, opened 2025-01-21
Add the bind_exec_ndc() interface

zpytela:


zpytela:
Not needed after all.

PR #2541, opened 2025-01-27
Fix rshim policy

pgeng-nv:
rshim policy is out of date and causes rshim to be broken in RHEL 9.5 when SELinux is changed to enforcing mode.  This patch added the missing permissions.

Tested on RHEL 9.5.

### DIFF
--- a/policy/modules/contrib/rshim.fc
+++ b/policy/modules/contrib/rshim.fc
@@ -1,5 +1,5 @@
 /usr/bin/rshim				--	gen_context(system_u:object_r:rshim_exec_t,s0)
 
-/run/rshim\.pid	--	gen_context(system_u:object_r:rshim_var_run_t,s0)
+/var/run/rshim\.pid	--	gen_context(system_u:object_r:rshim_var_run_t,s0)
 
 /usr/lib/systemd/system/rshim.*		--	gen_context(system_u:object_r:rshim_unit_file_t,s0)
--- a/policy/modules/contrib/rshim.if
+++ b/policy/modules/contrib/rshim.if
@@ -1,39 +1,53 @@
-## <summary>policy for rshim</summary>
-
-########################################
-## <summary>
-##	Execute rshim_exec_t in the rshim domain.
-## </summary>
-## <param name="domain">
-## <summary>
-##	Domain allowed to transition.
-## </summary>
-## </param>
-#
-interface(`rshim_domtrans',`
-	gen_require(`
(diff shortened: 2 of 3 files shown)


zpytela:
While audit2allow can be used to create a workaround, it cannot be use without understanding to create a regular policy update.
Please modify the PR to: use interfaces to access types from other modules; use permission sets and patterns like read_file_pattern; the service should not be allowed to write to var_run_t; for powerful capabilities put justification either to commit message or source; similar for writing to sysfs.

With the -R switch audit2allow can help which interface to use, but it still may not be the best one.

pgeng-nv:
@zpytela:

* Used interfaces and macros to the extent as possible
* Provided justifications for the permissions that are needed. Note that rshim is a user space device driver for the BlueField Rshim hardware. That's why it needs more privileged accesses than many ordinary daemons.

zpytela:
There seems to be a misunderstanding: with "use interfaces to access types from other modules" I reallymeant only from other modules. Types from this module should be referred to directly, therefore there is no need for rshim_use_udp_socket etc.
On the other hand, instead of rshim_use_tty_chr_file you should use term_use_unallocated_ttys - this one makes the tty_device_t type accessible from other modules.

Can we now continue with smaller, incremental changes? e.g.
`allow $1 sysfs_t:file { read write open map };` can be replaced with `dev_map_sysfs` and `dev_read_sysfs` with `dev_rw_sysfs`

Is there an easy way to reproduce the denials?

pgeng-nv:
Hi @zpytela,

I have reworked the change on the latest Rawhide branch and addressed your comments:

- rshim-local permissions are kept directly in `rshim.te`.
- Interfaces are used for types owned by other modules.
- The policy retains the dedicated `rshim_var_run_t` type instead of allowing writes to generic `var_run_t`.
- The privileged capabilities and sysfs write access are documented.

I tested the change on Rocky Linux 9.7 using rshim 2.8.5 without its bundled SELinux workaround and a real BF-3 PCIe device passed through to the VM.

Without the updated policy, rshim worked in Permissive mode but failed under Enforcing mode. With the updated policy, rshim worked in both Permissive and Enforcing modes, attached to the BF-3 through UIO, and created the expected rshim devices and TMFIFO interface.

The current Packit failures appear unrelated: Fedora dist-git still applies the `Support console version of initial-setup` patch even though that change is already present in upstream Rawhide, causing the patch to fail during `%prep`.

PR #2548, opened 2025-01-29
Iio device

zpytela:


policy/modules/kernel/devices.fc
```diff
@@ -54,6 +54,7 @@
 /dev/hwrng		-c	gen_context(system_u:object_r:random_device_t,s0)
 /dev/gpiochip[0-9]+		-c	gen_context(system_u:object_r:gpio_device_t,s0)
 /dev/i915		-c	gen_context(system_u:object_r:dri_device_t,s0)
+/dev/iio:device[0-9]	-c	gen_context(system_u:object_r:iio_device_t,s0)
```
WOnder93: The number suffix should rather be `[0-9]+` - according to the kernel code it could range up to `INT_MAX` in theory.

zpytela: Thank you, funny thing is that I put the + into the commit message.

policy/modules/kernel/devices.te
```diff
@@ -164,6 +164,12 @@ dev_node(gpio_device_t)
 type hfi1_device_t;
 dev_node(hfi1_device_t)
 
+#
+# Type for /dev/iio:device[0-9]
```
WOnder93: Nit: it should be also updated here

zpytela: nit picked

WOnder93 (approved):

PR #2555, opened 2025-02-04
Allow smbd to load libcephfs proxy conditionally

anoopcs9:
The design for [libcephfs proxy](https://docs.ceph.com/en/latest/dev/libcephfs_proxy/) was accepted sometime back with the corresponding implementation subsequently [merged](https://github.com/ceph/ceph/pull/58376) into upstream Ceph. This additional way to integrate CephFS with Samba is comprised of two components. The daemon i.e, libcephfsd is responsible for creating the unix domain socket as a starting point of communication. On the other side proxy library, which is [designed](https://gitlab.com/samba-team/samba/-/merge_requests/3792) to be loaded within _smbd_, connects to the already created socket and thereby communicates with the daemon.

During the development and testing stages we happened to notice some AVC denial messages in the audit logs:

> type=AVC msg=audit(1742367267.967:6919): avc:  denied  { write } for  pid=62992
  comm="smbd[192.168.12" name="libcephfsd.sock" dev="tmpfs" ino=6049
  scontext=system_u:system_r:smbd_t:s0 tcontext=unconfined_u:object_r:var_run_t:s0
  tclass=sock_file permissive=1

> type=AVC msg=audit(1742367267.967:6919): avc:  denied  { connectto } for  pid=62992
  comm="smbd[192.168.12" path="/run/libcephfsd.sock" scontext=system_u:system_r:smbd_t:s0
  tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
  tclass=unix_stream_socket permissive=1

> type=AVC msg=audit(1742367267.970:6920): avc:  denied  { read write } for  pid=62992
  comm="smbd[192.168.12" path="socket:[189595]" dev="sockfs" ino=189595 scontext=system_u:system_r:smbd_t:s0
  tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
  tclass=unix_stream_socket permissive=1

Therefore we propose a new boolean - **_samba_load_libcephfs_proxy_**, off by default, to contain the necessary allow rules for _smbd_t_ to operate in a proxy mode using libcephfs.

### DIFF
--- a/policy/modules/contrib/samba.te
+++ b/policy/modules/contrib/samba.te
@@ -87,6 +87,13 @@ gen_tunable(samba_share_fusefs, false)
 ## </desc>
 gen_tunable(samba_load_libgfapi, false)
 
+## <desc>
+## <p>
+## Allow smbd to load libcephfs proxy from ceph.
+## </p>
+## </desc>
+gen_tunable(samba_load_libcephfs_proxy, false)
+
 type nmbd_t;
 type nmbd_exec_t;
 init_daemon_domain(nmbd_t, nmbd_exec_t)
@@ -586,6 +593,13 @@ tunable_policy(`samba_load_libgfapi',`
     corenet_tcp_connect_all_ports(smbd_t)
     corenet_tcp_bind_all_ports(smbd_t)
     corenet_sendrecv_all_packets(smbd_t)
+')
+
+optional_policy(`
+    tunable_policy(`samba_load_libcephfs_proxy',`
+        unconfined_stream_connect_rw(smbd_t)
+        rw_sock_files_pattern(smbd_t, var_run_t, var_run_t)
+    ')
 ')
 
 optional_policy(`
--- a/policy/modules/roles/unconfineduser.if
+++ b/policy/modules/roles/unconfineduser.if
@@ -424,6 +424,26 @@ interface(`unconfined_stream_connect',`
 
 ########################################
 ## <summary>
+##	Connect to the unconfined domain using
+##	a unix domain stream socket with
+##	additional read+write permissions.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`unconfined_stream_connect_rw',`
+	gen_require(`
+		type unconfined_t;
+	')
+
+	allow $1 unconfined_t:unix_stream_socket { connectto read write };
+')
+
+########################################
+## <summary>
 ##	Do not audit attempts to read or write
 ##	unconfined domain tcp sockets.
 ## </summary>


anoopcs9:
@zpytela I hope **rpm-build:fedora-41-x86_64** job failure is expected.

anoopcs9:
@zpytela Can you please take a look at this change?

policy/modules/contrib/samba.te
```diff
@@ -584,6 +591,13 @@ tunable_policy(`samba_load_libgfapi',`
     corenet_sendrecv_all_packets(smbd_t)
 ')
 
+optional_policy(`
+    tunable_policy(`samba_load_libcephfs_proxy',`
+        unconfined_stream_connect(smbd_t)
+        rw_sock_files_pattern(smbd_t, var_run_t, var_run_t)
```
zpytela, line 601: Which service creates and/or works with files with the `var_run_t` type? A private type should be used in cases of allowing write access.

anoopcs9: Proxy daemon, _libcephfsd_, creates the socket file for read/write under _/run_. It is a standalone daemon outside _smbd_. The proxy client library loaded by _smbd_ connects to the socket post the creation by _libcephfsd_.

zpytela: There is no service in libcephsd, which service starts/uses it?
Which filename is used?

Edit: If it was this one, the label should be smbd_var_run_t
command: systemctl set-environment LIBCEPHFSD_SOCKET=/run/samba/libcephfsd.sock

anoopcs9: The catch here is that the socket file created by _/usr/sbin/libcephfsd_ can be used by any application designed to work with the proxy client library. IOW, Samba may not be the only consumer of the socket file created by _/usr/sbin/libcephfsd_ and thus it creates under _/run_.

Currently there is no service responsible for launching the daemon(_libcephfsd_). It is supposed to be running before any consumer attempts to communicate via the socket file.

anoopcs9:
I've updated the PR with an additional commit(explaining the situation) as we discovered a hidden denial breaking the communication between proxy client and daemon since https://github.com/ceph/ceph/pull/61877.

anoopcs9:
@zpytela How do we plan to take this forward?

zpytela:
Services should be allowed write access to their private types, not to generic ones. If a socket file is created by some service, the service needs to be confined, too.

anoopcs9:
> Services should be allowed write access to their private types, not to generic ones. If a socket file is created by some service, the service needs to be confined, too.

Ok, this would mean we create the required types for ceph related services which is more involved (although I see the necessary [content](https://github.com/ceph/ceph/tree/main/selinux) in upstream). I'll come back once ceph contexts and types are in place.

PR #2559, opened 2025-02-06
Support peer-to-peer migration of vms using ssh

zpytela:


martinpitt:
Cockpit's testLvmOnLuks started failing in rawhide in general, some regression in the storage stack. Not specific to this PR.

PR #2578, opened 2025-02-18
Permit init_t (systemd) to start a detached screen/tmux session

jcpunk:
This policy tweak permits running a screen/tmux session via a system service/unit file.

Given the interactive nature of the process it might be worth putting behind a boolean...

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -2026,3 +2026,7 @@ optional_policy(`
         ccs_read_config(daemon)
     ')
  ')
+
+optional_policy(`
+	screen_exec(init_t)
+')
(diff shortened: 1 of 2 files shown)


policy/modules/system/init.te
```diff
@@ -2026,3 +2026,7 @@ optional_policy(`
         ccs_read_config(daemon)
     ')
  ')
+
+optional_policy(`
+	screen_exec(init_t)
```
zpytela, line 2031: This would effect in screen running in the init_t context, that does not feel right.

zpytela (commented):
Can you elaborate on what you are trying to do and why?

jcpunk:
I'm trying to utilize the tmux systemd unit in fedora : https://src.fedoraproject.org/rpms/tmux/blob/rawhide/f/tmux%40.service

When run under selinux enforcing it fails to start.  I was attempting to whip up a policy modification that permits running tmux via this unit file.

PR #2584, opened 2025-02-28
Allow afterburn to mount and read config drives

mdbooth:
On certain platforms (OpenStack and PowerVS are two), afterburn will
look for a config drive before falling back to the metadata service.
This will be a filesystem with a platform-specific well-defined
label. It is expected to be iso9660 filesystem on a virtual cd-rom
drive. afterburn will mount it in a temporary directory and read its
contents.

### DIFF
--- a/policy/modules/contrib/afterburn.te
+++ b/policy/modules/contrib/afterburn.te
@@ -15,6 +15,9 @@ systemd_unit_file(afterburn_unit_file_t)
 type afterburn_runtime_t;
 files_pid_file(afterburn_runtime_t)
 
+type afterburn_tmp_t;
+files_tmp_file(afterburn_tmp_t)
+
 ########################################
 #
 # afterburn local policy
@@ -28,6 +31,9 @@ allow afterburn_t afterburn_runtime_t:dir manage_dir_perms;
 manage_files_pattern(afterburn_t, afterburn_runtime_t, afterburn_runtime_t)
 files_pid_filetrans(afterburn_t, afterburn_runtime_t, dir, "metadata")
 
+allow afterburn_t afterburn_tmp_t:dir { manage_dir_perms mounton };
+files_tmp_filetrans(afterburn_t, afterburn_tmp_t, dir)
+
 kernel_dgram_send(afterburn_t)
 kernel_read_all_proc(afterburn_t)
 
@@ -36,6 +42,11 @@ corenet_tcp_connect_http_port(afterburn_t)
 domain_use_interactive_fds(afterburn_t)
 
 files_read_etc_files(afterburn_t)
+
+# Mount, unmount, and read config drives
+fs_mount_iso9660_fs(afterburn_t)
+fs_unmount_iso9660_fs(afterburn_t)
+fs_read_iso9660_files(afterburn_t)
 
 optional_policy(`
 	auth_use_nsswitch(afterburn_t)
@@ -60,4 +71,9 @@ optional_policy(`
 
 optional_policy(`
 	sysnet_dns_name_resolve(afterburn_t)
+')
+
+# afterburn invokes udevadm if mount fails
+optional_policy(`
+	udev_exec(afterburn_t)
 ')


policy/modules/contrib/afterburn.te
```diff
@@ -37,6 +40,18 @@ domain_use_interactive_fds(afterburn_t)
 
 files_read_etc_files(afterburn_t)
 
+manage_dirs_pattern(afterburn_t, afterburn_tmp_t, afterburn_tmp_t)
+manage_files_pattern(afterburn_t, afterburn_tmp_t, afterburn_tmp_t)
+files_tmp_filetrans(afterburn_t, afterburn_tmp_t, { file dir })
```
zpytela: I saw only a directory reported in /tmp, so a plain file can also be created?

zpytela: This section should be before kernel_, t.w. with the related mounton permission.

policy/modules/contrib/afterburn.te
```diff
@@ -37,6 +40,18 @@ domain_use_interactive_fds(afterburn_t)
...
+
+# Mount, unmount, and read config drives
+fs_mount_iso9660_fs(afterburn_t)
+fs_unmount_iso9660_fs(afterburn_t)
+fs_read_iso9660_files(afterburn_t)
+allow afterburn_t afterburn_tmp_t:dir { mounton };
+# afterburn invokes udevadm if mount fails
+udev_exec(afterburn_t)
```
zpytela: udev is a module, so this interface needs to be in an optional block at the file end.

zpytela (changes requested):
Generally ok. I also noticed some additional permissions were requested for the /run filesystem, not applicable any longer?

mdbooth:
I have tested this manually as follows:

```
# Create a rawhide server on OpenStack with a config drive
openstack server create "mbooth-config-drive-test" --flavor m1.medium --image Fedora-Cloud-Base-rawhide-latest --nic <snip> --config-drive True --key-name <snip> --wait

## On the OpenStack rawhide server
$ sudo dnf install -y 'dnf*-command(copr)'
$ sudo dnf copr enable packit/fedora-selinux-selinux-policy-2584 fedora-rawhide-x86_64

# Install afterburn and update to the new policy
$ sudo dnf install -y afterburn
$ sudo dnf update -y selinux-policy-targeted

# Manually edit the afterburn systemd unit to workaround lack of ignition parameters on the kernel command line
# This is necessary because this is regular rawhide, not FCOS
$ sudo sed -i 's,^ExecStart=.*,ExecStart=/usr/bin/afterburn --provider openstack --attributes=/run/metadata/afterburn,' /usr/lib/systemd/system/afterburn.service

# Start afterburn
$ sudo systemctl daemon-reload
$ sudo systemctl start afterburn

# Check afterburn status
$ sudo systemctl status afterburn
# We should see success and no log output:
Feb 28 16:58:07 mbooth-config-drive-test systemd[1]: Starting afterburn.service - Afterburn (Metadata)...
Feb 28 16:58:07 mbooth-config-drive-test systemd[1]: Finished afterburn.service - Afterburn (Metadata).
# Afterburn produces no output when reading from config drive. It logs when reading from metadata service.

# Verify that there were no denials in the audit log
$ sudo grep denied /var/log/audit/audit.log
```

policy/modules/contrib/afterburn.te
mdbooth: It's actually only `dir`. I added `file` out of an abundance of caution, but I can remove it and re-test.

mdbooth:
> Generally ok. I also noticed some additional permissions were requested for the /run filesystem, not applicable any longer?

The default afterburn unit creates the `/run/metadata` directory and the file `/run/metadata/afterburn` in it. That is already permitted by the existing policy, and we still need it.

policy/modules/contrib/afterburn.te
```diff
@@ -28,6 +31,11 @@ allow afterburn_t afterburn_runtime_t:dir manage_dir_perms;
 manage_files_pattern(afterburn_t, afterburn_runtime_t, afterburn_runtime_t)
 files_pid_filetrans(afterburn_t, afterburn_runtime_t, dir, "metadata")
 
+manage_dirs_pattern(afterburn_t, afterburn_tmp_t, afterburn_tmp_t)
```
zpytela: Is managing directories required on top of permissions added by files_tmp_filetrans? Will in /run/metadata be only plain files or also other file types - dir, symlink, pipe, socket, ...

mdbooth: In /run/metadata, only the directory itself and a single plain file. Is there a more restrictive macro?

mdbooth: We also create a directory in /tmp which we mount the config drive on.

zpytela: > In /run/metadata, only the directory itself and a single plain file. Is there a more restrictive macro?

You can remove this manage_dirs_pattern then.

mdbooth: I think I see where you're going with this. /run/metadata is handled by afterburn_runtime_t above.

I think I can just add 2 explicit create and rmdir perms for directories of afterburn_tmp_t. I'll give it a go.

zpytela: > We also create a directory in /tmp which we mount the config drive on.

This should already be allowed, not sufficiently? You can use this for a quick (and incomplete) check:
```
f42# macro-expander 'files_tmp_filetrans(afterburn_t, afterburn_tmp_t, dir)'
allow afterburn_t tmp_t:dir { open read getattr lock search ioctl add_name remove_name write };
```

mdbooth: ~~From observation it is not allowed, because this is the first thing denied by policy when we try to mount a config drive (it creates the mountpoint in /tmp first). We see this in the audit logs.~~

You mean it's added by files_tmp_filetrans. Gotcha. I'll remove it entirely.

mdbooth: Think we still need `create` and `rmdir`:

```
type=AVC msg=audit(1740764366.643:270): avc:  denied  { create } for  pid=1988 comm="afterburn" name="afterburn-k4uXUK" scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:object_r:afterburn_tmp_t:s0 tclass=dir permissive=0
```

I'll add just those.

mdbooth:
FYI, in local testing I reduced the tmp stanza to:
```
allow afterburn_t afterburn_tmp_t:dir { create mounton rmdir };
files_tmp_filetrans(afterburn_t, afterburn_tmp_t, dir)
```

The unit succeeded, but it failed to cleanup its temporary directory and I saw the following denial:
```
type=AVC msg=audit(1740764890.969:272): avc:  denied  { getattr } for  pid=1986 comm="afterburn" path="/tmp/afterburn-WomxOT" dev="tmpfs" ino=26 scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:object_r:afterburn_tmp_t:s0 tclass=dir permissive=0                                                                
```

I think I'll just change it to `mounton manage_dir_perms` if that's ok?

policy/modules/contrib/afterburn.te
mdbooth: (See top-level comment: also needed getattr, so I switched to manage_dir_perms)

mdbooth:
I've retested the latest push against rawhide and afterburn is working correctly with no denials.

mdbooth:
I did a naive backport to RHEL-9.6 here: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=66843980

I tested it on RHCOS. It appears to fail because it's trying to load `fs-iso9660` as a kernel module:
```
type=AVC msg=audit(1740773684.578:75): avc:  denied  { module_request } for  pid=1236 comm="afterburn" kmod="fs-iso9660" scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=system permissive=0
type=SYSCALL msg=audit(1740773684.578:75): arch=c000003e syscall=165 success=no exit=-19 a0=7ffe59844ba0 a1=7ffe59844fa0 a2=7ffe59844208 a3=1 items=0 ppid=1 pid=1236 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="afterburn" exe="/usr/bin/afterburn" subj=system_u:system_r:afterburn_t:s0 key=(null)^]ARCH=x86_64 SYSCALL=mount AUID="unset" UID="root" GID="root" EUID="root" SUID="root" FSUID="root" EGID="root" SGID="root" FSGID="root"
type=PROCTITLE msg=audit(1740773684.578:75): proctitle=2F7573722F62696E2F61667465726275726E002D2D636D646C696E65002D2D617474726962757465733D2F72756E2F6D657461646174612F61667465726275726E
```

The udev fallback when the mount fails also seems to have different behaviour. I'm seeing:
```
type=AVC msg=audit(1740773684.584:76): avc:  denied  { create } for  pid=1238 comm="udevadm" scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:system_r:afterburn_t:s0 tclass=netlink_selinux_socket permissive=0
type=SYSCALL msg=audit(1740773684.584:76): arch=c000003e syscall=41 success=no exit=-13 a0=10 a1=80003 a2=7 a3=0 items=0 ppid=1236 pid=1238 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="udevadm" exe="/usr/bin/udevadm" subj=system_u:system_r:afterburn_t:s0 key=(null)^]ARCH=x86_64 SYSCALL=socket AUID="unset" UID="root" GID="root" EUID="root" SUID="root" FSUID="root" EGID="root" SGID="root" FSGID="root"
type=PROCTITLE msg=audit(1740773684.584:76): proctitle=7564657661646D00736574746C65
type=AVC msg=audit(1740773684.585:77): avc:  denied  { getattr } for  pid=1238 comm="udevadm" name="/" dev="cgroup2" ino=1 scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:object_r:cgroup_t:s0 tclass=filesystem permissive=0
type=SYSCALL msg=audit(1740773684.585:77): arch=c000003e syscall=137 success=no exit=-13 a0=7f2304ebbbf9 a1=7ffef1ed03b0 a2=7f2303da3560 a3=0 items=0 ppid=1236 pid=1238 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="udevadm" exe="/usr/bin/udevadm" subj=system_u:system_r:afterburn_t:s0 key=(null)^]ARCH=x86_64 SYSCALL=statfs AUID="unset" UID="root" GID="root" EUID="root" SUID="root" FSUID="root" EGID="root" SGID="root" FSGID="root"
type=PROCTITLE msg=audit(1740773684.585:77): proctitle=7564657661646D00736574746C65
type=AVC msg=audit(1740773684.586:78): avc:  denied  { write } for  pid=1238 comm="udevadm" name="control" dev="tmpfs" ino=524 scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:object_r:udev_var_run_t:s0 tclass=sock_file permissive=0
type=SYSCALL msg=audit(1740773684.586:78): arch=c000003e syscall=42 success=no exit=-13 a0=3 a1=5595aa983f40 a2=14 a3=0 items=0 ppid=1236 pid=1238 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="udevadm" exe="/usr/bin/udevadm" subj=system_u:system_r:afterburn_t:s0 key=(null)^]ARCH=x86_64 SYSCALL=connect AUID="unset" UID="root" GID="root" EUID="root" SUID="root" FSUID="root" EGID="root" SGID="root" FSGID="root"
type=PROCTITLE msg=audit(1740773684.586:78): proctitle=7564657661646D00736574746C65
```

mdbooth:
We *could* grant afterburn `module_request`, I guess, but if we modify the systemd unit to modprobe `fs-iso9660` before running afterburn it doesn't need it.

@zpytela do we have a policy on this?

mdbooth:
I created a drop-in `/etc/systemd/system/afterburn.service.d/00-isofs.conf`:
```
[Unit]
Wants=modprobe@isofs.service
After=modprobe@isofs.service
```
With this in place, systemd loads isofs before starting afterburn, and everything works correctly (on RHCOS).

We can ensure that this drop-in is created on platforms which might use a config drive. This isn't 100% ideal because it means we're always loading a kernel module which isn't always required, but I don't see this as a blocker.

mdbooth:
I see `kernel_dontaudit_request_load_module` is actually quite common. I'm inclined to add it here to avoid the additional complication in systemd. Would that be ok?

mdbooth:
I added it as a separate commit.

zpytela:
> > Generally ok. I also noticed some additional permissions were requested for the /run filesystem, not applicable any longer?
> 
> The default afterburn unit creates the `/run/metadata` directory and the file `/run/metadata/afterburn` in it. That is already permitted by the existing policy, and we still need it.

What I meant was reported for RHEL 9 where the policy may differ. Let's complete the policy in rawhide first.

zpytela:
> I created a drop-in `/etc/systemd/system/afterburn.service.d/00-isofs.conf`:
> 
> ```
> [Unit]
> Wants=modprobe@isofs.service
> After=modprobe@isofs.service
> ```
> 
> With this in place, systemd loads isofs before starting afterburn, and everything works correctly (on RHCOS).
> 
> We can ensure that this drop-in is created on platforms which might use a config drive. This isn't 100% ideal because it means we're always loading a kernel module which isn't always required, but I don't see this as a blocker.

From my point of view this is a better approach, but I cannot assess your arguments. You should not need the dontaudit rule then.

Will this work well also when the service is restarted, or if it fails an some temporary files disappear, e.g. as en effect of systemd-tmpfiles?

zpytela:
> I see `kernel_dontaudit_request_load_module` is actually quite common. I'm inclined to add it here to avoid the additional complication in systemd. Would that be ok?

Let's clarify first what it means: the service itself will not be allowed to load a kernel module. In case of any intervention, like admin's rmmod, and the module is needed, the only way would be to restart the service.

policy/modules/contrib/afterburn.te
```diff
@@ -36,6 +36,7 @@ files_tmp_filetrans(afterburn_t, afterburn_tmp_t, dir)
 
 kernel_dgram_send(afterburn_t)
 kernel_read_all_proc(afterburn_t)
+kernel_dontaudit_request_load_module
```
zpytela: The domain is missing here.

policy/modules/contrib/afterburn.te
zpytela: Now I am actually surprised that no rules for files are needed.

Edit: so this explains it I guess. "afterburn will mount it in a temporary directory and read its
contents". Where the content does come from?

You may now reconsider the manage_dir_perms:
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt#L145
perhaps replace it with rw_dir_perms?

But to understand me better: on one hand the policy should be as tight as possible, otherwise it would make little sense even to have any. On the other hand there are these permission sets, patterns, and interfaces to ease the maintenance and to add even more permissions when it is reasonable to expect them be requested in the future, or when the code path is different, etc. In such cases it is rather the component maintainer to decide.

policy/modules/contrib/afterburn.te
mdbooth: 🤦

Incidentally, the only reason I was using dontaudit here is because I hadn't found the non-dontaudit version of this macro. I see there is one, so I'll use it.

mdbooth:
> > I see `kernel_dontaudit_request_load_module` is actually quite common. I'm inclined to add it here to avoid the additional complication in systemd. Would that be ok?
> 
> Let's clarify first what it means: the service itself will not be allowed to load a kernel module. In case of any intervention, like admin's rmmod, and the module is needed, the only way would be to restart the service.

The denial when the module was not loaded was:
```
type=AVC msg=audit(1740773684.578:75): avc:  denied  { module_request } for  pid=1236 comm="afterburn" kmod="fs-iso9660" scontext=system_u:system_r:afterburn_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=system permissive=0
type=SYSCALL msg=audit(1740773684.578:75): arch=c000003e syscall=165 success=no exit=-19 a0=7ffe59844ba0 a1=7ffe59844fa0 a2=7ffe59844208 a3=1 items=0 ppid=1 pid=1236 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="afterburn" exe="/usr/bin/afterburn" subj=system_u:system_r:afterburn_t:s0 key=(null)^]ARCH=x86_64 SYSCALL=mount AUID="unset" UID="root" GID="root" EUID="root" SUID="root" FSUID="root" EGID="root" SGID="root" FSGID="root"
type=PROCTITLE msg=audit(1740773684.578:75): proctitle=2F7573722F62696E2F61667465726275726E002D2D636D646C696E65002D2D617474726962757465733D2F72756E2F6D657461646174612F61667465726275726E
```

So SELinux is 'accounting' the loading of the module to `afterburn_t`, even though afterburn is not directly loading this module.

Although there is a workaround here by having systemd load it explicitly, I feel like relying on this would create a fragile coupling between afterburn, its SELinux policy, and the systemd units which invoke it. I note, for example, that this only worked on rawhide because something in the default boot configuration has already triggered the loading of the isofs module. Some reconfiguration of some unknown other boot unit would cause this to also fail on rawhide. Also, the module only actually needs to be loaded for config drives which is uncommon on OpenStack. We aren't going to be able to determine this from systemd, at least not without a fragile duplication of afterburn logic in a systemd unit, so we would always have to load the module despite it usually not being required.

Dynamic module loading is an OS feature, and this would mean disabling it for afterburn. Now I've discovered the `kernel_request_load_module` macro I see a huge number of modules in contrib using it, presumably also because anticipating what modules might need to be loaded is unmaintainable. I'd be more comfortable following this pattern.

mdbooth:
I've fixed the `request_load_module` call and updated it not to use the dontaudit version. I've also squashed that change into the main commit because, as discussed above, I think this is a core behaviour of afterburn.

I'm just re-testing it locally on rawhide (which I obviously hadn't done on this morning's earlier commit, sorry!).

EDIT: I retested this successfully on rawhide. I also manually removed the `isofs` kernel module before running afterburn and confirmed that it was dynamically reloaded.

zpytela:
One more note: The ability to load a module is affected by the state of `domain_kernel_load_modules` tunable. It is enabled in Fedora and RHEL 10, but disabled in RHEL 9. So in a default system state you should need this commit only in RHEL 9. The next question is if it should be a subject of disabling. In the current policy, some of the explicit calls of `kernel_request_load_module` are justified, i.e. the domain should have the permission even when the tunable is disabled, some are only for historical reasons, but it is now quite troublesome to remove them when the range of use cases is not known.

mdbooth:
> One more note: The ability to load a module is affected by the state of `domain_kernel_load_modules` tunable. It is enabled in Fedora and RHEL 10, but disabled in RHEL 9. So in a default system state you should need this commit only in RHEL 9. The next question is if it should be a subject of disabling. In the current policy, some of the explicit calls of `kernel_request_load_module` are justified, i.e. the domain should have the permission even when the tunable is disabled, some are only for historical reasons, but it is now quite troublesome to remove them when the range of use cases is not known.

If I rephrase for my own understanding: in both RHEL 9 & 10 default allow kernel module loading is controlled by a tunable, but the default value of that tunable is different.

To make a judgement on whether to include or not include this I think I'd need to read the design discussion on this tunable. For example, I might guess that we switched it from false to true to dissuade policies from adding `kernel_request_load_module` and therefore give users the ability to strictly control which modules may be loaded? If that's the case, I'd say that we should not include it in the rawhide patch, but add it in the c9s backport? Either way, I think in the default configuration we want to allow afterburn to load whatever kernel modules it needs to avoid the tight coupling mentioned above.

Incidentally, I noticed that most of the existing policies are not well commented, if commented at all. That has dissuaded me from adding more comments as they looked out of place. You mentioning how it's troublesome to remove them suggests that perhaps you'd like more comments?

zpytela:
> If I rephrase for my own understanding: in both RHEL 9 & 10 default allow kernel module loading is controlled by a tunable, but the default value of that tunable is different.

Correct.

> 
> To make a judgement on whether to include or not include this I think I'd need to read the design discussion on this tunable. For example, I might guess that we switched it from false to true to dissuade policies from adding `kernel_request_load_module` and therefore give users the ability to strictly control which modules may be loaded? If that's the case, I'd say that we should not include it in the rawhide patch, but add it in the c9s backport?

That'd be my first suggestion.

> Either way, I think in the default configuration we want to allow afterburn to load whatever kernel modules it needs to avoid the tight coupling mentioned above.

Yes.

> 
> Incidentally, I noticed that most of the existing policies are not well commented, if commented at all. That has dissuaded me from adding more comments as they looked out of place. You mentioning how it's troublesome to remove them suggests that perhaps you'd like more comments?

A verbose commit message is sufficient. In-line comments are good for extraordinary cases or for powerful capabilities justification.

mdbooth:
Thanks for the explanation. I've removed `kernel_request_load_module`. I retested against rawhide with the `isofs` kernel module manually removed and confirmed that afterburn loaded it again without the requirement for `kernel_request_load_module` as you said.

I'll add it again with a note about why for the backport.

mdbooth:
I've been ignoring the failure to build for F41 as it doesn't look related, btw. I'm assuming that's not something I need to address? Let me know if that's not correct.

zpytela:
our f41 build currently does not work

zpytela:
Merging now, thank you for you cooperation and patience in fine-tuning.

PR #2585, opened 2025-03-03
Allow afterburn to mount and read config drives

mdbooth:
On certain platforms (OpenStack and PowerVS are two), afterburn will
look for a config drive before falling back to the metadata service.
This will be a filesystem with a platform-specific well-defined
label. It is expected to be iso9660 filesystem on a virtual cd-rom
drive. afterburn will mount it in a temporary directory and read its
contents.

RHEL-9 only: Allows afterburn to load kernel modules. Required because
afterburn may load the isofs module to read config drives, and
domain_kernel_load_modules defaults to false in RHEL 9.

Resolves: RHEL-79319


mdbooth:
This is a backport of https://github.com/fedora-selinux/selinux-policy/pull/2584

PR #2591, opened 2025-03-07
Allow insights_core create/unlink/setattr symbolic and character device files

xiangce:
The rules suggested by `audit2allow` tool are

```
#============= insights_core_t ==============
allow insights_core_t insights_core_tmp_t:chr_file { create setattr unlink };
allow insights_core_t insights_core_tmp_t:lnk_file { create setattr unlink };
allow insights_core_t self:capability mknod;
```

For the original AVC denials see RHINENG-16433

### DIFF
--- a/policy/modules/contrib/insights_core.te
+++ b/policy/modules/contrib/insights_core.te
@@ -25,7 +25,7 @@ files_tmp_file(insights_core_tmp_t)
 
 #permissive insights_core_t;
 
-allow insights_core_t self:capability { dac_read_search setgid sys_admin sys_rawio};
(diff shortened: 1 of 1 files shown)


zpytela:
Looking generally good; would it be consodered an excessive policy loosening if the 3 rules were replaced by manage_ pattern instead? Since it is insights' private file, there should not be an impact to other domains.

Anyway, please move the rules next to others dealing with tmp files.

zpytela:
I've noticed it is a subset of already existing PR: https://github.com/fedora-selinux/selinux-policy/pull/2590
sorry for that

PR #2594, opened 2025-03-11
Allow systemd domains write global pressure stall information

zpytela:


martinpitt:
Urgh, sorry about the cockpit failure -- no idea why this happens just now. I sent https://github.com/cockpit-project/cockpit/pull/21701 , that should help.

PR #2598, opened 2025-03-14
Label wine's windows libraries as textrel_shlib_t

rfrohl:
Fixes these AVCs when running 'winecfg' without having 'selinuxuser_execmod' enabled: 

> type=AVC msg=audit(..): avc:  denied  { execmod } for pid=8451 comm="wineboot.exe" path="/usr/lib64/wine/x86_64-windows/ntdll.dll" dev="vda2" ino=197231 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:lib_t:s0 tclass=file permissive=0 
> type=AVC msg=audit(..): avc:  denied  { execmod } for pid=8445 comm="start.exe" path="/usr/lib/wine/i386-windows/ntdll.dll" dev="vda2" ino=196385 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:lib_t:s0 tclass=file permissive=0

### DIFF
--- a/policy/modules/system/libraries.fc
+++ b/policy/modules/system/libraries.fc
@@ -166,6 +166,7 @@ ifdef(`distro_redhat',`
 
 /usr/.*\.so(\.[^/]*)*		--	gen_context(system_u:object_r:lib_t,s0)
 /usr/lib/wine/.+\.so	--	gen_context(system_u:object_r:textrel_shlib_t,s0)
+/usr/lib/wine/.+-windows/.+\.(exe|dll|acm|drv|sys)	--	gen_context(system_u:object_r:textrel_shlib_t,s0)
 /usr/NX/lib/libXcomp\.so.*		--	gen_context(system_u:object_r:textrel_shlib_t,s0)
 /usr/NX/lib/libjpeg\.so.* 		--	gen_context(system_u:object_r:textrel_shlib_t,s0)
 


rfrohl:
I was wondering if a change like this would be acceptable for you.

This allows on oS Tumbleweed to run 'wine' without the need to enable 'selinuxuser_execmod'. It is not fully perfect (there are AVCs from the emulated windows binaries), but windows programs run functionally.

I was wondering if there would be objections to accepting something like this.

rfrohl:
And for reference - I tested on a fedora40 system (because I had it still around). Something similar to the below.
```
semanage boolean -m -0 selinuxuser_execmod
winecfg 
# -> fails
semanage fcontext -a -t textrel_shlib_t -f f '/usr/lib/wine/.+-windows/.+'
restorecon -Rv /usr/lib64/wine/x86_64-windows/ /usr/lib/wine/i386-windows/
winecfg
# -> works
```

So it also does not seem to need any other of the diverged changes we might have in the selinux-policy in openSUSE.

zpytela:
LGTM, can you just put 1 sentence in the commit message why this is actually an improvement?
Is the regexp `.+-windows/.+` needed to be so broad?

rfrohl:
> can you just put 1 sentence in the commit message why this is actually an improvement? 

I tried to improve the commit message a bit, with the new commit. Hope that provides some more context.

> Is the regexp `.+-windows/.+` needed to be so broad?

I re-tested and that more limited version also seems to work for my test cases. But not sure if I might be missing some functionality in wine, that I do not hit. I would follow up in that case if we see reports with a later commit.

zpytela:
Merging, thank you.

PR #2602, opened 2025-03-19
Improve non policy files and rename the default policy

bachradsusi:
The goal is to have clear building instructions which actually work and don't call the default policy `refpolicy` as this policy does not follow refpolicy for a long time.


INSTALL
```diff
@@ -1,19 +1,28 @@
...
+fedora-selinux policy has the following build requirements:
+	* libsepol 3.6
+	* libsemanage 3.6
+	* checkpolicy 3.6
+	* policycoreutils 3.6
 	* Python PyXML
 	* Python >= 3.4
 
-To install Reference Policy sources into /etc/selinux/refpolicy/src/policy:
+The policy requires selinux-container policy files to be installed policy/modules/contrib
```
vmojzis: container-selinux source files need to be placed in policy/modules/contrib

doc/templates/header.html
```diff
@@ -1,12 +1,12 @@
 <html>
 <head>
 <title>
- Security Enhanced Linux Reference Policy
+ Security Enhanced Linux fedora-selinux Policy
```
vmojzis: "Security Enhanced Linux Policy for Fedora"
or maybe 
"Fedora based Security Enhanced Linux Policy"

zpytela:
Generally LGTM

Just 2 questions:
- PKGNAME is further used only for docsdir, shouldn't it rather be set to "selinux-policy" to follow standards?
- why are sources put to /etc?

bachradsusi:
> Generally LGTM
> 
> Just 2 questions:
> 
>     * PKGNAME is further used only for docsdir, shouldn't it rather be set to "selinux-policy" to follow standards?

The idea is to call this project `fedora-selinux` and also reference it like this. 

Do you need a tarball? It 

>     * why are sources put to /etc?

This is what `make install-src` does, probably some historic relict. Should I remove part with `install-src` completely?

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L151
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L545

bachradsusi:
> PKGNAME is further used only for docsdir, shouldn't it rather be set to "selinux-policy" to follow standards?

The idea is to call this project fedora-selinux and also reference it like this, from /usr/share/doc/refpolicy to /usr/share/doc/fedora-selinux , NAME is changed in the same way

The policy installed directly from this sources would appear in /etc/selinux/fedora-selinux and /var/lib/selinux/fedora-selinux and users would use it by setting `SELINUXTYPE=fedora-selinux`

Fedora uses overrides `PKGNAME=selinux-policy` and `NAME=targeted`

> why are sources put to /etc?

This is what make install-src does, probably some historic relict. Should I remove part with install-src completely?

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L151
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L545

zpytela:
> > PKGNAME is further used only for docsdir, shouldn't it rather be set to "selinux-policy" to follow standards?
> 
> The idea is to call this project fedora-selinux and also reference it like this, from /usr/share/doc/refpolicy to /usr/share/doc/fedora-selinux , NAME is changed in the same way
> 
> The policy installed directly from this sources would appear in /etc/selinux/fedora-selinux and /var/lib/selinux/fedora-selinux and users would use it by setting `SELINUXTYPE=fedora-selinux`
> 
> Fedora uses overrides `PKGNAME=selinux-policy` and `NAME=targeted`

OK

> 
> > why are sources put to /etc?
> 
> This is what make install-src does, probably some historic relict. Should I remove part with install-src completely?
> 
> https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L151 https://github.com/fedora-selinux/selinux-policy/blob/rawhide/Makefile#L545

FHS might be dead, but the suggestion to put source files to /usr/src or /usr/local/src seems to be reasonable. I understand it may be outside of scope of this pr.

bachradsusi:
> I understand it may be outside of scope of this pr.

I agree that it's outside of the scope of this pr. Thanks.

zpytela:
I'll merge this pr then, thank you.

PR #2615, opened 2025-04-01
Allow sysadm_t to write to /dev/kmsg

WOnder93:
This is needed to be able to do `echo '<message>' >/dev/kmsg`, which is, among other things, used in TMT's test wrapper scripts [1]. Allowing this makes it possible to run TMT tests as sysadm_t without AVC denials.

[1] https://github.com/teemtee/tmt/blob/f3a0e53c633e727ded45cc7fbfd60f63b79c9487/tmt/steps/execute/internal.py#L166


zpytela:
Merging, thank you.

PR #2618, opened 2025-04-02
Hyperv: allow fcopy_uio_daemon to run as unconfined_service_t

rfrohl:
The fcopy_uio_daemon is not functional before this change and does not even start.

While trying to find a resolution, I also found some seemingly unneeded pieces. I included the cleanup as well, but am a bit unsure if I might have missed something relevant.

Let me know if the cleanup should be dropped.


rfrohl:
ok, I missed the obvious. Please ignore.

PR #2619, opened 2025-04-02
Remove unused hypervkvp_unit_file_t

rfrohl:
/usr/lib/systemd/system/hypervkvpd.service is labeled systemd_unit_file_t. Remove the hypervkvp_unit_file_t type, as it is not used.

I worked on a hyperv issue that is specific to the openSUSE policy and after the previous PR I saw that there are major differences in the packaging. Never the less the type is not used and I thought I will offer the smaller PR anyway. Maybe you will want to take it.

### DIFF
--- a/policy/modules/contrib/hypervkvp.if
+++ b/policy/modules/contrib/hypervkvp.if
@@ -81,7 +81,7 @@ interface(`hypervkvp_manage_lib_files',`
 
 #######################################
 ## <summary>
-##  Execute hypervkvp server in the hypervkvp domain.
+##  Execute hypervkvp server in the hypervkvp domain. (Deprecated)
 ## </summary>
 ## <param name="domain">
 ##  <summary>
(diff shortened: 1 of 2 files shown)


rfrohl:
The type is still in use in policy/modules/contrib/hypervkvp.if, but it does not seem needed. Let me know if I should extend the cleanup or if I am missing something obvious.

Could also add the type to /usr/lib/systemd/system/hypervkvpd.service in fcontext instead.

Let me know what you would prefer.

zpytela:
Referencing non-existing type does not make much sense, so it should be removed also from the if file and subsequently the hypervkvp_systemctl interface deprecated. While there, the hypervvssd_unit_file_t is not referenced either, in this case there is a risk though someone has a custom policy using it. Surely someone has assigned hypervkvp_unit_file_t, possibly, too - removing types has downsides.

rfrohl:
Then I will work on extending the clean up. Thanks for the feedback.

rfrohl:
I think that would be it, but not sure if I missed something. Did not deprecate anything before. Also would not mind some scrutiny on my changes to hypervkvp_admin(), as I do not know how to test this really.

rfrohl:
@zpytela Would you want more changes or does that look okay ?

zpytela:
It's OK now, but I'd like to merge it after rawhide policy is split from F41 if you don't mind.

rfrohl:
sure, let me know if you want me to rebase to make the merge easier.

zpytela:
I'm fine with this state, I just want to merge it after split of f43 policy.

zpytela:
Merging now, sorry for the delay.

PR #2681, opened 2025-05-07
contrib: add module for TDX "qgs" daemon

berrange:
The `qgs` daemon provides an API for creating signed quotes of TDX confidential virtual machine, and is used by QEMU. `qgs` uses the SGX software / technology for its implementation, loading and creating several SGX enclaves to create a signed attestation report of a VM. It also talks to public intel.com servers to fetch and then cache x509 certificates.

This is my "best guess" at a policy for the `qgs` daemon.

```
# ps -axuwfZ | grep bin/qgs
system_u:system_r:qgs_t:s0      qgs         2249  0.3  0.0 201376 15076 ?        Ssl  09:31   0:00 /usr/bin/qgs --no-daemon

# ls -alZ /etc/qgs.conf /etc/sgx_default_qcnl.conf 
-rw-r--r--. 1 root root system_u:object_r:qgs_etc_t:s0        49 Feb 17 19:00 /etc/qgs.conf
-rw-r--r--. 1 root root system_u:object_r:sgx_qcnl_etc_t:s0 4911 Feb 17 19:00 /etc/sgx_default_qcnl.conf

# ls -alZ /var/lib/qgs/
total 0
drwx------. 1 qgs  qgs  system_u:object_r:qgs_var_lib_t:s0  20 May  7 09:31 .
drwxr-xr-x. 1 root root system_u:object_r:var_lib_t:s0     956 May  2 04:28 ..
drwxr-xr-x. 1 qgs  qgs  system_u:object_r:qgs_var_lib_t:s0 128 May  7 09:31 .dcap-qcnl

# ls -alZ /var/lib/qgs/.dcap-qcnl/8bcc0451ba880002e901ccf4e5c90341e010cead7744c29fd82c005befd1b52a 
-rw-r--r--. 1 qgs qgs system_u:object_r:qgs_var_lib_t:s0 4121 May  7 09:31 /var/lib/qgs/.dcap-qcnl/8bcc0451ba880002e901ccf4e5c90341e010cead7744c29fd82c005befd1b52a

# ls -alZ /run/tdx-qgs/
total 0
drwxr-xr-x.  2 qgs  qgs  system_u:object_r:var_run_t:s0       60 May  7 09:31 .
drwxr-xr-x. 54 root root system_u:object_r:var_run_t:s0     1460 May  7 09:30 ..
srwxrwxrwx.  1 qgs  qgs  system_u:object_r:qgs_var_run_t:s0    0 May  7 09:31 qgs.socket
```

The last example above is the bit that looks not quite right to me.

I was trying to get the `/run/tdx-qgs` directory to have the `qgs_var_run_t` type label, but I can't figure out how to make that happen. The `/run` dir is tmpfs and the subdir gets created on the fly, but I'm unclear what's missing to get it to automatically gain the `qgs_var_run_t` label when auto-created by `qgs_t`.

Also the type and access rule for the `/etc/sgx_default_qcnl.conf` file is bundled with the 'qgs' policy module. Technically this is not specific to the `qgs` default and is generic for any use of `sgx`. I didn't feel it was worth having a separate 'sgx' policy module just for that, so I bundled it with my 'qgs' policy.

Maybe though, I should have just called this whole thing the `sgx` module instead of `qgs` ?  If doing that though, it would suggest using an `sgx_`  prefix on all types, eg `sgx_qgs_t` instead of just `qgs_t`.

If there is any preference from SELinux maintainers I'll adjust naming.

I'm testing this on Fedora 42 but my target is inclusion in Fedora 43 to support https://fedoraproject.org/w/index.php?title=Changes/ConfidentialVirtHostIntelTDX

### DIFF
--- a/policy/modules/contrib/qgs.fc
+++ b/policy/modules/contrib/qgs.fc
@@ -1,0 +1,6 @@
+/etc/qgs\.conf     -- gen_context(system_u:object_r:qgs_etc_t,s0)
+
+/usr/bin/qgs       -- gen_context(system_u:object_r:qgs_exec_t,s0)
+
+/var/lib/qgs(/.*)?    gen_context(system_u:object_r:qgs_var_lib_t,s0)
+/run/tdx-qgs(/.*)?    gen_context(system_u:object_r:qgs_var_run_t,s0)
--- a/policy/modules/contrib/qgs.te
+++ b/policy/modules/contrib/qgs.te
@@ -1,0 +1,71 @@
+policy_module(qgs, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+type qgs_t;
+type qgs_exec_t;
+init_daemon_domain(qgs_t, qgs_exec_t)
+
+permissive qgs_t;
+
+type qgs_var_lib_t;
+files_type(qgs_var_lib_t);
+
+type qgs_var_run_t;
+files_pid_file(qgs_var_run_t);
+
+# Config file exclusively for SGX
+type qgs_etc_t;
+
+########################################
+#
+# qgs local policy
+#
+allow qgs_t self:fifo_file rw_fifo_file_perms;
+allow qgs_t self:unix_stream_socket create_stream_socket_perms;
+
+# /var/lib/qgs is the $HOME for 'qgs' and it caches some
+# data under subdirs
+manage_dirs_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_lnk_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+
+# /run/tdx/qgs is where 'qgs' creates UNIX socket
+manage_dirs_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_sock_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+files_pid_filetrans(qgs_t, qgs_var_run_t, { dir })
+
+domain_use_interactive_fds(qgs_t)
+
+# To read /etc/qgs.conf for its configuration
+files_config_file(qgs_etc_t)
+read_files_pattern(qgs_t, qgs_etc_t, qgs_etc_t)
+
+corenet_tcp_connect_http_port(qgs_t)
+
+# It loads enclaves with ...
+dev_rw_sgx_provision(qgs_t)
+
+# ...and executes enclaves to create quotes
+dev_rwx_sgx_enclave(qgs_t)
+
+optional_policy(`
+  logging_send_syslog_msg(qgs_t)
+')
+
+optional_policy(`
+  miscfiles_read_localization(qgs_t)
+')
+
+# It connects to intel.com to acquire certificates
+optional_policy(`
+  miscfiles_read_generic_certs(qgs_t)
+')
+
+optional_policy(`
+  sysnet_dns_name_resolve(qgs_t)
+')
--- a/policy/modules/contrib/virt.te
+++ b/policy/modules/contrib/virt.te
@@ -1355,6 +1355,10 @@ optional_policy(`
 	xserver_rw_shm(virt_domain)
 ')
 
+optional_policy(`
+        qgs_stream_connect(svirt_t)
+')
+
 ########################################
 #
 # xm local policy
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -6863,6 +6863,43 @@ interface(`dev_rw_sgx_vepc',`
 
 ########################################
 ## <summary>
+##      Allow read and write the sgx_enclave devices
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`dev_rwx_sgx_enclave',`
+        gen_require(`
+                type device_t, sgx_enclave_device_t;
+        ')
+
+        rw_chr_files_pattern($1, device_t, sgx_enclave_device_t)
+        allow $1 sgx_enclave_device_t:chr_file { map execute };
+')
+
+########################################
+## <summary>
+##      Allow read and write the sgx_provision devices
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`dev_rw_sgx_provision',`
+        gen_require(`
+                type device_t, sgx_provision_device_t;
+        ')
+
+        rw_chr_files_pattern($1, device_t, sgx_provision_device_t)
+')
+
+########################################
+## <summary>
 ##	Allow read the hfi1_[0-9]+ devices
 ## </summary>
 ## <param name="domain">
--- a/dist/mls/modules.conf
+++ b/dist/mls/modules.conf
@@ -1384,6 +1384,13 @@ ptchown = module
 # 
 pulseaudio = module
 
+# Layer: service
+# Module: qgs
+#
+# TDX QGS Daemon
+#
+qgs = module
+
 # Layer: services
 # Module: qmail
 #
--- a/dist/targeted/modules.conf
+++ b/dist/targeted/modules.conf
@@ -1803,6 +1803,13 @@ puppet = module
 # 
 pwauth = module
 
+# Layer: service
+# Module: qgs
+#
+# TDX QGS Daemon
+#
+qgs = module
+
 # Layer: services
 # Module: qmail
 #
--- a/policy/modules/contrib/qgs.if
+++ b/policy/modules/contrib/qgs.if
@@ -1,0 +1,97 @@
+## <summary>policy for qgs</summary>
+
+########################################
+## <summary>
+##      Execute qgs_exec_t in the qgs domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##      Domain allowed to transition.
+## </summary>
+## </param>
+#
+interface(`qgs_domtrans',`
+	gen_require(`
+                type qgs_t, qgs_exec_t;
+	')
+
+	corecmd_search_bin($1)
+        domtrans_pattern($1, qgs_exec_t, qgs_t)
+')
+
+######################################
+## <summary>
+##      Execute qgs in the caller domain.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`qgs_exec',`
+	gen_require(`
+		type qgs_exec_t;
+	')
+
+        corecmd_search_bin($1)
+	can_exec($1, qgs_exec_t)
+')
+
+########################################
+## <summary>
+##      All of the rules required to administrate
+##      an qgs environment
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+## <param name="role">
+##      <summary>
+##      Role allowed access.
+##      </summary>
+## </param>
+## <rolecap/>
+#
+interface(`qgs_admin',`
+        gen_require(`
+                type qgs_t;
+                type qgs_var_lib_t;
+                type qgs_var_run_t;
+        ')
+
+        allow $1 qgs_t:process { signal_perms };
+        ps_process_pattern($1, qgs_t)
+
+    tunable_policy(`deny_ptrace',`',`
+        allow $1 qgs_t:process ptrace;
+    ')
+
+        files_search_var_lib($1)
+        admin_pattern($1, qgs_var_lib_t)
+        admin_pattern($1, qgs_var_run_t)
+        optional_policy(`
+                systemd_passwd_agent_exec($1)
+                systemd_read_fifo_file_passwd_run($1)
+        ')
+')
+
+## <summary>
+##      Connect to qgs over an unix
+##      domain stream socket.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`qgs_stream_connect',`
+	gen_require(`
+                type qgs_t, qgs_var_run_t;
+        ')
+
+	stream_connect_pattern($1, qgs_var_run_t, qgs_var_run_t, qgs_t)
+')


berrange:
Any thoughts on this proposed policy module addition ?

zpytela:
I am sorry for the delay, I have only a few general comments:
- it is good if all types are consistent with the module name and have the same prefix (in which module will I find the sgx_qcnl_etc_ttype?)
- currently F41, F42, F43 share the same policy, but it will change soon, F43 as a target is fine
- files in /run can be created by systemd if referred to from the service unit file, or handled using files_pid_filetrans() 
I will now reply with some particular comments inline.

policy/modules/contrib/qgs.fc
```diff
@@ -0,0 +1,8 @@
+/etc/qgs.conf --             gen_context(system_u:object_r:qgs_etc_t,s0)
+
+/etc/sgx_default_qcnl.conf  gen_context(system_u:object_r:sgx_qcnl_etc_t,s0)
```
zpytela: Use "--" for a plain file.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,74 @@
...
+type qgs_exec_t;
+init_daemon_domain(qgs_t, qgs_exec_t)
+
+permissive qgs_t;
+
+type qgs_var_lib_t;
+files_type(qgs_var_lib_t)
+
+type qgs_var_run_t;
+files_type(qgs_var_run_t)
```
zpytela: not necessary, a subset of the next one

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,74 @@
...
+files_pid_file(qgs_var_run_t)
+files_create_var_run_dirs(qgs_t)
```
zpytela: This does not feel right, I expect only files_pid_filetrans() is needed.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,74 @@
...
+
+# Common config file for potentially many SGX
+# related apps doing SGX quoting
+type sgx_qcnl_etc_t;
+files_type(sgx_qcnl_etc_t)
```
zpytela: Do you expect these config file types will not (or should not) be read by any other domain? It would be convenient to mark them using files_config_file(), but that gives access to the type to some 20 other domains.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,74 @@
...
+manage_dirs_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_lnk_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+files_var_lib_filetrans(qgs_t, qgs_var_lib_t, { dir file lnk_file })
+
+# /run/tdx/qgs is where 'qgs' creates UNIX socket
+manage_dirs_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_sock_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+files_pid_filetrans(qgs_t, qgs_var_run_t, { dir file sock_file })
```
zpytela: This is exactly the interface which should suffice. If there are still mislabeled files in /run, may they be created by some other service?, or previously created by an unconfined user?

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,74 @@
...
+# It dynmically loads sgx qcnl which reads config
+# from /etc/sgx_default_qcnl.conf
+files_config_file(sgx_qcnl_etc_t)
+read_files_pattern(qgs_t, sgx_qcnl_etc_t, sgx_qcnl_etc_t)
+
+# To read /etc/qgs.conf for its configuration
+files_config_file(qgs_etc_t)
+read_files_pattern(qgs_t, qgs_etc_t, qgs_etc_t)
+
+logging_send_syslog_msg(qgs_t)
```
zpytela: calls to logging, miscfiles, sysnetwork interfaces should be in an optional block

policy/modules/contrib/virt.te
```diff
@@ -1355,6 +1355,10 @@ optional_policy(`
 	xserver_rw_shm(virt_domain)
 ')
 
+optional_policy(`
```
zpytela, line 1358: if it uses types from the same module, it does not need to be in an optional block. Actually using directly stream_connect_pattern() would be better.

policy/modules/kernel/devices.if
```diff
@@ -6861,6 +6861,43 @@ interface(`dev_rw_sgx_vepc',`
...
+##      </summary>
+## </param>
+#
+interface(`dev_rw_sgx_enclave',`
+        gen_require(`
+                type device_t, sgx_enclave_device_t;
+        ')
+
+        rw_chr_files_pattern($1, device_t, sgx_enclave_device_t)
+        allow $1 sgx_enclave_device_t:chr_file { map execute };
```
zpytela, line 6880: Please create a new interface for execution or use the _rwx_ interface name, depending on how it will possibly be used by other domains.

zpytela (changes requested):
Apart from the minor flaws it is looking good.

policy/modules/contrib/qgs.te
berrange: I don't believe any of the data in the config file is security sensitive, so it is probably fine to use `files_config_file`.

policy/modules/contrib/qgs.te
berrange: I see almost no other existing modules wrap these in optional ?

policy/modules/contrib/virt.te
berrange: This is the 'virt' module using types from the 'qgs' module, so surely 'optional_policy' is needed here ?

policy/modules/kernel/devices.if
berrange: Yes, this should have been 'rwx' .

policy/modules/contrib/qgs.te
zpytela: It's getting better, but you are right that 5 years ago it was almost never. Anyway refer to e.g. https://github.com/fedora-selinux/selinux-policy/blob/rawhide/dist/targeted/modules.conf#L323

policy/modules/contrib/virt.te
zpytela: You are right, I overlooked this actually is in virt.te.

policy/modules/contrib/qgs.fc
```diff
@@ -0,0 +1,8 @@
+/etc/qgs.conf --             gen_context(system_u:object_r:qgs_etc_t,s0)
```
zpytela: One more note: fc entries are regexps, so dots should be escaped.

policy/modules/contrib/qgs.fc
berrange: Are there any docs that explain what the syntax of the '.fc' file is  ? I'm particularly curious what the different options mean in between the filename & the `gen_context` call, as I struggle to understand what the pattern/guidance is for when to omit this, or when to use '--', or when to use something else.

zpytela: Can't find a good one now, but there are
https://github.com/OwlCyberDefense/refpolicy/wiki/GettingStarted#module-fc-policy-modulefcpolicy
and
https://access.redhat.com/articles/6999267#file-context-13

For the blank file type: applies to all types if not specified.

policy/modules/contrib/qgs.te
berrange: Yes, this was able to be removed

policy/modules/contrib/qgs.te
berrange: Yes, removed now.

policy/modules/contrib/qgs.te
berrange: I've removed the 'sgx_qcl_etc_t' type entirely and just leave this file labelled 'etc_t'.

policy/modules/contrib/qgs.te
berrange: I am wrapping these in optional now.

policy/modules/contrib/qgs.fc
berrange: I have fixed the regex escaping.

berrange:
I have pushed an update of the policy, aside from the changes you requested, the other notable changes are:

* Introducing some new interfaces for systemd, related to its WorkingDirectory and RuntimeDirectory  directives and the permissions they require
* Adding rules to the 'unconfineduser' to avoid AVCs when unconfined ssh sessions look at /var/lib/qgs and /var/run/tdx-qgs
* Split the changes out into separate logically independent commits

With this I'm now able todo the expected operations without seeing any AVCs

berrange:
Any thoughts on this 2nd attempt at the qgs policy ?

zpytela:
It's generally okay, I am just blocked (by myself) since I didn't quite get the systemd attributes purpose, sorry for that. I'll make my best to finish the review soon.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+#
+allow qgs_t self:fifo_file rw_fifo_file_perms;
+allow qgs_t self:unix_stream_socket create_stream_socket_perms;
+
+# /var/lib/qgs is the $HOME for 'qgs' and it caches some
+# data under subdirs
+manage_dirs_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+manage_lnk_files_pattern(qgs_t, qgs_var_lib_t, qgs_var_lib_t)
+files_var_lib_filetrans(qgs_t, qgs_var_lib_t, { dir file lnk_file })
```
zpytela: I see /var/lib/qgx is a part of rpm, so this should not be needed at all.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+
+# /run/tdx/qgs is where 'qgs' creates UNIX socket
+manage_dirs_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+manage_sock_files_pattern(qgs_t, qgs_var_run_t, qgs_var_run_t)
+files_pid_filetrans(qgs_t, qgs_var_run_t, { dir file sock_file })
```
zpytela: Based on fc entries, only dir should be required for files_pid_filetrans.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+
+domain_use_interactive_fds(qgs_t)
+
+files_read_etc_files(qgs_t)
```
zpytela: I believe this is not needed. All domains are allowed to read base_ro_file_type which etc_t is a part of.

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+
+# "RuntimeDirectory" directive causes systemd to
+# create a directory and mount a tmpfs on it for
+# service runtime state, and purge it on shutdown
+optional_policy(`
+  systemd_runtime_dir(qgs_var_run_t)
+  systemd_working_dir(qgs_var_lib_t)
+')
+
+corenet_tcp_connect_http_port(qgs_t)
```
zpytela: corenet_ and dev_ interfaces calls should be before optional blocks (L49 here), sorted alphabetically.

policy/modules/system/systemd.if
```diff
@@ -2752,6 +2752,90 @@ interface(`systemd_mount_dir',`
...
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`systemd_runtime_dir',`
+    gen_require(`
+        attribute systemd_runtime_directory;
+    ')
+
+    files_type($1)
```
zpytela: I think this is not just not needed as $1 expectedly has been declared a file type before, but it actually adds the type to other attributes which is probably not wanted.

berrange:
In case it helps, *without* the following calls to the new systemd interfaces:

```
  systemd_runtime_dir(qgs_var_run_t)
  systemd_working_dir(qgs_var_lib_t)
```

then the system will show the following AVCs upon `systemctl start qgs.service` 

```
type=AVC msg=audit(1750077380.192:345): avc:  denied  { create } for  pid=121433 comm="(qgs)" name="tdx-qgs" scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.192:346): avc:  denied  { associate } for  pid=121433 comm="(qgs)" name="tdx-qgs" scontext=system_u:object_r:qgs_var_run_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=1
type=AVC msg=audit(1750077380.192:347): avc:  denied  { getattr } for  pid=121433 comm="(qgs)" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.193:348): avc:  denied  { read } for  pid=121433 comm="(qgs)" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.193:349): avc:  denied  { open } for  pid=121433 comm="(qgs)" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.193:350): avc:  denied  { setattr } for  pid=121433 comm="(qgs)" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.198:351): avc:  denied  { mounton } for  pid=121433 comm="(qgs)" path="/run/systemd/mount-rootfs/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077380.201:352): avc:  denied  { search } for  pid=121433 comm="(qgs)" name="qgs" dev="sda3" ino=430043 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_lib_t:s0 tclass=dir permissive=1
```

Most of those are related to systemd dynamically-creating the `/run/tdx-qgs` directory declared via the `RuntimeDirectory` directive in the unit file, while the last one is due to systemd trying to `chdir` into the `/var/lib/qgs` directory declared via the `WorkingDirectory` directive in the unit file

Upon `systemctl stop qgs.service` the following additional AVCs are triggered

```
type=AVC msg=audit(1750077637.789:355): avc:  denied  { search } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:356): avc:  denied  { getattr } for  pid=1 comm="systemd" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:357): avc:  denied  { read } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:358): avc:  denied  { open } for  pid=1 comm="systemd" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:359): avc:  denied  { write } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:360): avc:  denied  { remove_name } for  pid=1 comm="systemd" name="qgs.socket" dev="tmpfs" ino=3577 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1750077637.789:361): avc:  denied  { unlink } for  pid=1 comm="systemd" name="qgs.socket" dev="tmpfs" ino=3577 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=sock_file permissive=1
type=AVC msg=audit(1750077637.789:362): avc:  denied  { rmdir } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
```

which are caused by systemd recursively purging all content from the `/run/tdx-qgs` directory after stopping the service.

I was a little surprised that no existing systemd services using `RuntimeDirectory` or `WorkingDirectory` have needed this adding to the policy already, but the few I looked at have some subtle differences that perhaps avoided it.

policy/modules/system/systemd.if
```diff
@@ -2752,6 +2752,90 @@ interface(`systemd_mount_dir',`
...
+## <rolecap/>
+#
+interface(`systemd_allow_manage_runtime_dir',`
+    gen_require(`
+        attribute systemd_runtime_directory;
+    ')
+
+    allow $1 systemd_runtime_directory:dir manage_dir_perms;
+    allow $1 systemd_runtime_directory:file manage_file_perms;
+    allow $1 systemd_runtime_directory:sock_file manage_sock_file_perms;
```
zpytela: mounton permission needs to be here, it should pop up when init_create_dirs boolean is off

setattr may need to be added later

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+
+type qgs_t;
+type qgs_exec_t;
+init_daemon_domain(qgs_t, qgs_exec_t)
+
+permissive qgs_t;
+
+type qgs_var_lib_t;
+
+type qgs_var_run_t;
```
zpytela, line 17: files_pid_file(qgs_var_run_t;) should be here

policy/modules/contrib/qgs.te
```diff
@@ -0,0 +1,80 @@
...
+# Declarations
+#
+
+type qgs_t;
+type qgs_exec_t;
+init_daemon_domain(qgs_t, qgs_exec_t)
+
+permissive qgs_t;
+
+type qgs_var_lib_t;
```
zpytela, line 14: files_type(qgs_var_lib_t;) should be here

zpytela (changes requested):

zpytela:
Why it took me so long is that the systemd changes overlap with existing concept of init_create_dirs. You though followed another existing concept in systemd.if which is fine. Again sorry for the delay. Only small changes are needed now.

zpytela:
> In case it helps, _without_ the following calls to the new systemd interfaces:
> 
> ```
>   systemd_runtime_dir(qgs_var_run_t)
>   systemd_working_dir(qgs_var_lib_t)
> ```
> 
> then the system will show the following AVCs upon `systemctl start qgs.service`
> 
> ```
> type=AVC msg=audit(1750077380.192:345): avc:  denied  { create } for  pid=121433 comm="(qgs)" name="tdx-qgs" scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.192:346): avc:  denied  { associate } for  pid=121433 comm="(qgs)" name="tdx-qgs" scontext=system_u:object_r:qgs_var_run_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=1
> type=AVC msg=audit(1750077380.192:347): avc:  denied  { getattr } for  pid=121433 comm="(qgs)" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.193:348): avc:  denied  { read } for  pid=121433 comm="(qgs)" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.193:349): avc:  denied  { open } for  pid=121433 comm="(qgs)" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.193:350): avc:  denied  { setattr } for  pid=121433 comm="(qgs)" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.198:351): avc:  denied  { mounton } for  pid=121433 comm="(qgs)" path="/run/systemd/mount-rootfs/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077380.201:352): avc:  denied  { search } for  pid=121433 comm="(qgs)" name="qgs" dev="sda3" ino=430043 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_lib_t:s0 tclass=dir permissive=1
> ```
> 
> Most of those are related to systemd dynamically-creating the `/run/tdx-qgs` directory declared via the `RuntimeDirectory` directive in the unit file, while the last one is due to systemd trying to `chdir` into the `/var/lib/qgs` directory declared via the `WorkingDirectory` directive in the unit file
> 
> Upon `systemctl stop qgs.service` the following additional AVCs are triggered
> 
> ```
> type=AVC msg=audit(1750077637.789:355): avc:  denied  { search } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:356): avc:  denied  { getattr } for  pid=1 comm="systemd" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:357): avc:  denied  { read } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:358): avc:  denied  { open } for  pid=1 comm="systemd" path="/run/tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:359): avc:  denied  { write } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:360): avc:  denied  { remove_name } for  pid=1 comm="systemd" name="qgs.socket" dev="tmpfs" ino=3577 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> type=AVC msg=audit(1750077637.789:361): avc:  denied  { unlink } for  pid=1 comm="systemd" name="qgs.socket" dev="tmpfs" ino=3577 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=sock_file permissive=1
> type=AVC msg=audit(1750077637.789:362): avc:  denied  { rmdir } for  pid=1 comm="systemd" name="tdx-qgs" dev="tmpfs" ino=3575 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:qgs_var_run_t:s0 tclass=dir permissive=1
> ```
> 
> which are caused by systemd recursively purging all content from the `/run/tdx-qgs` directory after stopping the service.
> 
> I was a little surprised that no existing systemd services using `RuntimeDirectory` or `WorkingDirectory` have needed this adding to the policy already, but the few I looked at have some subtle differences that perhaps avoided it.

I believe with using proper files_ interfaces at least some of them will not be needed, just will require init_create_dirs boolean on.

berrange:
> I believe with using proper files_ interfaces at least some of them will not be needed, just will require init_create_dirs boolean on.

Putting the rules behind a boolean feels wrong to me, as if the boolean is disabled that is giving a guaranteed broken virt host.

zpytela:
> > I believe with using proper files_ interfaces at least some of them will not be needed, just will require init_create_dirs boolean on.
> 
> Putting the rules behind a boolean feels wrong to me, as if the boolean is disabled that is giving a guaranteed broken virt host.

If the boolean is disabled, many services will fail, but that's the admin's responsibility and the usual tradeoff security-usability. Your solution is different though, so I am not challenging it. What I had in mind is that some permissions (mounton) are built on top of that boolean though, so you'd rather put them in your interfaces explicitly.

berrange:
> > In case it helps, _without_ the following calls to the new systemd interfaces:
> > ```
> >   systemd_runtime_dir(qgs_var_run_t)
> >   systemd_working_dir(qgs_var_lib_t)
> > ```
... snip...

> > then the system will show the following AVCs upon `systemctl start qgs.service`

... snip...

> > I was a little surprised that no existing systemd services using `RuntimeDirectory` or `WorkingDirectory` have needed this adding to the policy already, but the few I looked at have some subtle differences that perhaps avoided it.
> 
> I believe with using proper files_ interfaces at least some of them will not be needed, just will require init_create_dirs boolean on.

Sigh, it looks like you are correct about this. After adding the 'files_*' rules, it appears systemd "just works" as needed, so that was all an unfortunate diversion. I'll respin without the systemd changes.

berrange:
Updated with all suggested changes

zpytela:
LGTM now, I'll just wait for the build to check on a system and then I can merge it.

zpytela:
There is an issue with
```
ExecStartPre=+mkdir -p /var/run/tdx-qgs
```

berrange:
> There is an issue with
> 
> ```
> ExecStartPre=+mkdir -p /var/run/tdx-qgs
> ```

That is gone in rawhide, but not in F42.  I use "RuntimeDirectory" instead, so that systemd just "does the right thing(tm)".

I was only expecting this selinux policy to be pushed to rawhide, and backporting to RHEL-9/10, but if you want, I can change the service file in F42 too, to make it selinux friendly.

zpytela:
We currently use the same policy content for F41/42/43, hope that's not a problem. You don't need to backport the code changes, I just happened to use F42 this time.

berrange:
> We currently use the same policy content for F41/42/43, hope that's not a problem. 

That's absolutely fnie.

> You don't need to backport the code changes, I just happened to use F42 this time.

I should probably just do that regardless of everyone's sanity.

zpytela:
Merging now, thanks for the effort and great cooperation.

PR #2686, opened 2025-05-12
Label /dev/diag as power_device_t

WOnder93:
According to [1] it is used to get "power readings", so presumably power_device_t fits well for it. It is an s390(x)-specific device.

[1] https://github.com/ibm-s390-linux/s390-tools/blob/6dd671f24e56275380ab25eb73b217fbb49fd07e/zpwr/zpwr.c

### DIFF
--- a/policy/modules/kernel/devices.fc
+++ b/policy/modules/kernel/devices.fc
@@ -24,6 +24,7 @@
 /dev/crash		-c	gen_context(system_u:object_r:crash_device_t,mls_systemhigh)
 /dev/crypto/nx-gzip	-c	gen_context(system_u:object_r:accelerator_device_t,mls_systemhigh)
 /dev/dahdi/.*		-c	gen_context(system_u:object_r:sound_device_t,s0)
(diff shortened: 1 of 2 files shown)


zpytela:
I am not quite convinced it is power-related (power as in current), perhaps memory topology querying instead [1]? The power keyword may refer just to the architecture.

Anyway I was unable to reliably find the purpose of the device; the same filename also seems to be reused as a diagnostics monitor for qualcomm modems [2], so we'd better look for it.

[1] https://github.com/ibm-s390-linux/s390-tools/blob/master/zmemtopo/zmemtopo.c#L207
[2] https://github.com/P1sec/QCSuper/blob/master/docs/The%20Diag%20protocol.md

WOnder93:
You're right, it can be used to get either power readings or memory topology depending on the ioctl number... I'll label it with a new `diagnostic_device_t` type then (which would also match the description at the top of `arch/s390/kernel/diag/diag_misc.c` in the kernel).

zpytela:
Well, one should rather use current kernel sources than those for current fedora. It is clearer now.

zpytela:
Merging, thank you.

Issue #2637, opened 2025-04-10
Failed to login as systemd-homed managed user on Fedora 42


ekatwired:
I am trying out Fedora 42 right now and decided to create a user using systemd-homed. Unless SELinux enforcing disabled, I cannot log in. Reproduced on booted live system in VM and on manually installed system.

When I first run `homectl activate <user>` and then log-in, user is logged in sucessfully.

Attachment: [audit.log](https://github.com/user-attachments/files/19685962/audit.log)

# Error
```
liveuser login: meow
Password:
Last login: Thu Apr 10 12:36:53 on tty2
Called without reference on home token, can't operate.
```

# User creation
User created using: `homectl create meow --storage=subvolume`

zpytela:
@erin-desu Can you try if the following module is sufficient?

```
# cat local_homed_login.cil
(allow local_login_t systemd_homed_t (dbus (send_msg)))
(allow systemd_homed_t local_login_t (dbus (send_msg)))
# semodule -i local_homed_login.cil
```

ekatwired:
Now it asks for password twice.

Errors with
```
-- meow: /home/meow: change directory failed: Permission denied
```
And then just how it goes on from that.

If i create a new user, it hangs until i manually activate it first, then after deactivation it proceeds with same error.

If i first disable enforce, it works fine. Seems like it fails to chown the homedir.

zpytela:
Are there any new AVC denials? Does journal contain any particular error reported?

ekatwired:
> Are there any new AVC denials? Does journal contain any particular error reported?

[audit.log](https://github.com/user-attachments/files/19693493/audit.log)

zpytela:
@erin-desu Can you temporarily switch the system to permissive mode and get all denials?

setenforce 0

ekatwired:
Now I am *somehow* getting the same issue as on the beginning even with SELinux off.

s/off/permissive

zpytela:
@erin-desu you can now try the coprbuild from
https://github.com/fedora-selinux/selinux-policy/pull/2650

PR #2690, opened 2025-05-13
Allow systemd-homed to start services.

rfrohl:
Follow up to recent homed PR [0], adding a solution for the mentioned, but missing, AVC in that PR. 

We had a parallel bug [1], but I did not progress to much until I found the PR. But I now could get feedback from the reporter which issues he still sees. Maybe this is the proper solution.

Please let me know if it needs changes.

[0] https://github.com/fedora-selinux/selinux-policy/issues/2637
[1] https://bugzilla.suse.com/show_bug.cgi?id=1240411#c21

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -720,6 +720,7 @@ optional_policy(`
 
 optional_policy(`
 	systemd_logind_stream_connect(login_pgm)
+	systemd_start_systemd_services(login_pgm)
 ')
 
 optional_policy(`


rfrohl:
waiting for feedback, if solution works

rfrohl:
ok, got confirmation. dropping draft state.

zpytela:
@rfrohl Is start the only permission needed? Checking the current state, there is also status for other unit types:
```
f43# sesearch -A -s xdm_t -t systemd_unit_file_type  -c service -p start
allow xdm_t dbusd_unit_file_t:service { disable enable reload start status stop };
allow xdm_t power_unit_file_t:service { start status };
allow xdm_t xdm_unit_file_t:service { start status };
```

policy/modules/system/authlogin.te
```diff
@@ -727,3 +727,7 @@ optional_policy(`
 	systemd_homed_write_pipes(login_pgm)
 	systemd_homed_dbus_chat(login_pgm)
 ')
+
+optional_policy(`
```
zpytela: There already is an optional block for the systemd module interfaces.

zpytela (changes requested):

rfrohl:
> @rfrohl Is start the only permission needed? Checking the current state, there is also status for other unit types:
> 

yes, as far as I know. Should I add 'status' as well, just to be on the safe side ?

zpytela:
If status was not requested during testing, it may not be needed.
Merging, thank you.

PR #2691, opened 2025-05-13
Allow named_t to read NetworkManager's runtime files

WOnder93:
It may need to read the no-stub-resolv.conf file (probably via the /etc/resolv.conf symlink).

Original AVC:
type=AVC msg=audit(1745926926.638:13790): avc:  denied  { read } for  pid=95889 comm="unbound-anchor" name="no-stub-resolv.conf" dev="tmpfs" ino=14695 scontext=system_u:system_r:named_t:s0 tcontext=system_u:object_r:NetworkManager_var_run_t:s0 tclass=file permissive=0

Resolves: rhbz#2362879

### DIFF
--- a/policy/modules/contrib/bind.te
+++ b/policy/modules/contrib/bind.te
@@ -238,6 +238,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	networkmanager_read_pid_files(named_t)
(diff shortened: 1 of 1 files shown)


zpytela:
Merging, thank you.

PR #2692, opened 2025-05-13
Allow svirt_t to read certificates in user home directories

WOnder93:
This appears to be needed for user session VMs with PKCS Smartcard device added.

It would probably better under an opt-in boolean, but the userdom_read_home_certs() interface assigns the type to an attribute, which can't be controlled by a boolean.

Original AVC:
type=AVC msg=audit(1743337047.390:2649): avc:  denied  { read } for  pid=17278 comm="qemu-system-x86" name="pkcs11.txt" dev="dm-0" ino=1557247 scontext=unconfined_u:unconfined_r:svirt_t:s0:c45,c54 tcontext=unconfined_u:object_r:home_cert_t:s0 tclass=file permissive=0

Resolves: rhbz#2356061

### DIFF
--- a/policy/modules/contrib/virt.te
+++ b/policy/modules/contrib/virt.te
@@ -1251,6 +1251,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	userdom_read_home_certs_tunable(virt_use_pcscd, virt_domain)
 	tunable_policy(`virt_use_pcscd',`
 		pcscd_stream_connect(virt_domain)
 	')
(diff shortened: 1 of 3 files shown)


WOnder93:
Opened this as draft because it doesn't seem right to allow this without the boolean... Maybe we should instead ditch the `userdom_home_reader_certs_type` attribute and move the rules directly to the interface?

WOnder93:
So, I tried to move the `userdom_read_home_certs()` under the `virt_use_pcscd` boolean, but it turned out a bit complicated due to the tunable_policy(`use_ecryptfs_home_dirs', ...) part... @zpytela Let me know if you prefer one (current state) or the other (5301ee2ef1f37bf3d34ebab9da73031221c1cff9).

WOnder93:
@zpytela Ping.

zpytela:
LGTM
I admit it took me a while to get through it, I am sorry for that

PR #2693, opened 2025-05-14
Label plasma user service files as xdm_unit_file_t.

rfrohl:
I am 90% sure that the PR [0] from yesterday is actually wrong and far to broad as a result. Because of the homed context I was to focused on the homed related problem solution, but I now think the reported AVC is just a simple plasma labeling problem. 

The sesearch command did keep nagging me that something was wrong with my PR:

> $ sesearch -A -s xdm_t -t systemd_unit_file_type  -c service -p start
> [..]
> allow xdm_t xdm_unit_file_t:service { start status };

Trying to confirm that the proposed change also fixes the issue. Sorry for the wrong PR from yesterday and only catching it now. Will drop the draft state once I got confirmation.

[0] https://github.com/fedora-selinux/selinux-policy/pull/2690


rfrohl:
got confirmation that this solutions also works.

zpytela:
OK, merging, rawhide build will follow soon.

PR #2701, opened 2025-05-22
mariadb bin can be located in /usr/sbin instead of /usr/bin

PopiBrossard:
Dear maintainer,

When installing mariadb using their official repository (https://mirror.mariadb.org/yum/) the binaries are located in `/usr/sbin` instead of `/usr/bin`. Due to this, Mariadb service is not confined:

```bash
system_u:system_r:unconfined_service_t:s0 mysql 221126 1 37 12:22 ?        00:00:03 /usr/sbin/mariadbd
```

If you look at `mysqld.fc`, we can see the context is only set for binaries in `/usr/bin`.

Because SELinux is already configured, but for `/usr/bin`, I think the related SELinux configuration should also confine mariadb binaries located in `/usr/sbin` directory.

I've used a simple regex for this, but I can (if preferred) just copy/paste the lines with `/usr/bin` to `/usr/sbin`.

Regards,

### DIFF
--- a/policy/modules/contrib/mysql.fc
+++ b/policy/modules/contrib/mysql.fc
@@ -21,30 +21,30 @@ HOME_DIR/\.my\.cnf -- gen_context(system_u:object_r:mysqld_home_t, s0)
 #
 # /usr
 #
-/usr/bin/mysqld_safe	--	gen_context(system_u:object_r:mysqld_safe_exec_t,s0)
-/usr/bin/mysqld_safe_helper    --      gen_context(system_u:object_r:mysqld_exec_t,s0)
-/usr/bin/mysql_upgrade	--	gen_context(system_u:object_r:mysqld_exec_t,s0)
+/usr/(s)?bin/mysqld_safe	--	gen_context(system_u:object_r:mysqld_safe_exec_t,s0)
+/usr/(s)?bin/mysqld_safe_helper    --      gen_context(system_u:object_r:mysqld_exec_t,s0)
+/usr/(s)?bin/mysql_upgrade	--	gen_context(system_u:object_r:mysqld_exec_t,s0)
 
 /usr/libexec/mysqld	--	gen_context(system_u:object_r:mysqld_exec_t,s0)
 /usr/libexec/mysqld_safe-scl-helper --  gen_context(system_u:object_r:mysqld_safe_exec_t,s0)
 
 
-/usr/bin/mysqld(-max|-debug)?	--	gen_context(system_u:object_r:mysqld_exec_t,s0)
-/usr/bin/mysqlmanager	--	gen_context(system_u:object_r:mysqlmanagerd_exec_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
These changes are not needed. As a part of [1] all /sbin and /usr/sbin paths were changed to /usr/bin. There is now an equivalency rule [2] to handle some possible misconfigurations. The current state in Fedora is
```
f43# ls -lZ /usr/bin/mariadbd /usr/libexec/mariadbd
lrwxrwxrwx. 1 root root system_u:object_r:bin_t:s0               21 May  7 03:00 /usr/bin/mariadbd -> /usr/libexec/mariadbd
-rwxr-xr-x. 1 root root system_u:object_r:mysqld_exec_t:s0 22351624 May  7 03:00 /usr/libexec/mariadbd

```
so I guess you use some other distribution? Does it use files in the /config directory?

[1] https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin
[2] https://github.com/fedora-selinux/selinux-policy/blob/rawhide/config/file_contexts.subs_dist#L35

PopiBrossard:
Thanks for the quick reply.

I'm running mariadb on a RockyLinux 9 server, which doesn't seems to implement this unification between bin and sbin. I thought this github repository was for fedora but also centos & redhat-like OS, am I wrong ?

I'm not 100% sure the /config directory, because I need to add an extra fcontext for /usr/sbin/mariadbd otherwise it's unconfined. If I create an empty file in `/usr/bin` SELinux sets the right context:
```
root@server: touch /usr/bin/mariadbd
root@server: restorecon /usr/bin/mariadbd
root@server: ls -lZ /usr/bin/mariadbd
-rw-r--r--. 1 root root unconfined_u:object_r:mysqld_exec_t:s0 0 May 22 15:50 /usr/bin/mariadbd
```

zpytela:
You can use the c9s branch in this repository, but I am afraid even there if you want to use some custom software, you also need to handle the needed changes in a custom module. The c9s policy expect /usr/sbin/mariadbd be a symlink and then the bin_t type is correct. Types for /usr/bin/mariadbd and /usr/libexec/mariadbd are correct.

PopiBrossard:
I agree types for /usr/bin/mariadbd and /usr/libexec/mariadbd are correct, and these are the path used by the package from the OS repository (the C9s dnf repository for example).

Shouldn't the SELinux policy also confine the mariadb server if installed from the official repository given by mariadb project itself ? The link i gave before (mirror.mariadb.org) is the official repository given by the MariaDB foundation.

The version MariaDB foundation give for fedora also put mariadbd inside /usr/sbin, but because sbin seems to be a symlink to bin in fedora I'm not sure it's affected.

zpytela:
> Shouldn't the SELinux policy also confine the mariadb server if installed from the official repository given by mariadb project itself ? The link i gave before (mirror.mariadb.org) is the official repository given by the MariaDB foundation.

We don't do this, instead there is the equivalency defined. Moreover, with the equivalency in place, no individual /usr/sbin entry applies.

PR #2712, opened 2025-05-31
Clean up locations for 'random-seed' file

keszybz:


zpytela:
So I guess this is of no use too now:
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/init.if#L3472

keszybz:
On Mon, Jun 02, 2025 at 02:58:24AM -0700, Zdeněk Pytela wrote:
> So I guess this is of no use too now:
> https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/init.if#L3472

Indeed. I updated the PR.

zpytela (approved):
LGTM

keszybz:
In the bodhi update https://bodhi.fedoraproject.org/updates/FEDORA-2025-3fc35b11d8 the coreos test that checks whether all files are labeled as expected is failing.

I installed the build from the update in a VM, and indeed, I get `system_u:object_r:init_var_lib_t:s0` instead of `random_seed_t`. One thing I noticed is that we have
```
files_var_lib_filetrans($1, random_seed_t, file, "random_seed")
```
(not "random-seed"). I tried with that changed and it didn't seem to make a difference.

I'm confused why the file is created with a dfiferent context that what `matchpathcon /var/lib/systemd/random-seed` prints.

zpytela:
- random_seed in the interface clearly is a mistake
- fc rules are independent of policy rules
- is the seed file always created by systemd-random-seed.service? it anyway needs to be confined

dustymabe:
xref: https://github.com/coreos/fedora-coreos-tracker/issues/1965

PR #2714, opened 2025-06-03
Remove pcp module

jsegitz:
The pcp module is shipped with pcp:
https://github.com/performancecopilot/pcp/tree/main/src/selinux


jsegitz:
The module in the package isn't great, but the one here will be shadowed by the pcp nodule from the package, which is confusing.

zpytela:
It is needed to keep the if file because the interfaces are used by other modules.

jsegitz:
yeah stumbled over this myself. I always forget that optional blocks don't solve this, because in my mind they should. I really don't like these out of tree modules and having to keep the .if files in the main policy is really a major part in that, but it is what it is

zpytela:
At the moment it is the current state and no other option is known, but do you think this should be better documented?

In the commit, there should also be ifndef statements for pcp to compile, refer to another example:
https://github.com/fedora-selinux/selinux-policy/commit/866309092d

This is our helper tool:
https://src.fedoraproject.org/rpms/selinux-policy/blob/rawhide/f/ifndefy.py

jsegitz:
Thanks for the link to the tool, I used it and pushed the changes.

Describing this better is certainly helpful, but in the end I mostly just disagree with the decision to move modules to the packages. Having to keep the .if files in the policy is pretty ugly and as seen with the pcp module the resulting modules aren't that great.

But that's a bigger topic, maybe we can discuss it in our next meeting

zpytela:
Merging, thank you.

PR #2716, opened 2025-06-04
systemd: rework systemd_manage_random_seed

keszybz:
Follow-up for 2497483ae16e172a5fd361c7b09c6152c87144a1. When the service created the file, it'd have init_var_lib_t, not random_seed_t as expected. It seems that the transition rule didn't match, because /var/lib/systemd is init_var_lib_t, not var_lib_t. (The transition rule made sense when the path was /var/lib/random-seed.)

With this change, when the file is created at boot, it gets the expected context.

Fixes issue reported in
https://bodhi.fedoraproject.org/updates/FEDORA-2025-3fc35b11d8.

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -1657,7 +1657,7 @@ interface(`systemd_manage_random_seed',`
 	')
 
 	allow $1 random_seed_t:file manage_file_perms;
-	files_var_lib_filetrans($1, random_seed_t, file, "random_seed")
+	init_var_lib_filetrans($1, random_seed_t, file, "random-seed");
 ')
 
 ########################################


policy/modules/system/systemd.if
```diff
@@ -1657,7 +1657,7 @@ interface(`systemd_manage_random_seed',`
 	')
 
 	allow $1 random_seed_t:file manage_file_perms;
-	files_var_lib_filetrans($1, random_seed_t, file, "random_seed")
+	filetrans_pattern($1, init_var_lib_t, random_seed_t, file)
```
zpytela: init is a different module, so this should be init_var_lib_filetrans(), in an optional block, and with the correct filename

Once sd-random-seed is confined, it would not matter, but as of now it applies to all files created by systemd. Or is there a problem with a temporary file name?

keszybz: I'm not sure if I understood your comment correctly. I updated the patch to use `init_var_lib_filetrans`. In the same file `init_var_lib_filetrans` is also used in `systemd_filetrans_named_content`, I tried to make the new use similar to the previous one.

I rebuilt the policy rpm with this version and it still seems to work as expected.

zpytela:
LGTM, thank you, let's have a new build.

PR #2726, opened 2025-06-06
Define file equivalency for /usr/etc

jcpunk:
/usr/etc is used by a number of immutable/sealed /usr systems. This permits easily keeping the file contexts up to date.

### DIFF
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -32,4 +32,5 @@
 /var/usrlocal        /usr/local
 /var/mnt             /mnt
 /bin                 /usr/bin
+/usr/etc             /etc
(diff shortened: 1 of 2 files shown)


zpytela:
Please remove the existing /usr/etc entry. Otherwise LGTM.

jcpunk:
> Please remove the existing /usr/etc entry.

Where might I find that...

zpytela:
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/files.fc#L262

jcpunk:
Thanks!

Updated the PR

zpytela:
Merging, thank you.

Issue #2500, opened 2024-12-22
`run0` is almost unusable.


fxzxmicah:
I only tried the `restorecon` and `semodule` commands.

Log:
```
audit[108591]: AVC avc:  denied  { entrypoint } for  pid=108591 comm="(storecon)" path="/usr/sbin/setfiles" dev="nvme0n1p3" ino=754816 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:setfiles_exec_t:s0 tclass=file permissive=0
audit[116566]: AVC avc:  denied  { entrypoint } for  pid=116566 comm="(semodule)" path="/usr/sbin/semodule" dev="nvme0n1p3" ino=754814 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:semanage_exec_t:s0 tclass=file permissive=0
```

fxzxmicah:
```
audit[29322]: AVC avc:  denied  { entrypoint } for  pid=29322 comm="(userdel)" path="/usr/sbin/userdel" dev="nvme0n1p3" ino=588622 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:useradd_exec_t:s0 tclass=file permissive=0
audit[29611]: AVC avc:  denied  { entrypoint } for  pid=29611 comm="(groupdel)" path="/usr/sbin/groupdel" dev="nvme0n1p3" ino=588611 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:groupadd_exec_t:s0 tclass=file permissive=0
```

fxzxmicah:
```
audit[15468]: AVC avc:  denied  { entrypoint } for  pid=15468 comm="(dnf)" path="/usr/bin/dnf5" dev="nvme0n1p3" ino=924188 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:rpm_exec_t:s0 tclass=file permissive=0
[...]

PR #2736, opened 2025-06-20
Miscellaneous fixes for confined users

gucci-on-fleek:
I've been running using SELinux-confined users on my Fedora IoT server for a few months now, but I've needed a few custom rules for it to work properly. This patch contains those rules.

There are quite a few (very small) changes here, so please let me know if you have any questions or want me to make any changes.

Full disclosure: I've been using a [CIL version of these rules](https://github.com/gucci-on-fleek/maxchernoff.ca/blob/1d593bb8dfc281845c7efd7215d2da7b51c4eedf/etc/selinux/local-policies/local-systemd.cil) for a couple months now, and I've confirmed that the policies still build (`make all`) and validate (`make validate`) correctly, but I haven't tested installing the `.pp` modules from this patch on my server.

(Vaguely related to containers/container-selinux#383)

### DIFF
--- a/policy/modules/contrib/journalctl.te
+++ b/policy/modules/contrib/journalctl.te
@@ -61,3 +61,9 @@ userdom_rw_inherited_user_home_content_files(journalctl_t)
 optional_policy(`
 	rhcd_read_fifo_files(journalctl_t)
 ')
+
+# Needed for journalctl (with the "less" pager) to stop complaining about
+# "lesshst" files
+optional_policy(`
+	gnome_dontaudit_read_config(journalctl_t)
+')
--- a/policy/modules/contrib/snapper.te
+++ b/policy/modules/contrib/snapper.te
@@ -98,3 +98,7 @@ optional_policy(`
 optional_policy(`
 	systemd_exec_systemctl(snapperd_t)
 ')
+
+optional_policy(`
+    seutil_dontaudit_read_file_contexts(snapperd_t)
+)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -126,6 +126,8 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 
 /run/systemd/generator		-d	gen_context(system_u:object_r:systemd_unit_file_t,s0)
 /run/systemd/generator/.+	<<none>>
+/run/user/%{USERID}/systemd/generator		-d	gen_context(system_u:object_r:systemd_unit_file_t,s0)
+/run/user/%{USERID}/systemd/generator/.+	<<none>>
 
 /run/systemd/io\.systemd\.Login	-s	gen_context(system_u:object_r:systemd_logind_var_run_t,s0)
 /run/systemd/io\.systemd\.NamespaceResource	-s	gen_context(system_u:object_r:systemd_nsresourced_runtime_t,s0)
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -2845,6 +2845,27 @@ interface(`systemd_map_resolved_exec_files',`
 
 ########################################
 ## <summary>
+##	Send and receive messages from
+##	systemd networkd over dbus.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_dbus_chat_networkd',`
+	gen_require(`
+		type systemd_networkd_t;
+		class dbus send_msg;
+	')
+
+	allow $1 systemd_networkd_t:dbus send_msg;
+	allow $1 systemd_networkd_t:unix_stream_socket connectto;
+')
+
+########################################
+## <summary>
 ##	Exchange messages with
 ##	systemd resolved over dbus or varlink.
 ## </summary>
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -967,6 +967,11 @@ init_read_state(systemctl_domain)
 init_list_pid_dirs(systemctl_domain)
 init_use_fds(systemctl_domain)
 
+# Needed for "sudo systemctl --user --machine=USER@ ..."
+optional_policy(`
+	userdom_rw_admindom_stream()
+')
+
 #######################################
 #
 # Localed policy
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -19,6 +19,7 @@
 /usr/local/lib64 /usr/lib
 /usr/local/lib32 /usr/lib
 /etc/systemd/system /usr/lib/systemd/system
+/etc/systemd/user /usr/lib/systemd/user
 /var/lib/xguest/home /home
 /var/named/chroot/usr/lib64 /usr/lib
 /var/named/chroot/lib64 /usr/lib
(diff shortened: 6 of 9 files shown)


policy/modules/system/systemd.if
```diff
@@ -2843,6 +2843,29 @@ interface(`systemd_map_resolved_exec_files',`
...
+interface(`systemd_dbus_chat_networkd',`
+	gen_require(`
+		type systemd_networkd_t;
+		class dbus send_msg;
+	')
+
+	allow $1 systemd_networkd_t:dbus send_msg;
+	allow systemd_networkd_t $1:dbus send_msg;
+	allow $1 systemd_networkd_t:unix_stream_socket connectto;
+	ps_process_pattern(systemd_networkd_t, $1)
```
zpytela: This looks like 2 or 3 different permission sets. Can you elaborate for what is that needed?

policy/modules/contrib/snapper.te
```diff
@@ -98,3 +98,7 @@ optional_policy(`
 optional_policy(`
 	systemd_exec_systemctl(snapperd_t)
 ')
+
+optional_policy(`
+    seutil_read_file_contexts(snapperd_t)
```
zpytela: I am not convinced this should be added to policy.

policy/modules/contrib/journalctl.te
```diff
@@ -61,3 +61,14 @@ userdom_rw_inherited_user_home_content_files(journalctl_t)
...
 ')
+
+# Needed for journalctl (with the "less" pager) to stop complaining about
+# "lesshst" files
+optional_policy(`
+	gen_require(`
+		type gconf_home_t;
+	')
+
+	dontaudit journalctl_t gconf_home_t:dir { add_name remove_name write };
```
zpytela: For using types from other modules, interfaces need to be used.

policy/modules/system/systemd.fc
```diff
@@ -126,6 +126,8 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 
 /run/systemd/generator		-d	gen_context(system_u:object_r:systemd_unit_file_t,s0)
 /run/systemd/generator/.+	<<none>>
+/run/user/%{USERID}/systemd/generator		-d	gen_context(system_u:object_r:systemd_unit_file_t,s0)
+/run/user/%{USERID}/systemd/generator/.+	<<none>>
```
zpytela, line 130: Why is it needed to label user units, or is it intended to have private types?

policy/modules/system/systemd.te
```diff
@@ -237,6 +237,10 @@ files_type(systemd_zram_generator_conf_t)
 allow systemd_zram_generator_t systemd_zram_generator_conf_t:file read_file_perms;
 # domain for a generic generator
 systemd_generator_template(systemd_generic_generator)
+# User domains
+allow userdomain systemd_unit_file_t:dir manage_dir_perms;
+allow userdomain systemd_unit_file_t:file manage_file_perms;
+allow userdomain systemd_unit_file_t:lnk_file manage_lnk_file_perms;
```
zpytela: Types/attributes from other modules need to be accessed via interfaces. Anyway allowing any user type manage systemd_unit_file_t does not look good.

policy/modules/system/systemd.te
```diff
@@ -971,6 +971,15 @@ init_read_state(systemctl_domain)
 init_list_pid_dirs(systemctl_domain)
 init_use_fds(systemctl_domain)
 
+# Needed for "sudo systemctl --user --machine=USER@ ..."
+optional_policy(`
+	gen_require(`
+		type sysadm_t;
+	')
+
+	allow userdomain sysadm_t:unix_stream_socket { getattr getopt read setopt write };
```
zpytela: Only staff_t and admindomains are allowed to run sudo. Interface needs to be used.

zpytela (changes requested):

policy/modules/system/systemd.if
gucci-on-fleek: I copied this from `systemd_chat_resolved` almost exactly, except with `systemd_networkd_t` instead of `systemd_resolved_t`. My personal testing showed that 

```cil
(allow sysadm_t systemd_networkd_t (dbus               (send_msg )))
(allow sysadm_t systemd_networkd_t (unix_stream_socket (connectto)))
```

was sufficient to make this work though, so I can reduce the permissions to that if you want.

policy/modules/contrib/snapper.te
gucci-on-fleek: It seems to be required if you're using snapper to snapshot `/etc/` with unconfined disabled. Specifically snapshotting `/etc/` isn't very common, but snapshotting all of `/` is (and `/etc/` is included in that). But I can remove this if you want.

policy/modules/contrib/journalctl.te
gucci-on-fleek: Ah, I had no idea about that, sorry. I'll fix that for the next push.

policy/modules/system/systemd.fc
gucci-on-fleek: `systemctl --user` can't access/manipulate them otherwise.

policy/modules/system/systemd.te
gucci-on-fleek: `~/.config/systemd/user/` is labelled as `systemd_unit_file_t`

https://github.com/fedora-selinux/selinux-policy/blob/66492b704f2e5d365a701ce05b8623764db93aea/policy/modules/system/systemd.fc#L2

so this is needed for confined users to be able to edit the files there with a text editor. So if you don't want for users to be able to edit all `systemd_unit_file_t`s, maybe we should add a new `systemd_user_unit_file_t`?

policy/modules/system/systemd.te
gucci-on-fleek: > Only staff_t and admindomains are allowed to run sudo.

Right, but this is needed for the confined user session created with `--machine=USER@` to be able to connect back to the `systemctl` command. Or should we maybe create a new label for that socket file so that we don't need to use `sysadm_t`?

policy/modules/system/systemd.if
gucci-on-fleek: Ok, I've narrowed `systemd_chat_resolved` down to only the needed permissions now.

policy/modules/contrib/snapper.te
gucci-on-fleek: I was wrong above; replacing this with `seutil_dontaudit_read_file_contexts` fixes the problem.

policy/modules/contrib/journalctl.te
gucci-on-fleek: Ok, this should be fixed now.

policy/modules/system/systemd.te
gucci-on-fleek: Ok, I tried to define new `systemd_stdio_bridge_t` and `systemd_stdio_bridge_exec_t` types, label `/usr/bin/systemd-stdio-bridge` as `systemd_stdio_bridge_exec_t`, and then let only `systemd_stdio_bridge_t` be allowed to read/write `admindomain` `unix_stream_socket`s, but that didn't work, because [systemd runs `systemd-stdio-bridge` via `systemd-run`](https://github.com/systemd/systemd/blob/391ad5d8aa533010bed02079c95eab34de0408c5/src/libsystemd/sd-bus/sd-bus.c#L1614-L1622), and processes do not transition correctly when ran that way, so that solution is blocked by https://github.com/fedora-selinux/selinux-policy/issues/2500 and https://github.com/systemd/systemd/issues/32668.

So I've left the current functionality in, but I've changed it to use the proper interfaces now.

gucci-on-fleek:
(sorry for the long delay)

I've updated the code to use interfaces for all external types, and I've addressed most of the other comments. For the comments that I haven't addressed, I've replied with a few questions because I'm not sure how to proceed. And thanks again for the review!

PR #2758, opened 2025-07-03
virtqemud_t fixes from cockpit tests (bsc#1242998)

ca-hu:
These AVCs were found during cockpit tests on SUSE, might be interesting for fedora as well

Use case:
- set up cockpit
- on cockpit web interface create and start a virtual machine
- see denials in audit log

Versions:
- libvirt 11.2.0
- cockpit 334
- policy: opensuse selinux-policy with fedora commits until 23514206ea45e1d1d2f8a4c08288065c813fcc91

### DIFF
--- a/policy/modules/contrib/virt.te
+++ b/policy/modules/contrib/virt.te
@@ -396,6 +396,9 @@ files_lock_file(virtqemud_lock_t)
 type virtqemud_tmp_t;
 files_tmp_file(virtqemud_tmp_t)
 
+type virtqemud_tmpfs_t;
+files_tmpfs_file(virtqemud_tmpfs_t)
+
 type virtqemud_var_run_t, virt_driver_var_run;
 files_pid_file(virtqemud_var_run_t)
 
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/virt.te
```diff
@@ -2292,6 +2293,7 @@ fs_setattr_tmpfs_chr_blk_files(virtqemud_t)
 fs_relabel_tmpfs_blk_file(virtqemud_t)
 fs_relabel_tmpfs_chr_file(virtqemud_t)
 fs_relabelfrom_tmpfs_lnk_files(virtqemud_t)
+fs_rw_inherited_tmpfs_files(virtqemud_t)
```
zpytela: Which service created the memfd:displaysurface file?

ca-hu: qemu creates it as far as i could see: https://github.com/qemu/qemu/blob/e240f6cc25917f3138d9e95e0343ae23b63a3f8c/ui/console.c#L494

zpytela: So it can use its private virtqemud_tmpfs_t type, right?

policy/modules/contrib/virt.te
```diff
@@ -2307,6 +2309,9 @@ storage_manage_fixed_disk(virtqemud_t)
 sysnet_exec_ifconfig(virtqemud_t)
 sysnet_manage_config(virtqemud_t)
 
+term_use_generic_ptys(virtqemud_t)
+term_use_ptmx(virtqemud_t)
+
```
zpytela, line 2319: Is there a specific use case when these terms are needed? Is it worth having it behind a boolean?

policy/modules/contrib/virt.te
ca-hu: yep changed it, thanks :)

policy/modules/contrib/virt.te
ca-hu: its default that the serial console is connected to /dev/pts/x, so i dont think it is a special use case

zpytela:
I believe all issues have been resolved now.

ca-hu:
yes, thank you :)

zpytela:
Merging, thank you.

PR #2771, opened 2025-07-15
Fix kernel_secretmem_use() and add kernel_secretmem_use_inherited()

zpytela:
Fix the existing kernel_secretmem_use() interface content so that it matches other anon_inode types related interfaces and add kernel_secretmem_use_inherited().

### DIFF
--- a/policy/modules/kernel/kernel.if
+++ b/policy/modules/kernel/kernel.if
@@ -4749,6 +4749,24 @@ interface(`kernel_secretmem_domtrans',`
 
 ########################################
 ## <summary>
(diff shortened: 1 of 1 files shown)


WOnder93 (approved):
I guess this is related to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=cbe4134ea4bc493239786220bd69cb8a13493190 - indeed now more permissions can be checked on top of just `create`.

zpytela:
OK, thanks, so I updated the commit message.

PR #2780, opened 2025-07-18
Introduce systemd_cryptsetup_generator_var_run_t file type (bsc#1244459)

jsegitz:
When the key material is on a USB stick this currently doesn't work since cryptsetup will create a directory with a generic type

Solves avc:  denied  { associate } for  pid=16385 comm="systemd-cryptse" name="cryptsetup" scontext=system_u:object_r:systemd_cryptsetup_generator_var_run_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=0

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1404,6 +1404,15 @@ optional_policy(`
 
 #manage_files_pattern(systemd_cryptsetup_generator_t, systemd_fstab_generator_unit_file_t, systemd_fstab_generator_unit_file_t)
 
+type systemd_cryptsetup_generator_var_run_t;
+files_type(systemd_cryptsetup_generator_var_run_t)
+
+init_pid_filetrans(systemd_cryptsetup_generator_t, systemd_cryptsetup_generator_var_run_t, dir, "cryptsetup")
+
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:dir manage_dir_perms;
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:file manage_file_perms;
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:lnk_file manage_lnk_file_perms;
+
 ### debug generator
 fs_read_tmpfs_files(systemd_debug_generator_t)
 


policy/modules/system/systemd.te
```diff
@@ -1404,6 +1404,17 @@ optional_policy(`
...
+type systemd_cryptsetup_generator_var_run_t;
+files_type(systemd_cryptsetup_generator_var_run_t)
+
+init_var_lib_filetrans(systemd_cryptsetup_generator_t, systemd_cryptsetup_generator_var_run_t, dir, "cryptsetup")
+
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:dir manage_dir_perms;
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:file manage_file_perms;
+allow systemd_cryptsetup_generator_t systemd_cryptsetup_generator_var_run_t:lnk_file manage_lnk_file_perms;
+
+fs_associate_ramfs(systemd_cryptsetup_generator_var_run_t)
```
zpytela: ramfs has been deprecated in favor of tmpfs in policy since commit 8e908b8d98

Anyway I am even more confused: tmpfs_t, but the type is ..._var_run_t, and filetrans interface is init_var_lib_filetrans()? Where is the dir actually created?

jsegitz: The directory gets created in /run/systemd/, so the type should be fine

The fs_associate_ramfs call isn't necessary on my system, but the reporter has a special setup. I'll remove it, so we can get the basic change in there and then sent another PR just for the tmpfs change

zpytela: I believe init_pid_filetrans() is the right match given
```
2736 ## <summary>
2737 ##  Create objects in /run/systemd directory
2738 ##  with an automatic type transition to
2739 ##  a specified private type.
2740 ## </summary>

$ ls -lZd /run/systemd
drwxr-xr-x. 26 root root system_u:object_r:init_var_run_t:s0 720 27. srp 09.34 /run/systemd
```

The cryptsetup dir content is not used outside this generator?

jsegitz: sorry I didn't see the comment. I'll give init_pid_filetrans a try, but I can only do this after next week because my test vm needs an USB dongle attached, which I currently can't plug into the machine. Once I have access to the machine again I'll give it a go

jsegitz: yes `init_pid_filetrans` is the better match, changed to that.

I don't know of any users of the cryptsetup dir as of now

jsegitz: let me tag @zpytela to that he sees the change :)

zpytela:
Merging, thank you.

PR #2785, opened 2025-07-21
Remove the mysql module sources

zpytela:
The package now maintains its own selinux policy module.


policy/modules/contrib/mysql.if
vmojzis, line 1: Please replace the header of mysql.if file with: 
```
########################################
#
# MySQL module interfaces left over for backwards compatibility
#
# For current MySQL SELinux policy please see: https://github.com/devexp-db/mysql-selinux
```

PR #2788, opened 2025-07-23
Label /var/lib/lastlog with lastlog_t

zpytela:
With the [1] Fedora change in place, lastlog database has a new path: /var/lib/lastlog instead of /var/log/lastlog.

[1] https://fedoraproject.org/wiki/Changes/Migrate_to_lastlog2


keszybz (commented):
Post-merge, but LGTM ;)

PR #2793, opened 2025-07-25
introduce initrc_nnp_daemon_domain interface and fix locate systemd service AVCs

jsegitz:
when you run shell commands in ExecStart this is needed. This is the case for our locate systemd service, so use it there

### DIFF
--- a/policy/modules/contrib/slocate.te
+++ b/policy/modules/contrib/slocate.te
@@ -9,6 +9,7 @@ type locate_t;
 type locate_exec_t;
 init_system_domain(locate_t, locate_exec_t)
 init_nnp_daemon_domain(locate_t)
+initrc_nnp_daemon_domain(locate_t)
 
 type locate_var_lib_t;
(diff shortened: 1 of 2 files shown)


zpytela:
I wonder what is the use case for having initrc domains today?

jsegitz:
When a packager uses a shell construct in ExecStart this is run in initrc. Yes, I know that this is not a good way of doing things, but we still have cases like this and it will take some time until we get rid of them

jsegitz:
@zpytela I fear that these constructs will be with us for a while, so it would be nice to have an interface for this

jsegitz:
@zpytela can you please have another look? If that's not okay for you then I'll try to find another solution for this

jsegitz:
@zpytela If you're not okay with this it's fine for me, but it would be great to get a decision here. Thanks

zpytela:
Merging, thank you.

PR #2813, opened 2025-08-04
Update kmscon policy module to kmscon version 9 (bsc#1238137)

ca-hu:
We have some downstream commits in the opensuse selinux policy regarding kmscon, after a developer notified us that
the development will be continued in a fork: https://github.com/Aetf/kmscon
So maybe you want to have the kmscon commits from us as well so that they can take it?

Because: I saw that they also added a custom module for fedora 42, maybe it would make sense to ask them to take this kmscon module instead....
see: https://github.com/Aetf/kmscon/pull/109/files

Also we are enabling the module downstream in openSUSE, let me know if i should add this commit as well.
(not sure what fedoras plans are with kmscon, i think it is dropped https://src.fedoraproject.org/rpms/kmscon)

If it is not needed, feel free to close :)

### DIFF
--- a/dist/targeted/modules.conf
+++ b/dist/targeted/modules.conf
@@ -1247,6 +1247,13 @@ keystone = module
 # 
 kismet = module
 
+# Layer: contrib
+# Module: kmscon
+#
+# Linux KMS/DRM based virtual Console Emulator
+#
+kmscon = module
+
 # Layer: services
 # Module: ksmtuned
 #
--- a/policy/modules/contrib/kmscon.te
+++ b/policy/modules/contrib/kmscon.te
@@ -31,9 +31,13 @@ term_tty(kmscon_devpts_t)
 allow kmscon_t self:capability {sys_admin sys_tty_config};
 
 dontaudit kmscon_t self:capability2 block_suspend;
+# kmscon tries to read /proc/<pid_of_login_process>/cgroup
+# preliminarily dontaudit this since there is some discussion upstream
+# https://github.com/Aetf/kmscon/pull/105
+domain_dontaudit_read_all_domains_state(kmscon_t)
 
 # Create an udev monitor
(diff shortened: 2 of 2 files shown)


kdj0c:
I'm trying to unretire kmscon in Fedora https://bugzilla.redhat.com/show_bug.cgi?id=2373128

Currently I install the selinux rules from the kmscon repository, but I think it would be easier to use this module instead.

zpytela:
I'm fine with merging this PR, but be aware that at some point we will review all existing modules and remove those which are for software not currently in Fedora.

ca-hu:
Thanks, I assume that means you don't want to have the module activated yet in `dist/targeted/modules.conf` until the [kmscon review request](https://bugzilla.redhat.com/show_bug.cgi?id=2373128) is through?

If so, then @kdj0c would know best how far the process is. 

@kdj0c you would likely need [this commit](https://github.com/openSUSE/selinux-policy/commit/0bf20044876d40a27c7a35fe621e7fb927c0ddac) as well to activate the kmscon module in the fedora selinux-policy, maybe you could create a PR here for this when the time is right? (please note: i dont know how the fedora review process works)

zpytela:
@ca-hu no, I can merge it right away, especially if there are good prospects to have the package back in Fedora. I haven't even noticed the module is not enabled in modules.conf.

ca-hu:
okay, thanks for the clarification, added the commit to this PR

zpytela (approved):

zpytela:
Merging, thank you.

kdj0c:
Thank you both, I will try to use that in the kmscon package.

kdj0c:
Hi, there is something missing in the selinux policy for kmscon.

You can see the detail here:
https://discussion.fedoraproject.org/t/f44-change-proposal-usekmsconvtconsole-systemwide/172602/18

ca-hu:
https://github.com/fedora-selinux/selinux-policy/pull/2946

PR #2820, opened 2025-08-07
Allow systemd to map files under /sys to fix AVC

keszybz:


keszybz:
`No match for argument: policycoreutils-devel >= 3.9` — somebody broke F42, but fortunately that wasn't me ;)

zpytela:
Merging, thank you. Nice findings.

PR #2842, opened 2025-08-20
allow tuned-ppd getattr on /

JJanasek:


zpytela:
LGTM for the content, but why is the other commit here?

zpytela:
Merging, thank you.

PR #2868, opened 2025-09-11
Add a new type for systemd-ssh-issue PID files

vmojzis:
systemd-ssh-issue creates /run/issue.d/50-ssh-vsock.issue, which is then read by agetty.

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1142,7 +1142,15 @@ optional_policy(`
 
 permissive systemd_ssh_issue_t;
 
+type systemd_ssh_issue_var_run_t;
+files_pid_file(systemd_ssh_issue_var_run_t)
+
 allow systemd_ssh_issue_t self:vsock_socket create_socket_perms;
+
+# /run/issue.d/50-ssh-vsock.issue
+manage_dirs_pattern(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, systemd_ssh_issue_var_run_t)
+manage_files_pattern(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, systemd_ssh_issue_var_run_t)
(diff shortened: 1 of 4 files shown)


policy/modules/system/systemd.te
```diff
@@ -1142,8 +1142,17 @@ optional_policy(`
...
+type systemd_ssh_issue_var_run_t;
+files_pid_file(systemd_ssh_issue_var_run_t)
+
 allow systemd_ssh_issue_t self:vsock_socket create_socket_perms;
 
+# /run/issue.d/50-ssh-vsock.issue
+manage_dirs_pattern(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, systemd_ssh_issue_var_run_t)
+manage_files_pattern(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, systemd_ssh_issue_var_run_t)
+manage_sock_files_pattern(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, systemd_ssh_issue_var_run_t)
+files_pid_filetrans(systemd_ssh_issue_t, systemd_ssh_issue_var_run_t, { dir file sock_file })
```
zpytela: Does sock_file appear anywhere? Based on my observation so far, filetrans is needed only for dir and manage pattern only for files.

vmojzis: No, I wasn't sure whether to include sock_file since this feature is about vsock functionality. But the man page only talks about file, so I'll remove it. My testing shows that systemd-ssh-issue needs to be able to create the /run/issue.d/ directory, so manage pattern needs to cover both files and dirs.

vmojzis:
Updated.

policy/modules/system/systemd.te
zpytela: Right, filetrans_pattern does not support create as I thought previously. LGTM then

zpytela:
Merging, thank you.

PR #2874, opened 2025-09-18
Allow cloud init to domtrans into ssh keygen (bsc#1249964)

jsegitz:
Ohterwise creating ssh keys via cloud init will result in type etc_t for the key files

### DIFF
--- a/policy/modules/contrib/cloudform.te
+++ b/policy/modules/contrib/cloudform.te
@@ -159,8 +159,9 @@ optional_policy(`
 ')
 
 optional_policy(`
-    ssh_exec_keygen(cloud_init_t)
     ssh_read_user_home_files(cloud_init_t)
+    ssh_domtrans_keygen(cloud_init_t)
+
 ')
 
 optional_policy(`


policy/modules/contrib/cloudform.te
```diff
@@ -128,6 +128,10 @@ optional_policy(`
 	rhsmcertd_dbus_chat(cloud_init_t)
 ')
 
+optional_policy(`
+    ssh_domtrans_keygen(cloud_init_t)
```
zpytela: This should be together with other ssh interfaces in one optional block. If you move it, you can notice ssh_exec_keygen() is now ineffective, so I need to ask for justification.

jsegitz: thanks, I got bitten by assuming the interface calls were ordered and overlooked the other interface calls. Justification is that the key files created upon the first boot when using this mechanism will have etc_t as type instead of the dedicated type. Doesn't cause issues AFAICS

jsegitz: @zpytela is that acceptable to you?

zpytela: LGTM, but I am kind-of reluctant pushing it to current releases as we only have basic test coverage, so cannot evaluate the impact.

jsegitz: Understood. Works on SLE 16 cloud images, but unfortunately they're not public yet, so I can't make them available. There's no hurry, can be kept here until after the release

ca-hu:
waiting for policy split upstream, then it will be merged

jsegitz:
@zpytela back after a longer sick leave. Is the policy split done so that this can be merged? Thanks

zpytela:
Merging, thank you for your contribution and patience.

jsegitz:
thanks for merging this :)

PR #2877, opened 2025-09-19
Systemd sysctl fixes

WavyEbuilder:


zpytela:
When exactly is this denial triggered, or what is the use case?

WavyEbuilder:
Both should be in commit description (not sure why it wasn't put in the Github description box, sorry more of a patches guy):
```
systemd-sysctl: allow rw on binfm_misc_fs_t to set binfmt_misc status
Used for fs.binfmt_misc.status sysctl: /proc/sys/fs/binfmt_misc/status.

audit[938]: AVC avc:  denied  { read write } for  pid=938 comm="systemd-sysctl" name="status" dev="binfmt_misc" ino=2 scontext=system_u:system_r:systemd_sysctl_t:s0 tcontext=system_u:object_r:binfmt_misc_fs_t:s0 tclass=file permissive=0

Signed-off-by: Rahul Sandhu <nvraxn@gmail.com>
```

Tl;dr if you have the `fs.binfmt_misc.status` sysctl set in a sysctl config file, systemd-sysctl will try and load it, which requires writing to that file.

WavyEbuilder:
@zpytela Ping

zpytela:
Sorry for the delay, I considered assigning the type to sysctl_type instead, but rather not.

PR #2878, opened 2025-09-19
auditd: don't grant write as implied by manage_files_pattern for logs

WavyEbuilder:
auditd doesn't actually need to be able to write logs, only create, append, read, rename, and setattr them. Given that great lengths are already taken to ensure audit log confidentiality and integrity (e.g. marking as mls_systemhigh and granting cap_sys_nice to prioritise over other processes to not miss audit events), it makes sense to not grant an unnecessary permission which would allow a comprimised audit daemon to tamper with the audit logs.


WavyEbuilder:
@zpytela Ping

WavyEbuilder:
Any updates on this one?

WavyEbuilder:
Ping

Issue #2864, opened 2025-09-09
`init_t` is denied access to `vfio_device_t`


eaibmz:
Hi,
i wrote a systemd service which uses the directive `OpenFile=` on Fedora 42.
The service unit also sets "User=" and "Group=" to non-root IDs.
systemd fails to start the service with:
```
AVC avc:  denied  { read write } for  pid=73120 comm="(myservice)" name="sclp" dev="devtmpfs" ino=170 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:vfio_device_t:s0 tclass=chr_file permissive=0
```
In my understanding the systemd should open the file specified by `OpenFile=` and pass it to the service because the service has no privileges to access the device.

eaibmz:
I think the following statement must be added to fix the problem:
```
dev_rw_vfio_dev(init_t)
```

@zpytela What do you think ?

zpytela:
Can you include some details? The full service content and ausearch with full auditing enabled.
https://fedoraproject.org/wiki/SELinux/Debugging#Enable_full_auditing

eaibmz:
### systemd service

```
[Unit]
Description=SCLP Daemon
ConditionPathExists=/dev/sclp
After=network.target

[Service]
Type=simple
User=tester
Group=tester
ExecStart=/bin/bash -c 'echo Hello world'
DeviceAllow=/dev/sclp
OpenFile=/dev/sclp

[Install]
WantedBy=default.target
```

systemd's documentation for `OpenFile` directive: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#OpenFile=

### Test

Start service:
```
systemctl start sclp
```

Journal output:
```
systemd[1]: Started sclp.service - SCLP Daemon.
(bash)[59584]: sclp.service: Failed to get OpenFile= file descriptor for '/dev/sclp': Permission denied
(bash)[59584]: sclp.service: Failed to get OpenFile= file descriptors: Permission denied
(bash)[59584]: sclp.service: Failed at step FDS spawning /bin/bash: Permission denied
systemd[1]: sclp.service: Main process exited, code=exited, status=202/FDS
systemd[1]: sclp.service: Failed with result 'exit-code'.
```

ausearch:
```
❯ ausearch -m AVC -ts recent
----
time->Tue Sep 16 15:53:51 2025
[...]

PR #2892, opened 2025-10-02
Remove vfio_device_t label from /dev/sclp[0-9]* and /dev/vmcp[0-9]*

eaibmz:
/dev/sclp[0-9]* and /dev/vmcp[0-9]* are s390x-specific character devices which are unrelated to VFIO. Therefore, change the label back to device_t.

Fixes: 61939dae5285 ("Add support for /dev/vmcp and /dev/sclp")

Related issue: https://github.com/fedora-selinux/selinux-policy/issues/2864

### DIFF
--- a/policy/modules/kernel/devices.fc
+++ b/policy/modules/kernel/devices.fc
@@ -168,8 +168,6 @@ ifdef(`distro_suse', `
 /dev/vc-mem		-c	gen_context(system_u:object_r:memory_device_t,mls_systemhigh)
 /dev/vfio/(vfio)?[0-9]*	-c	gen_context(system_u:object_r:vfio_device_t,s0)
 /dev/clp[0-9]*            -c  gen_context(system_u:object_r:vfio_device_t,s0)
-/dev/sclp[0-9]*            -c  gen_context(system_u:object_r:vfio_device_t,s0)
-/dev/vmcp[0-9]*     -c  gen_context(system_u:object_r:vfio_device_t,s0)
 /dev/vhost-net		-c	gen_context(system_u:object_r:vhost_device_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
@WOnder93 What do you think would be appropriate label for these device files? The current version was added in 61939dae528 and b1f7f8e3435 (plus a fix in 2ce8aee582b).

WOnder93:
`/dev/sclp` (I believe it doesn't have numbered suffixes, just a single device) is a simple ioctl-only device that supports just one ioctl. It's apparently used as a control device for the IBM SCLP terminal. Seems to be very niche and could probably have its own label.

`/dev/vmcp` (I think this one isn't numbered either) is for communicating with the "z/VM control program" (i.e. for z/VM virtual machines to communicate with the hypervisor). I'd say it's likewise pretty niche and could have its own label.

`/dev/clp` is also s390x-specific, but unrelated to `/dev/sclp`. It appears to be some control device for PCI bus on s390x machines. Likewise should probably have its own label.

Or they could all be `misc_device_t` (in the kernel they are all registered as a "misc" device), but that seems to be unused currently.

eaibmz:
The reason i changed the label to `device_t` is because the current label `vfio_device_t` breaks systemd's
device access (https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#DeviceAllow=).

We would like to implement a systemd service which runs as a non-root user but give it access to `/dev/sclp`.
And SELinux prevents this from working due to the label `vfio_device_t`. With `device_t` it works as expected.

Do you have another idea how to fix this ?

eaibmz:
Am i right in assuming that
```
dev_rw_generic_chr_files(init_t)
```
and
```
systemd_exec_systemctl(init_t)
```
together allow systemd to open a generic char device
but at the same time deny opening nodes labeled with `vfio_device_t` ?

PR #2908, opened 2025-10-14
Set temporary no-stub resolv.conf file from NetworkManager as net_conf_t

bengal:
Currently NetworkManager writes
`/run/NetworkManager/no-stub-resolv.conf `by using `mkstemp()`, which creates a temporary file name with a random suffix and then renames it. The policy doesn't handle this, and the file gets wrongly labeled as `NetworkManager_var_run_t`.

NetworkManager is being updated to write the file with a fixed ".tmp" suffix here:

  https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/2298

Adapt the policy accordingly.

### DIFF
--- a/policy/modules/system/sysnetwork.if
+++ b/policy/modules/system/sysnetwork.if
@@ -1198,6 +1198,7 @@ interface(`sysnet_filetrans_named_content',`
 
 	optional_policy(`
 		networkmanager_pid_filetrans($1, net_conf_t, file, "no-stub-resolv.conf")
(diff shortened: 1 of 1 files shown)


jsegitz:
The changed NM in combination with this policy change (available for openSUSE here https://build.opensuse.org/project/show/home:jsegitz:branches:security:SELinux_1248136) fixes the issue and the file has the proper label

zpytela:
LGTM, together with the change in nm it is going to prevent many future misconfigurations.

PR #2926, opened 2025-11-03
Fix syntax error in userdomain.if

rfrohl:
userdomain.if: Syntax error on line 6081 $1 [type=IDENTIFIER]
userdomain.if: Syntax error on line 6083 ' [type=SQUOTE]
userdomain.if: Syntax error on line 6089 ' [type=SQUOTE]


rfrohl:
fix confirmed with sepolgen-ifgen, not sure if there is another option.

zpytela:
Merging, thank you.

PR #2956, opened 2025-11-21
Allow virtqemud_t to read/write device_t (bsc#1251789)

jsegitz:
If you have VMs with macvtap networking you run into issues after a while (probably after ~30 VM restarts) because there are only transition rules for the first 30 tap devices to be created as tun_tap_device_t. After that they first get created as device_t, then are changed to tun_tap_device_t and later on to something like tun_tap_device_t:s0:c491,c870. To allow that virtqemud_t needs these permissions.

The whole story is in https://bugzilla.suse.com/show_bug.cgi?id=1251789, but it's a very long bug report and that's the gist

Solves
`avc:  denied  { read write } for  pid=8209 comm="rpc-virtqemud" name="tap32" dev="devtmpfs" ino=808 scontext=system_u:system_r:virtqemud_t:s0 tcontext=system_u:object_r:device_t:s0 tclass=chr_file permissive=1
avc:  denied  { open } for  pid=8209 comm="rpc-virtqemud" path="/dev/tap32" dev="devtmpfs" ino=808 scontext=system_u:system_r:virtqemud_t:s0 tcontext=system_u:object_r:device_t:s0 tclass=chr_file permissive=1
avc:  denied  { ioctl } for  pid=8209 comm="rpc-virtqemud" path="/dev/tap32" dev="devtmpfs" ino=808 ioctlcmd=0x54d2 scontext=system_u:system_r:virtqemud_t:s0 tcontext=system_u:object_r:device_t:s0 tclass=chr_file permissive=1`


zpytela:
No device should have `device_t` assigned, so no rules for access should be needed. In this case it probably is because auto file transitions are defined for tap0..tap29, where the proper label would be assigned immediately. In other cases, udev should take care of the label, but there can be noticable delay. One of the possible solutions is to implement a static delay, or a loop, or wait for udev in the application before processing further. Defining more file names in transitions in the policy does not scale.

Edit - link added:
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/corenetwork.if.in#L4421

jsegitz:
I agree. I see the code path which takes care of the relabeling, but I don't have a proper understanding of the libvirt codebase to fix this properly, so I'll open an upstream issue

PR #2975, opened 2025-12-01
Allow setroubleshoot_fixit to write Python cache in /usr/share

JJanasek:
The setroubleshoot fixit domain was denied the ability to create and write Python bytecode cache files (e.g., cpython-313.pyc) in the /usr/share directory, which is labeled usr_t.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2298271


bachradsusi:
This needs to be fixed on setroubleshoot packaging level, see https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_files_to_include

PR #2981, opened 2025-12-03
fix: unbreak thumbnailing for Thunar/tumblerd

RoyalOughtness:
Thumbnailing on Thunar is currently broken on a fresh, up-to-date Fedora Sway 43 VM: https://bugzilla.redhat.com/show_bug.cgi?id=2418633

### DIFF
--- a/policy/modules/contrib/thumb.te
+++ b/policy/modules/contrib/thumb.te
@@ -99,6 +99,7 @@ files_map_non_security_files(thumb_t)
 files_mounton_rootfs(thumb_t)
 files_watch_etc_dirs(thumb_t)
 files_watch_usr_dirs(thumb_t)
+files_mounton_generic_tmp_dirs(thumb_t)
 
 fs_getattr_all_fs(thumb_t)
 fs_read_dos_files(thumb_t)
@@ -108,6 +109,7 @@ fs_mmap_removable_files(thumb_t)
 fs_dontaudit_getattr_nsfs_files(thumb_t)
 fs_mounton_tmpfs(thumb_t)
 fs_all_mount_fs_perms_xattr_fs(thumb_t)
+fs_all_mount_fs_perms_tmpfs(thumb_t)
 
 auth_read_passwd(thumb_t)
 
@@ -134,6 +136,7 @@ sysnet_read_config(thumb_t)
 
 
 term_dontaudit_use_unallocated_ttys(thumb_t)
+term_mount_pty_fs(thumb_t)
 
 userdom_dontaudit_setattr_user_tmp(thumb_t)
 userdom_read_user_tmp_files(thumb_t)


policy/modules/contrib/thumb.te
```diff
@@ -46,6 +46,7 @@ allow thumb_t self:sem create_sem_perms;
 allow thumb_t self:cap_userns { net_admin setpcap sys_admin sys_ptrace };
 allow thumb_t self:process setcap;
 allow thumb_t self:user_namespace create;
+allow thumb_t tmp_t:dir mounton;
```
zpytela: It is no allowed to refer to types from other modules. If an appropriate interface does not exist, you need to create one - see https://github.com/fedora-selinux/selinux-policy/pull/2982

zpytela (changes requested):
Can you also add some details into the commit message?

RoyalOughtness:
@zpytela done

zpytela:
Can you specify which functional issues you had?

RoyalOughtness:
@zpytela Thumbnailing is entirely nonfunctional. No new thumbnails are created.

RoyalOughtness:
@zpytela Here's a freshly installed and updated Fedora Sway 43 VM. It reproduces consistently:

image

RoyalOughtness:
Installing a .cil (created based on the AVCs) with the following contents resolves the issue:

```
(allow thumb_t tmp_t (dir mounton)))
(allow thumb_t tmpfs_t (filesystem (mount unmount)))
(allow thumb_t devpts_t (filesystem (mount)))
```

zpytela:
Merging, thank you.

Mark-4158:
> Merging, thank you.

In my experience, the proper solution is to first make `container-selinux` a dependency of the `bubblewrap` package and to then switch the default SELinux type of `bwrap(1)` from `bin_t` to `container_runtime_exec_t`. With those revisions, the `bwrap(1)` processes run with SELinux type `container_runtime_t` and, thus, have the various SELinux permissions that they depend on. So, this MR - being unnecessarily sweeping in its security implications - should be reverted.

RoyalOughtness:
@Mark-4158 Why would we want bwrap to run as container_runtime_t globally? This PR affects bwrap confinement only when it's invoked by tumblerd. Making a broad change to bwrap's default context just to resolve this narrow gap is a recipe for breakage. Not to mention the fact that the container_runtime domain was to my knowledge developed without bubblewrap in mind, so simply assigning bwrap to that domain because it happens to have the right permissions in this case would also be in error. 

Also, the statement about making container-selinux a dependency of bubblewrap doesn't follow. bubblewrap's selinux policy would be handled by the selinux-policy package managed in this repo, which already ingests a container-selinux tarball in the spec.

PR #2988, opened 2025-12-09
Allow systemd-homework to remove ~/.identity-blob

janvhs:
This PR addresses the following AVC denial:
`type=AVC msg=audit(1763025211.718:1480): avc:  denied  { rmdir } for  pid=13456 comm="systemd-homewor" name=".identity-blob" dev="dm-1" ino=266 scontext=system_u:system_r:systemd_homework_t:s0 tcontext=unconfined_u:object_r:systemd_homed_record_t:s0 tclass=dir permissive=0`

I have this policy change laying around for some time, but I sadly didn't find the time to set up a rawhide VM for testing yet. However, this should fix the AVC denial I experienced.

### DIFF
--- a/policy/modules/system/systemd-homed.te
+++ b/policy/modules/system/systemd-homed.te
@@ -189,6 +189,7 @@ files_create_home_dir(systemd_homework_t)
 files_delete_home_dir(systemd_homework_t)
 files_search_home(systemd_homework_t)
 files_home_filetrans(systemd_homework_t, systemd_homed_crypto_luks_t, file)
+delete_files_pattern(systemd_homework_t, systemd_homed_record_t, systemd_homed_record_t)
 
 # unlabeled home directories
 files_manage_isid_type_dirs(systemd_homework_t)


policy/modules/system/systemd-homed.te
```diff
@@ -189,6 +189,7 @@ files_create_home_dir(systemd_homework_t)
 files_delete_home_dir(systemd_homework_t)
 files_search_home(systemd_homework_t)
 files_home_filetrans(systemd_homework_t, systemd_homed_crypto_luks_t, file)
+delete_files_pattern(systemd_homework_t, systemd_homed_record_t, systemd_homed_record_t)
```
zpytela, line 192: Can you move this line down to the block with similar types?

janvhs: You mean under the comment `# /run/systemd/user-home-mount` underneath the line stating `delete_files_pattern(systemd_homework_t, systemd_homed_runtime_work_dir_t, systemd_homed_record_t)`?

richiedaze: @zpytela it request deleting files in the .identity-blobs directory in the home directory. Thus, should belong in the /home section of systemd-homework's policy.

zpytela: I am not a regular homed user, so I tend to group rules based on the types. If you insist, it can be left this way.

janvhs: I think it makes sense where it is atm, because it's explicitly about paths present in `${HOME}` and the other rules are already grouped by path

zpytela:
Merging, thank you.

PR #2992, opened 2025-12-10
systemd: allow getattrs on fixed devices

xnox:
systemd-networkd is trying to monitor hot-plug of network devices, which seems to getattr on the block devices.

I don't know if there are better ways around this:
- make networkd not getattr on fixed disks
- use dont audit rule
- or if some other macro should have covered this but didn't

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -674,6 +674,9 @@ kernel_rw_net_sysctls(systemd_networkd_t)
 kernel_read_xen_state(systemd_networkd_t)
 kernel_read_network_state(systemd_networkd_t)
 kernel_rw_psi(systemd_networkd_t)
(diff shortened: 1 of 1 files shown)


xnox:
Not sure why revdeps failed, some of it looks like transient errors. Will try to repush this PR to retrigger tests.

zpytela:
@xnox At which moment the denial appears? Is it reproducible?

xnox:
> @xnox At which moment the denial appears? Is it reproducible?

every boot, of every VM with virtio networking and vda block device.

zpytela:
Can you show me your setup for a reproducer, or obtain traceback according to this?
https://fedoraproject.org/wiki/SELinux/Debugging#Using_perf_to_trace_all_system_denials

PR #2928, opened 2025-11-05
Add support for socket activated polkit helper agent

rfrohl:
The next polkit version is moving from setuid polkit-helper-agent to a socket activated version.

The socket is placed in a separate location under /run/polkit/.

polkit change:

https://github.com/polkit-org/polkit/commit/c007940054392b67b84b6044dda8a215040d8eb5

PR #2993, opened 2025-12-10
Question/RFC: pid 1 with no_new_privs implementation

ca-hu:
Context: There is some push to disable setuid binaries by starting pid 1 (systemd) with no_new_privs, which will lead subsequent processes to also need no_new_privs and so on.  I am not sure if you have something like this on the fedora end as well.

https://www.thkukuk.de/blog/no_new_privs/
https://lore.kernel.org/selinux/CAEjxPJ4eE_gau==Fj6fytyQb=jciQo+x391iDUh4EbJUMfrGKw@mail.gmail.com/T/#t

Question:
So the problem is for me, that I have found no clean and simple way to implement support for this in SELinux.

Ideas that I had and why it did not implement them:
- trace all `process transition`s from `init_t` and its children and add `nnp_transition` to them: would be likely really, really big patch. I am not sure, would you be interested in taking something like that?

The current implementation:
- adding it to all domtrans that use the `domain_transition_pattern` and manually add it to all that don't use the pattern. this is broader, but a smaller patch
- it is hidden behind a compile time flag, because it is not possible to have a `gen_tunable` inside of `domain_transition_pattern` as the pattern can be used inside of other `gen_tunable`, which will generate nested if(), which does not work with the macro processor. I tried to rewrite the `gen_tunable` as well, but then I keep hitting issues with `gen_bool`

I do not like this implementation myself, so my question is, do you by chance have a better idea? And would you want to have this in the fedora policy or should we better just keep this downstream until you have the need?

### DIFF
--- a/policy/modules/contrib/cron.if
+++ b/policy/modules/contrib/cron.if
@@ -101,6 +101,9 @@ interface(`cron_role',`
 
     tunable_policy(`cron_userdomain_transition',`
         allow crond_t $2_t:process transition;
+	ifdef(`init_server_no_new_privs',`
+	        allow crond_t $2_t:process2 nnp_transition;
+	')
         allow crond_t $2_t:fd use;
         allow crond_t $2_t:key manage_key_perms;
 
@@ -113,6 +116,9 @@ interface(`cron_role',`
         ps_process_pattern($2_t, cronjob_t)
     ',`
         dontaudit crond_t $2_t:process transition;
+	ifdef(`init_server_no_new_privs',`
+	        dontaudit crond_t $2_t:process2 nnp_transition;
+	')
         dontaudit crond_t $2_t:fd use;
         dontaudit crond_t $2_t:key manage_key_perms;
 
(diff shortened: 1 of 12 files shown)


ca-hu:
Reproducer can be found here: https://www.thkukuk.de/blog/no_new_privs/#opensuse-microos-and-opensuse-tumbleweed

zpytela:
I think I need more time to go through this.
A link to implementation in polkit or other tools would be nice.

ca-hu:
Link to polkit change: https://github.com/polkit-org/polkit/commit/c007940054392b67b84b6044dda8a215040d8eb5

I think @rfrohl has already created the PR for that change in the policy: https://github.com/fedora-selinux/selinux-policy/pull/2928


But this PR here is more part of this general poc described here: https://www.thkukuk.de/blog/no_new_privs/

The change that will make all processes spawned by systemd no_new_privs is this one:
https://build.opensuse.org/projects/home:kukuk:no_new_privs/packages/enable-no_new_privs/files/10-enable-NoNewPrivs.conf?expand=1

PR #3014, opened 2026-01-07
allow abrt_dump_oops to write to init sockets

JJanasek:
The abrt-dump-journal-core utility requires write
access to the NamespaceResource socket located
at /run/systemd/userdb/io.systemd.NamespaceResource

resolves: bz2283201


zpytela:
Are you sure this permission is sufficient?

Please rebase the commit now anyway, one wrong commit has been reverted.

zpytela:
Merging, thank you.

PR #3027, opened 2026-01-14
Allow kernel_t to read/write all domains' pipes

WOnder93:
We already allow it to manage all files_type pipes via files_manage_all_files(kernel_t), but pipes can sometimes be labeled with types from the domain attribute, so allow R/W access to those as well.

Resolves: RHEL-124442


zpytela:
Merging, thank you.

PR #3030, opened 2026-01-22
Allow wireguard to setup DNS using dns_hatchet

rfrohl:
(successor of an earlier [draft PR](https://github.com/fedora-selinux/selinux-policy/pull/2840))

The change allows the user to use the DNS config line with wireguard, example config and  usage for context/reproduction:

```sh
$ cat /etc/wireguard/wg0.conf 
[Interface]
PrivateKey = [cut]
Address = 240.0.0.1/24
DNS = 8.8.8.8
$ systemctl start wg-quick@wg0
[..]
$ systemctl stop wg-quick@wg0
[..]
```

The last line in the config file (`DNS =`) would cause problems without the changes in this PR.

What made this change complicated for me was the workaround about the `chcon` [call](https://git.zx2c4.com/wireguard-tools/tree/contrib/dns-hatchet/hatchet.bash#n27) in `wg-quick`. Already [discussed](https://github.com/fedora-selinux/selinux-policy/pull/2840#issuecomment-3224512990) before as well.

To summarize: I work around chcon with a couple of new dontaudit interfaces and use a filetrans for the proper context on `resolv.conf`.

### DIFF
--- a/policy/modules/contrib/wireguard.te
+++ b/policy/modules/contrib/wireguard.te
@@ -41,6 +41,32 @@ domain_use_interactive_fds(wireguard_t)
 
 files_read_etc_files(wireguard_t)
 
+# DNS hatchet part
+allow wireguard_t self:capability sys_admin;
+
+sysnet_create_config(wireguard_t)
+sysnet_mount_file(wireguard_t)
+sysnet_write_config(wireguard_t)
+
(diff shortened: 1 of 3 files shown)


rfrohl:
@zpytela  is there something I can do to help with move this change forward ?

rfrohl:
I only realized yesterday that Fedora removed the dns_hatchet with 33. Therefor we will carry a version of this change in the openSUSE downstream policy.

PR #3046, opened 2026-01-30
Allow lttng-sessiond to use sd_notify

mjeanson:
The next version of lttng-tools will use the SD_NOTIFY protocol to signal readiness of the daemon to systemd. This is currently blocked by the policy:

----
time->Fri Jan 30 16:16:23 2026
type=AVC msg=audit(1769807783.178:120): avc:  denied  { create } for  pid=1330 comm="lttng-sessiond" scontext=system_u:system_r:lttng_sessiond_t:s0 tcontext=system_u:system_r:lttng_sessiond_t:s0 tclass=unix_dgram_socket permissive=1 ----
time->Fri Jan 30 16:16:23 2026
type=AVC msg=audit(1769807783.178:121): avc:  denied  { sendto } for  pid=1330 comm="lttng-sessiond" path="/systemd/notify" scontext=system_u:system_r:lttng_sessiond_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=unix_dgram_socket permissive=1

### DIFF
--- a/policy/modules/contrib/lttng-tools.te
+++ b/policy/modules/contrib/lttng-tools.te
@@ -28,7 +28,15 @@ allow lttng_sessiond_t self:capability2 block_suspend;
 allow lttng_sessiond_t self:process { setrlimit signal_perms };
 allow lttng_sessiond_t self:fifo_file rw_fifo_file_perms;
 allow lttng_sessiond_t self:tcp_socket listen;
+allow lttng_sessiond_t self:unix_dgram_socket create;
 allow lttng_sessiond_t self:unix_stream_socket { create_stream_socket_perms connectto };
+
+# FIXME: this is required because of systemd's notify socket is created while
+# in the initramfs, hence as kernel_t. Once SELinux permits relabeling socket
+# objects created before the policy is loaded, this should be removed and
+# systemd fixed to relabel the socket appropriately.
+# Tracked by [systemd PR](https://github.com/systemd/systemd/pull/31336).
+kernel_dgram_send(lttng_sessiond_t)
 
 manage_dirs_pattern(lttng_sessiond_t, lttng_sessiond_var_run_t, lttng_sessiond_var_run_t)
 manage_files_pattern(lttng_sessiond_t, lttng_sessiond_var_run_t, lttng_sessiond_var_run_t)


zpytela:
Thanks for the proactive approach, 2 questions here:
1. In the commit message, only create permission is listed for lttng and unix_dgram_socket. Are there any others audited later?
2. The init_use_notify() expects init running in init_t domain, in the AVC there is kernel_t though which means kernel_dgram_send() should be used instead. Is this expected?

mjeanson:
I completely fumbled the testing of the updated policy, I was still in permissive mode and my modified policy wasn't loaded like I thought it was, sorry I'm not very familiar with SELinux. I'll update this PR once I have a better test setup.

> Thanks for the proactive approach, 2 questions here:
> 
> 1. In the commit message, only create permission is listed for lttng and unix_dgram_socket. Are there any others audited later?
> 2. The init_use_notify() expects init running in init_t domain, in the AVC there is kernel_t though which means kernel_dgram_send() should be used instead. Is this expected?

I tried replacing this with: 

```
allow lttng_sessiond_t kernel_t:unix_dgram_socket sendto;
```

Like it's done in `policy/modules/system/systemd.te` but make fails with:

```
/usr/bin/checkmodule -M -m tmp/lttng-tools.tmp -o tmp/lttng-tools.mod
policy/modules/contrib/lttng-tools.te:39:ERROR 'unknown type kernel_t' at token ';' on line 3003:
allow lttng_sessiond_t kernel_t:unix_dgram_socket sendto;
# Tracked by [systemd PR](https://github.com/systemd/systemd/pull/31336).
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
```

I suppose theres a kind of include path directive somewhere but I haven't figured it out yet.

zpytela:
Using types from other modules is not expected, that's what interfaces like kernel_dgram_send() are for. You can give "audit2allow -R" a go, while not always the best match, at least it discloses the right module name.

I am not saying though there was something bad, just asking for justification which I can't see.

mjeanson:
Ok, I so replaced the direct reference to `kernel_t` with:

```
kernel_dgram_send(lttng_sessiond_t)
```

Just to satisfy my curiosity, similarly it would be cleaner to use the following in `systemd.te`?

```
kernel_dgram_send(systemd_notify_t)
```

Also are there guidelines in incrementing the version in `policy_module()` when doing changes?

zpytela:
> Ok, I so replaced the direct reference to `kernel_t` with:
> 
> ```
> kernel_dgram_send(lttng_sessiond_t)
> ```
> 
> Just to satisfy my curiosity, similarly it would be cleaner to use the following in `systemd.te`?
> 
> ```
> kernel_dgram_send(systemd_notify_t)
> ```

Absolutely. Imagine I approved that PR without noticing, thank you for spotting this.

> 
> Also are there guidelines in incrementing the version in `policy_module()` when doing changes?

We don't use this field.

zpytela:
Merging, thank you.

PR #3047, opened 2026-01-31
fc for systemd-import

marcelvdh:
### Assigns systemd_importd_exec_t to systemd-import

The command `importctl --class machine import-raw fedora-43.raw fedora` generates the follow AVC errors:

```
type=AVC msg=audit(1769882462.092:1408): avc:  denied  { execute } for  pid=292101 comm="(sd-transfer)" name="systemd-import" dev="dm-0" ino=399506 scontext=system_u:system_r:systemd_importd_t:s0 tcontext=system_u:object_r:init_exec_t:s0 tclass=file permissive=1
----
type=AVC msg=audit(1769882462.092:1409): avc:  denied  { read open } for  pid=292101 comm="(sd-transfer)" path="/usr/lib/systemd/systemd-import" dev="dm-0" ino=399506 scontext=system_u:system_r:systemd_importd_t:s0 tcontext=system_u:object_r:init_exec_t:s0 tclass=file permissive=1
----
type=AVC msg=audit(1769882462.093:1410): avc:  denied  { execute_no_trans } for  pid=292101 comm="(sd-transfer)" path="/usr/lib/systemd/systemd-import" dev="dm-0" ino=399506 scontext=system_u:system_r:systemd_importd_t:s0 tcontext=system_u:object_r:init_exec_t:s0 tclass=file permissive=1
----
type=AVC msg=audit(1769882462.093:1411): avc:  denied  { map } for  pid=292101 comm="systemd-import" path="/usr/lib/systemd/systemd-import" dev="dm-0" ino=399506 scontext=system_u:system_r:systemd_importd_t:s0 tcontext=system_u:object_r:init_exec_t:s0 tclass=file permissive=1


```

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -26,7 +26,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 
 /usr/lib/systemd/systemd-initctl	--	gen_context(system_u:object_r:systemd_initctl_exec_t,s0)
 
-/usr/lib/systemd/systemd-import     --	gen_context(system_u:object_r:systemd_importd_exec_t,s0)
+/usr/lib/systemd/systemd-import		--	gen_context(system_u:object_r:systemd_importd_exec_t,s0)
 /usr/lib/systemd/systemd-pull		--	gen_context(system_u:object_r:systemd_importd_exec_t,s0)
 /usr/lib/systemd/systemd-userdbd	--	gen_context(system_u:object_r:systemd_userdbd_exec_t,s0)
 /usr/lib/systemd/systemd-userwork	--	gen_context(system_u:object_r:systemd_userdbd_exec_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
I am not sure this is the right approach. For start, any service which needs to execute /usr/lib/systemd/systemd-import would have to get this access. Also, unlike systemd-importd, this is not a system service, so it should have its own policy.

marcelvdh:
> I am not sure this is the right approach. For start, any service which needs to execute /usr/lib/systemd/systemd-import would have to get this access. Also, unlike systemd-importd, this is not a system service, so it should have its own policy.

Thank you for coming back @zpytela.

`systemd-import` and `systemd-pull` are both part of systemd (systemd-container-258.3-2.fc43) and are executed by  [systemd-importd](https://www.freedesktop.org/software/systemd/man/latest/systemd-importd.service.html) only AFAIK. I believe they are both helper applications. 

Currently `systemd-pull` is executed under the same context or is that one incorrect then too?

https://github.com/fedora-selinux/selinux-policy/blob/f905f0b133eba177e7f435ee94dcb767dad8d412/policy/modules/system/systemd.fc#L29

This one in the existing systemd policy rules is used by the `importctl pull-tar` the other other one (that I added) is executed by `importctl import-tar`. They are both are executed by the same daemon I believe.

PR #3054, opened 2026-02-05
Allow lttng tracing in default configuration

mjeanson:
In its default configuration trace data is written to the user home directory by lttng-sessiond in `~/lttng-traces/` and session configuration can be saved to `~/.lttng/`. Add the rules to allow the session daemon to manage these directories and their content in the users home directory.

Global configuration can be stored in `/etc/lttng/`, add rules to allow the session daemon to read it.

Also add the missing permissions for the session daemon to exec the consumer daemon when tracing starts.

This makes it possible to do userspace tracing which is blocked by the current policy.

### DIFF
--- a/policy/modules/contrib/lttng-tools.fc
+++ b/policy/modules/contrib/lttng-tools.fc
@@ -1,4 +1,12 @@
+HOME_DIR/\.lttng(/.*)?		gen_context(system_u:object_r:lttng_home_t,s0)
+HOME_DIR/lttng-traces(/.*)?	gen_context(system_u:object_r:lttng_home_t,s0)
+/root/\.lttng(/.*)?		gen_context(system_u:object_r:lttng_home_t,s0)
+/root/lttng-traces(/.*)?	gen_context(system_u:object_r:lttng_home_t,s0)
+
+/etc/lttng/(/.*)?		gen_context(system_u:object_r:lttng_conf_t,s0)
+
 /usr/bin/lttng-sessiond		--	gen_context(system_u:object_r:lttng_sessiond_exec_t,s0)
+/usr/lib/lttng/libexec/lttng-consumerd	--	gen_context(system_u:object_r:lttng_sessiond_exec_t,s0)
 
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/lttng-tools.fc
```diff
@@ -1,4 +1,12 @@
+HOME_DIR/\.lttng(/.*)?		gen_context(system_u:object_r:lttng_home_t,s0)
+HOME_DIR/lttng-traces(/.*)?	gen_context(system_u:object_r:lttng_home_t,s0)
+/root/\.lttng(/.*)?		gen_context(system_u:object_r:lttng_home_t,s0)
+/root/lttng-traces(/.*)?	gen_context(system_u:object_r:lttng_home_t,s0)
+
+/etc/lttng/(/.*)?		gen_context(system_u:object_r:lttng_conf_t,s0)
```
zpytela, line 6: Note labeling conf files is usually not needed unless services need to write to them, but that's ok.

policy/modules/contrib/lttng-tools.te
```diff
@@ -18,18 +18,26 @@ files_pid_file(lttng_sessiond_var_run_t)
...
+type lttng_conf_t;
+files_config_file(lttng_conf_t)
+
 ########################################
 #
 # lttng_sessiond local policy
 #
 
-allow lttng_sessiond_t self:capability { chown setgid setuid fsetid net_admin sys_resource };
+allow lttng_sessiond_t self:capability { chown dac_read_search setgid setuid fsetid net_admin sys_resource };
```
zpytela: Why is this capability needed?

zpytela (commented):
LGTM otherwise

policy/modules/contrib/lttng-tools.te
mjeanson: I had to go back and redo some testing, this capability only shows up in audit when the `lttng-traces` folder was created manually an thus not properly tagged.

I'll remove it and update this PR.

zpytela:
Merging, thank you.

PR #3055, opened 2026-02-05
Allow lttng kernel tracing

mjeanson:
The session daemon can be configured to load the lttng kernel modules by either calling the kmod utilities or by using the libkmod bindings. Add the required rules that were missing for the latter scenario.

The session daemon communicates with the kernel modules by writing to "/proc/lttng", add a rule to allow that.


zpytela:
Merging, thank you.

PR #3062, opened 2026-02-11
Grant systemd_generic_generator_t sys_resource capability (bsc#1257754)

rfrohl:
My first exposure to these generators. Found by automatic internal testing on ppc64le.

Fixes the following AVC:
```
audit(..): avc:  denied  { sys_resource } for  pid=532 comm="ln" capability=24  scontext=system_u:system_r:systemd_generic_generator_t:s0 tcontext=system_u:system_r:systemd_generic_generator_t:s0 tclass=capability permissive=1
```


zpytela:
Which generator it is? The generic label is intended for unknown or non-standard ones.

Is there a justification for sys_resource or can the system be possibly overloaded?

rfrohl:
thanks for the feedback!

Turns out this was a garbage change and there are 5 new generators that I missed. Will work on a proper separate fix.

PR #3065, opened 2026-02-12
Allow mdadm to use CAP_BPF during RAID monitoring

VeronikaSyncakova:
mdadm_t should be permitted to use CAP_BPF if required for legitimate RAID monitoring operations

Fixes: https://issues.redhat.com/browse/RHEL-135764


zpytela:
The fix LGTM, but can you include some information into the commit message? Anything to support the change - it can be the reproducing scenario, AVC denial, etc. Just a few lines.

zpytela:
Merging, thank you.

PR #3070, opened 2026-02-16
TLP Fixes (bsc#1257527) 

ca-hu:
see https://bugzilla.suse.com/show_bug.cgi?id=1257527

TLP uses `ps aux | grep ...` to check for different running services such as tuned-ppd or power-profiles-daemon and sets different profiles depending on it.

see for example:
https://github.com/linrunner/TLP/blob/700c4caf8b8139e292e9ef44335a897587a2c035/tlp-func-base.in#L470

So dontaudit generally and allow only tuned, ppd and unconfined_service_t (for tlp-pd, which not confined ATM in openSUSE)
Also dontaudit ps permissions that tlp_t does not need

FYI, in openSUSE the tlp package has a `Conflicts:` with power-profiles-daemon and tuned, so this scenario where both are running is very unlikely (so the `ps aux | grep` check is likely always false). In fedora rawhide it seems only tuned and ppd are conflicting in the package spec file, but when I tested installng `tlp` and then `power-profiles-daemon`, it gives conflicting files, so that combination is also not installable.

### DIFF
--- a/policy/modules/contrib/powerprofiles.if
+++ b/policy/modules/contrib/powerprofiles.if
@@ -1,1 +1,19 @@
 ## <summary>Power profiles handling over D-Bus</summary>
+
+########################################
+## <summary>
+##	Allow the domain to read powerprofiles state files in /proc.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`powerprofiles_read_state',`
+	gen_require(`
+		type powerprofiles_t;
+	')
+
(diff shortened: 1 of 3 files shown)


zpytela:
LGTM

Do you happen to have a test or configuration example for tlp-pd?

ca-hu:
not a good one, i just started tlp-pd, it was in unconfined_service_t by default. I checked that in the policy and it is not confined yet, so I modified the systemd service of tlp to run ps to see if the tlp-pd service shows up with the new policy

zpytela:
Merging then as is, thank you.

PR #3073, opened 2026-02-17
Allow mdadm use modprobe (bsc#1257793)

ca-hu:
mdadm needs permissions to execute modprobe:
https://github.com/md-raid-utilities/mdadm/blob/43d5b7866827ccb645d5eaa92aeba96fa85571c6/util.c#L2615

Leads to a failed boot otherwise:
https://bugzilla.suse.com/show_bug.cgi?id=1257793

Fixes:
Feb 02 08:05:31 workstation kernel: audit: type=1400 audit(1770015931.612:4): avc:  denied  { getattr } for  pid=1161 comm="sh" path="/usr/bin/kmod" dev="dm-0" ino=1739916 scontext=system_u:system_r:mdadm_t:s0-s0:c0.c1023 tcontext=system_u:object_r:kmod_exec_t:s0 tclass=file permissive=0

### DIFF
--- a/policy/modules/contrib/raid.te
+++ b/policy/modules/contrib/raid.te
@@ -121,6 +121,8 @@ fs_search_all(mdadm_t)
 mls_file_read_all_levels(mdadm_t)
 mls_file_write_all_levels(mdadm_t)
 
+modutils_domtrans_kmod(mdadm_t)
+
 storage_dev_filetrans_fixed_disk(mdadm_t)
 storage_manage_fixed_disk(mdadm_t)
 storage_read_scsi_generic(mdadm_t)


policy/modules/contrib/raid.te
```diff
@@ -40,7 +40,7 @@ logging_log_file(mdadm_log_t)
 # Local policy
 #
 
-allow mdadm_t self:capability { dac_read_search dac_override ipc_lock sys_admin sys_ptrace };
+allow mdadm_t self:capability { dac_read_search dac_override ipc_lock sys_admin sys_ptrace sys_module };
```
zpytela: This should not be needed because of
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/kernel.te#L583

Or is there any other reason for allowing explicitly?

policy/modules/contrib/raid.te
```diff
@@ -121,6 +125,10 @@ fs_search_all(mdadm_t)
 mls_file_read_all_levels(mdadm_t)
 mls_file_write_all_levels(mdadm_t)
 
+modutils_exec_kmod(mdadm_t)
```
zpytela: I wonder why modutils_domtrans_kmod() was not used: was it a deliberate decision?

policy/modules/contrib/raid.te
ca-hu: ah nice catch, thanks!

policy/modules/contrib/raid.te
ca-hu: I decided to use exec instead of domtrans because we don't need all the permissions of kmod_t and adding the permissions this way I can see which are used by mdadm_t. But I am not opposed to make this a domain transition, would you prefer that?

zpytela: I probably don't quite understand. If a transition applies on modprobe execution, modprobe gets all permissions it needs. It's not related to mdadm which executed it. One of the effects also is that mdadm_t does not need additional permissions.

In general, transition usually is the first choice unless there is e.g. some output which needs to be dealt with. Certainly there are special cases, I'd just like to know if this is one.

ca-hu: okay thanks for the feedback, changed it

zpytela:
Does this solution works for you now?

ca-hu:
> Does this solution works for you now?

yes, I recreated the reproducer yesterday and after installing the policy it works.


reproducer in case you need it:
create VM with tumbleweed, and add two storage devices vdb and vdc
```
mdadm --create --verbose /dev/md0 --level=0 --raid-devices=2 /dev/vdb /dev/vdc
mkfs.xfs /dev/md0
mount /dev/md0 /mnt
rsync -a /home/ /mnt/
umount /mnt
umount /home
echo "UUID=$(blkid -s UUID -o value /dev/md0) /home xfs defaults 0 0" >> /etc/fstab
mdadm --detail --scan >> /etc/mdadm.conf
```
delete the other "/home" lines that are not the new one in /etc/fstab if available and reboot

without the policy the boot will hang and when you check the journal for avcs, you can see the kmod ones.
if you boot again in permissive or selinux off, and install the policy with the change it will boot up fine.

lsblk should show you:
```
vdb    254:16   0    5G  0 disk  
└─md0    9:0    0   10G  0 raid0 /home
vdc    254:32   0    5G  0 disk  
└─md0    9:0    0   10G  0 raid0 /home
```

zpytela:
Merging then, thank you.

mwilck:
Late to the party here, but let me remark what I said on the SUSE bug already:

mdadm only needs to execute modprobe to make sure that the md module (`md_mod`) is loaded. This can be easily achieved using an entry in /etc/modules-load.d, without relaxing the current policy in any way. That might be the cleaner solution.

PR #3083, opened 2026-02-27
New systemd generators

rfrohl:
From NetworkManager
- nm-initrd-generator.sh 

systemd:
- systemd-run-generator
- systemd-system-update-generator

systemd-udev
- systemd-hibernate-resume-generator
- systemd-integritysetup-generator
- systemd-veritysetup-generator


rfrohl:
I only care about the `nm-initrd-generator.sh`. 

When debugging the AVC, the test framework made it hard to extract what the problematic generator is, so I added the types for all generic generators on the host. Thought I would include the systemd generators as well in the PR, if it makes sense to have them in the policy. 

Let me know if I should remove them instead.

zpytela:
Generally LGTM, but I wonder if you have any tests/usage given how little rules are present

rfrohl:
Only the `nm-initrd-generator.sh` was used, the others types are just for easier identification what is causing the issue.

If that is not enough then I am fine to remove these as well.

PR #3085, opened 2026-03-01
Allow httpds to connect to fcgiwrap

bkgood:
Prior to f31, nginx (and other processes running with httpd_t) were able to connect to fcgiwrap's FastCGI UNIX socket, with fcgiwrap running with unconfined_service_t, with policy:

    $ sudo sesearch -A -s httpd_t -t unconfined_service_t -c unix_stream_socket
    allow domain unconfined_service_t:unix_stream_socket { connectto getattr };

This policy no longer exists in f31, I presume via 936db96bc, and the connection is denied:

    type=AVC msg=audit(1771851421.922:190): avc:  denied  { connectto } for  pid=938 comm="nginx" path="/run/fcgiwrap/fcgiwrap-nginx.sock" scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=unix_stream_socket permissive=0

Applying httpd_exec_t to fcgiwrap, similar to php-fpm (another FastCGI application that nginx and other covered httpds connect to), reenables this connection.

Discussion at https://bugzilla.redhat.com/show_bug.cgi?id=2441936

Do I need to file a separate bug here or one for selinux-policy on Bugzilla?

I intend to backport to c{9,10}s if this is merged.

### DIFF
--- a/policy/modules/contrib/apache.fc
+++ b/policy/modules/contrib/apache.fc
@@ -73,6 +73,7 @@ HOME_DIR/((www)|(web)|(public_html))(/.*)?/logs(/.*)?	gen_context(system_u:objec
 /usr/bin/rotatelogs		--	gen_context(system_u:object_r:httpd_rotatelogs_exec_t,s0)
 /usr/bin/suexec		--	gen_context(system_u:object_r:httpd_suexec_exec_t,s0)
 /usr/bin/thttpd        -- gen_context(system_u:object_r:httpd_exec_t,s0)
+/usr/bin/fcgiwrap		--	gen_context(system_u:object_r:httpd_exec_t,s0)
 
 ifdef(`distro_suse', `
 /usr/bin/httpd2-.*		--	gen_context(system_u:object_r:httpd_exec_t,s0)


bkgood:
Build failures seem to be due to ongoing copr issues.

bkgood:
Hi @zpytela , would you mind having a look at this? I filed bugs and waited a while but haven't heard anything and I don't know if I failed to do something or this just fell off the radar.

zpytela:
I am fine with this solution from technical perspective. However, I am not aware of all use cases of using fcgiwrap so there could be a regression. Can you point me to some usage description?

Definitely a no-go for centos 9.

bkgood:
The use cases are pretty broad -- roughly "I have an old CGI script, and I want to service it with a web server that doesn't support CGI but does support FastCGI". With an old CGI script, the web server would traditionally fork and exec your script, communicate with it over stdin/out/err, and run with your web server's user/group/etc unless otherwise specified. FastCGI pipes all the stdin/out/err contents over sockets and is left to do any fork/exec'ing or thread spawns it needs to do, the web server is relieved of this.

Nginx and Caddy (in fedora repos) both support FastCGI but not CGI (the other packages in f44 providing webserver, lighthttp and apache, still support cgi). For these, fcgiwrap listens on a socket and spawns whatever old CGI script it is told to, and proxies the IO between the server and the script. If the web server can't connect to fcgiwrap's socket, fcgiwrap simply can't work. So I can't imagine how we could regress in terms of fcgiwrap use cases from the current situation.

> Definitely a no-go for centos 9.

Understood! I wasn't sure what could or could not be done at this point so I was just trying to spread my fix as wide as I could. My usecase is on co10 so if I can get it there, I'd be happy. I'm just using fedora to test.

PR #3087, opened 2026-03-03
Allow fs_write_tmpfs_files for systemd_sleep_t

aaronp24:
The NVIDIA driver [uses `filp_open()`](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/590.48.01/kernel-open/nvidia/os-interface.c#L1840) to create an anonymous temporary file during system suspend, which it uses to temporarily store the contents of GPU video memory when video memory preservation is enabled. This anonymous file is created in the filesystem backing `/tmp` by default, but the path can be configured with the
[`NVreg_TemporaryFilePath=` kernel module parameter](https://download.nvidia.com/XFree86/Linux-x86_64/590.48.01/README/powermanagement.html). A future version of the driver is planned to switch to using an anonymous shmem (tmpfs) file by default instead.

Currently, the code to do this is triggered by a write to `/proc/driver/nvidia/suspend`, which comes from the `nvidia-sleep.sh` script that is run by the `nvidia-suspend.service` systemd unit. This service runs in `unconfined_service_t`.

A future version of the NVIDIA driver will instead run this code in response to `systemd-sleep`'s write to `/sys/power/state`, which runs in the very restrictive `systemd_sleep_t` context, which blocks this filesystem access, causing suspend to fail.

Fix this problem by allowing tmpfs access in `systemd_sleep_t`.

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1909,6 +1909,7 @@ dev_rw_sysfs(systemd_sleep_t)
 dev_write_kmsg(systemd_sleep_t)
 
 fs_manage_efivarfs_files(systemd_sleep_t)
+fs_rw_tmpfs_files(systemd_sleep_t)
 
 fstools_rw_swap_files(systemd_sleep_t)
 


zpytela:
I wonder why a tmpfs is used when it is for a file in /tmp?

Another question: The file is created from kernel space and later used from userspace? It should anyway rather have a transition to a private type.

aaronp24:
Sorry, I guess the change description wasn't clear enough. The _current_ driver behavior is to use /tmp by default. A _future_ driver is planned to switch to using `shmem_file_setup()` by default which creates an anonymous file without going through the VFS layer. I can probably post patches to demonstrate that if it will help even though the driver with that change won't be out for a while.

The file is only ever used from kernel space. The driver just needs a place to temporarily write video memory data that will be lost when power to the GPU's VRAM is cut, so it can read it back after resume.

leigh123linux:
> file is created in the filesystem backing `/tmp` by default, but the path can be configured with the [`NVreg_TemporaryFilePath=` kernel module parameter](https://download.nvidia.com/XFree86/Linux-x86_64/590.48.01/README/powermanagement.html).


Fedora packaging uses /var/tmp


https://github.com/negativo17/nvidia-kmod-common/blob/master/nvidia.conf#L24

https://github.com/rpmfusion/nvidia-kmod/blob/master/set_driver_defaults.patch


selinux is still an issue

https://github.com/negativo17/nvidia-driver/issues/196

https://bugzilla.rpmfusion.org/show_bug.cgi?id=7428

aaronp24:
Thanks leigh123linux. This pull request targets what the default behavior of the driver will be once the [new driver behavior](https://github.com/NVIDIA/open-gpu-kernel-modules/compare/main...aaronp24:open-gpu-kernel-modules:suspend-resume-shmem-file) is implemented. If users or the distribution are changing `NVreg_TemporaryFilePath=`, then I think it makes sense to require updating the SELinux policy for that separately. Unless Fedora wants to allow all filesystem access to `systemd_sleep_t`, I think it makes sense to keep the policy targeted.

kwizart:
Using /var/lib/systemd/sleep instead of /var/tmp (or /tmp) avoid the need to setup a special fcontext

semanage fcontext -l |grep sleep
...
/var/lib/systemd/sleep(/.*)?                       all files          system_u:object_r:systemd_sleep_var_lib_t:s0


That said there is a need to actually create this directory (seems to work with root:root 0700 perms)
It can be done with tmpfiles.d/nvidia-sleep.conf (or by script)

I'm not sure how to fix the second AVC with perfmon, but sleep seems to work without it.

aaronp24:
What's still needed to make forward progress on this? This bug is blocking internal testing of the new kernel suspend notifier path by default.

kwizart:
@aaronp24 
Why would you need to allow systemd_sleep to use tmpfs when there is a dedicated file context for that ?

aaronp24:
Because `shmem_file_setup()` uses tmpfs and this path inherits the `systemd_sleep_t` context. The driver can't rely on users specifying a specific file path by default, the fact that it was using `/tmp` by default to being with wasn't ideal.

scaronni:
Any chance to see this merged?

aaronp24:
Is there anything still needed before this can be merged? As a quick reminder, this is going to impact the default configuration of the release 615 NVIDIA drivers when they come out, and I expect users to trip on this immediately if this isn't merged by then.

aaronp24:
I filed a bugzilla bug: https://bugzilla.redhat.com/show_bug.cgi?id=2512142

aaronp24:
I managed to work around this in the driver. Closing.

aaronp24:
The driver workaround didn't make it into the 615.71.09 driver release, so it's still exposed to the problem that this change tried to fix. You can work around it by doing any of the following:

- Patch the driver: https://forums.developer.nvidia.com/t/corrupted-video-memory-and-crashes-with-nvidia-open-driver-for-debian/379365/10
- Load nvidia.ko with `NVreg_UseKernelSuspendNotifiers=0` and use the older suspend method that avoids the problem
- Disable SELinux
- Apply this selinux-policy patch

leigh123linux:
@aaronp24  Can you reopen this so the lazy  fedora dev's can fix their mess!

I think taking more than 6 months for a dev response is taking the fucking piss!!

sunnyyangyangyang:
F44 report with this exact denial just filed: https://bugzilla.rpmfusion.org/show_bug.cgi?id=7543
Confirmed on 615.x open module (shmem tmpfs file is already the default, not future).
rw, not just write: write-only fixes suspend but resume read-back fails (black screen on wake).
This PR's fs_rw_tmpfs_files(systemd_sleep_t) is the minimal correct fix — could it be re-based and re-landed for rawhide?

kwizart:
Thanks for the clarifications about this issue. At rpmfusion we are now patching the driver and reverted to use /var/tmp as default NVreg_TemporaryFilePath (using /tmp with tmpfs still leads to a crash, but it's a separate issue unrelated to SELinux).

PR #3099, opened 2026-03-15
colord: ensure correct labeling for ICC profiles in user home

codingWiz-rick:
colord: ensure correct labeling for ICC profiles in user home

Previously, this used gnome_filetrans_home_content(colord_t) to
ensure correct labeling of ICC profiles created under
~/.local/share/icc/.

However, that approach granted overly broad home directory file
transition permissions.

Introduce a new, minimal interface
gnome_manage_home_icc_data_content() to allow managing
icc_data_home_t content without requiring file transitions.

Since ~/.local/share/icc is already labeled as icc_data_home_t,
new files inherit the correct type, and only manage permissions
are required.

Update colord_t to use this new interface, aligning with
least-privilege principles.

### DIFF
--- a/policy/modules/contrib/colord.te
+++ b/policy/modules/contrib/colord.te
@@ -144,7 +144,7 @@ optional_policy(`
 
 optional_policy(`
 	gnome_read_home_icc_data_content(colord_t)
-	gnome_filetrans_home_content(colord_t)
+	gnome_manage_home_icc_data_content(colord_t)
 	# Fixes lots of breakage in F16 on upgrade
 	gnome_read_generic_data_home_files(colord_t)
     gnome_map_generic_data_home_files(colord_t)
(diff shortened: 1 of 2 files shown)


zpytela:
Reading the description, it seems ~/.local/share already exists, so would it be sufficient if there was an interface just for icc_data_home_t?

Additionally, current state in the policy should not end up with unlabeled_t, is there another problem at the system?

codingWiz-rick:
Thanks for the clarification @zpytela — that helped narrow this down.

Upon recalling prior to the installation of fedora, I used ubuntu as my os I migrated to fedora.

The earlier `unlabeled_t` I observed was due to reusing a `/home` directory from a non-SELinux system, not a policy issue. After relabeling, files are correctly created as `icc_data_home_t` that 's my suspect now.

Given that `~/.local/share` already exists and labeling is correct, a file transition is unnecessary here. The real issue is that `colord_t` lacks the required permissions on `icc_data_home_t`.

I’ll rework the patch to drop the use of `gnome_filetrans_home_content()` and instead add a minimal, type-specific interface granting `colord_t` the required access to `icc_data_home_t`.

This should address the AVC denials without introducing broader home directory transitions.

I’ll send an updated patch shortly.

PR #3100, opened 2026-03-16
Allow create kerberos files in mysql db home

ndavidova:
This mirrors the logic applied to PostgreSQL in commit <b464336f5cefe0feddcc5b6787f4ab718a54bb86>.

Related: RHEL-119660

### DIFF
--- a/policy/modules/contrib/kerberos.if
+++ b/policy/modules/contrib/kerberos.if
@@ -636,6 +636,10 @@ interface(`kerberos_filetrans_named_content',`
 	kerberos_tmp_filetrans_host_rcache($1, "ldap_487")
 	kerberos_tmp_filetrans_host_rcache($1, "ldap_55")
 
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5identity")
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5login")
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5users")
+
 	postgresql_db_filetrans($1, krb5_home_t, file, ".k5identity")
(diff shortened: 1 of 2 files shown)


zpytela:
Interesting findings (like mistakes introduced 18 years ago, never noticed since then?).

Can you just expand a little bit on the commit messages/headers, at least the module name it fixes typos in?

ndavidova:
I squashed the 'fix typo' commit as it was just a typo I made in the first commit. The 'Fix names' makes sense to me to keep separate because I am also fixing a name in a summary I didn't add, but I updated the header, does this look better? @zpytela

policy/modules/contrib/kerberos.if
```diff
@@ -639,6 +639,10 @@ interface(`kerberos_filetrans_named_content',`
 	postgresql_db_filetrans($1, krb5_home_t, file, ".k5identity")
 	postgresql_db_filetrans($1, krb5_home_t, file, ".k5login")
 	postgresql_db_filetrans($1, krb5_home_t, file, ".k5users")
+
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5identity")
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5login")
+	mysql_db_filetrans($1, krb5_home_t, file, ".k5users")
```
zpytela, line 641: please move this block up to keep alphabetical order

zpytela (changes requested):
otherwise lgtm

zpytela:
Can you now squash the 2 related commits?

zpytela:
Merging, thank you.

PR #3111, opened 2026-03-20
Allow sys_resource on execution of generic executables conditionally

zpytela:
Allow the sys_resource capability to each domain which is allowed to execute generic programs in system bin directories without domain transition when the coreutils_single tunable is on.

coreutils-single is a collection of GNU core utilities, but packaged as a single multicall binary. Each program now executes prctl() with PR_SET_MM which requires the CAP_SYS_RESOURCE capability.

prctl(PR_SET_NAME, "mkdir")             = 0
prctl(PR_SET_MM, PR_SET_MM_ARG_START, 0x7fff20a92dc7, 0, 0) = 0

capabilities(7)
      CAP_SYS_RESOURCE
             •  employ the prctl(2) PR_SET_MM operation;

PR_SET_MM(2const)
DESCRIPTION
       Modify certain kernel memory map descriptor fields of the calling
process. Usually these fields are set by the kernel and dynamic loader (see ld.so(8) for more information) and a regular application should not use this feature. However, there are cases, such as self-modifying programs, where a program might find it useful to change its own memory map. The calling process must have the CAP_SYS_RESOURCE capability.

Resolves: RHEL-141858

### DIFF
--- a/policy/global_tunables
+++ b/policy/global_tunables
@@ -153,3 +153,10 @@ gen_tunable(use_virtualbox, false)
 ## </p>
 ## </desc>
 gen_tunable(deny_bluetooth,false)
+
+## <desc>
+## <p>
+## Allow the sys_resource capability to all domains allowed to execute bin_t
+## </p>
+## </desc>
+gen_tunable(corecmd_bin_sys_resource, false)
--- a/policy/modules/admin/netutils.te
+++ b/policy/modules/admin/netutils.te
@@ -254,7 +254,7 @@ corenet_sctp_bind_generic_node(traceroute_t)
 corenet_icmp_bind_generic_node(traceroute_t)
 corenet_rawip_bind_unreserved_port(traceroute_t)
 
-corecmd_exec_bin(traceroute_t)
+corecmd_exec_bin_wrapper(traceroute_t)
 
 dev_getattr_infiniband_dev(traceroute_t)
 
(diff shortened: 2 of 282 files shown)


WOnder93:
> coreutils-single is a collection of GNU core utilities, but packaged as a single multicall binary. Each program now executes prctl() with PR_SET_MM which requires the CAP_SYS_RESOURCE capability.

How does this work for unprivileged users? Is the `EPERM` error from `prctl(PR_SET_MM, ...)` non-fatal with some fallback? Could it be better to (conditionally) dontaudit the sys_resource permission instead?

zpytela:
Good point, Ondrej. Nothing else happens but argv[0] will not change, thus keep being intransparent https://github.com/coreutils/coreutils/blob/master/src/coreutils.c#L115

Although it's questionable if it should be allowed, I really tend to dontaudit it, at least for now. Another issue is that checkmodule does not allow for the current solution with a tunable.

zpytela:
checkmodule fails because corecmd_exec_bin already is in some tunable blocks (29) and it would be a major change to pull the interface calls outside the tunable to make it working

WOnder93:
You could probably work around that by introducing an attribute for the interface, applying the rules/interfaces directly to the attribute, and the interface would then only add individual types to the attribute.

zpytela:
I chose a bit different approach in the end. I am just a bit worried though if the increased number of rules will impact the policy size.

PR #3122, opened 2026-03-27
Allow virt_qemu_ga_run_unconfined control guest-exec (bsc#1237450)

ca-hu:
The qemu-ga guest-exec command allows a host to execute arbitrary commands on the guest.

Before this commit, when the guest-exec command was used with qemu-guest-agent, SELinux allowed the the host to execute processes that were generally allowed by the virt_qemu_ga_t domain and denied accessing processes that were not allowed.

However, this bad in security and usability as not all commands are denied and not all commands are allowed and there is no way to configure which ones are allowed, except by giving the main virt_qemu_ga_t domain more permissions in general.

This breaks use cases such as ansible with the qemu-guest-agent plugin.

Quick reproducer for the problem that this commit tries to solve: Using virsh on the host:
- list VM names and grab your guest where you want to execute commands on
- for example execute this with cloud-init installed on the guest:
```
$ qemu-agent-command <your-vm-name> '{"execute": "guest-exec", "arguments": { "path": "/usr/bin/cloud-init", "arg": [ "/" ], "capture-output": true }}'
error: guest agent command failed: unable to execute QEMU agent command 'guest-exec': Failed to execute child process “/usr/bin/cloud-init” (Permission denied)

$ qemu-agent-command <your-vm-name> '{"execute": "guest-exec", "arguments": { "path": "/usr/bin/ls", "arg": [ "/" ], "capture-output": true }}'
{"return":{"pid":109083}}
```

With this commit and a change in libvirt, this can be controlled with the virt_qemu_ga_run_unconfined boolean.

virt_qemu_ga_run_unconfined off -> both ls and cloud-init will fail with guest-exec virt_qemu_ga_run_unconfined on -> both ls and cloud-init will be executed with guest-exec

Requires a change with a selinux-helper in libvirt to be functional!

### DIFF
--- a/policy/modules/contrib/virt_supplementary.te
+++ b/policy/modules/contrib/virt_supplementary.te
@@ -73,6 +73,14 @@ files_tmp_file(virt_qemu_ga_tmp_t)
 type virt_qemu_ga_data_t;
 files_type(virt_qemu_ga_data_t)
 
+type virt_qemu_ga_selinux_helper_t;
+domain_type(virt_qemu_ga_selinux_helper_t)
+
+type virt_qemu_ga_selinux_helper_exec_t;
+application_executable_file(virt_qemu_ga_selinux_helper_exec_t)
+domain_entry_file(virt_qemu_ga_selinux_helper_t, virt_qemu_ga_selinux_helper_exec_t)
+role system_r types virt_qemu_ga_selinux_helper_t;
+
 type virt_qemu_ga_unconfined_exec_t;
 application_executable_file(virt_qemu_ga_unconfined_exec_t)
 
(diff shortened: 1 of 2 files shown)


policy/modules/contrib/virt_supplementary.te
```diff
@@ -243,11 +251,13 @@ tunable_policy(`virt_qemu_ga_read_nonsecurity_files',`
 ')
 
 tunable_policy(`virt_qemu_ga_run_unconfined',`
+    domtrans_pattern(virt_qemu_ga_t, virt_qemu_ga_selinux_helper_exec_t, virt_qemu_ga_selinux_helper_t)
     domtrans_pattern(virt_qemu_ga_t, virt_qemu_ga_unconfined_exec_t, virt_qemu_ga_unconfined_t)
```
berrange, line 255: IMHO this is a bad idea.  The "virt_qemu_ga_run_unconfined" rule is a toggle for the guest admin to allow qemu-ga to run scripts that the guest admin has defined in ``/etc/qemu-ga/fsfreeze-hook\.d``

This is overloading that tunable to provide a way for an external / remote user to leverage qemu-ga to execute arbitrary commands. The guest owner has no control over what is run this way. So this turns "virt_qemu_ga_run_unconfined" into a massive hole.

ca-hu:
https://mail.gnu.org/archive/html/qemu-devel/2026-03/msg07752.html

As upstream also discourages this, we will check to disable "exec" in openSUSE/SUSE and find a different solution as well. Closing for the time being.

PR #3130, opened 2026-04-01
Allow Polkit to get attributes of user terminals

HastD:
Polkit 127 supports a new authentication caching mechanism that allows processes to skip authentication if another process meeting certain conditions was recently authenticated in the same terminal. For reference, see: https://github.com/polkit-org/polkit/pull/533

For this to work, Polkit needs to be allowed to get attributes of user terminals. This is granted by the following rule:
`userdom_getattr_user_terminals(policykit_t)`

I tested this on a freshly created and updated VM running Fedora 44 Workstation, and the addition of this rule does allow Polkit authentication caching to work. (You can test this by running `run0 true` or `pkexec true` twice in a row in the same terminal.)


zpytela:
Merging, thank you.

PR #3146, opened 2026-04-13
WIP: systemd_run policy

JJanasek:
Introduce the module to systemd_run and extend policy so run0 sessions can run common commands (dnf, shell, systemctl, password prompts) under systemd_run_t without relying on broad unconfined_t entrypoint rules.

### DIFF
--- a/policy/modules/admin/systemd_run.fc
+++ b/policy/modules/admin/systemd_run.fc
@@ -1,7 +1,8 @@
 
-/usr/bin/systemd-run(edit)?	--	gen_context(system_u:object_r:systemd-run_exec_t,s0)
+/usr/bin/systemd-run(edit)?	--	gen_context(system_u:object_r:systemd_run_exec_t,s0)
+/usr/bin/run0(edit)?		--	gen_context(system_u:object_r:systemd_run_exec_t,s0)
 
-/var/db/systemd-run(/.*)?		gen_context(system_u:object_r:systemd-run_db_t,s0)
+/var/db/systemd-run(/.*)?		gen_context(system_u:object_r:systemd_run_db_t,s0)
 
-/var/log/systemd-run-io(/.*)?		gen_context(system_u:object_r:systemd-run_log_t,s0)
-/var/log/systemd-run\.log	--	gen_context(system_u:object_r:systemd-run_log_t,s0)
+/var/log/systemd-run-io(/.*)?		gen_context(system_u:object_r:systemd_run_log_t,s0)
+/var/log/systemd-run\.log	--	gen_context(system_u:object_r:systemd_run_log_t,s0)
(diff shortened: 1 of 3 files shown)


zpytela:
I had to add some rules:
auth_domtrans_chkpwd() policykit_domtrans_auth() systemd_passwd_agent_exec()
policykit_dbus_chat() init_status() init_start()

probably just because of systemd and other updates in the meantime. Using systemd_passwd_agent_domtrans() would be better not to add additional rules, but it would need some more tweaking.

Confined users will also need init_ interfaces or some smart transition.

Otherwise good for simple use cases. Have you tried some more complex? It will also require some assessment which transitions to allow, e.g. passwd does not work now.

JJanasek:
Thanks for the review. You are right regarding the missing interfaces and the systemd-tty-ask-password-agent transition.

To answer your questions regarding more complex use cases and user profiles, I built a dedicated automated test. I explicitly tested both unconfined and confined caller sessions, ensuring that confined-user paths were fully exercised in addition to the unconfined ones:

Unconfined admin session (unconfined_t)
Confined staff session (staff_u:staff_r:staff_t)
Confined standard user session (user_u:user_r:user_t)

>Confined users will also need init_ interfaces or some smart transition.

Agreed. To get baseline Polkit/systemd IPC working for those confined sessions in the test setup, I had to add the extra rules you mentioned (auth_domtrans_chkpwd(), policykit_domtrans_auth(), policykit_dbus_chat(), init_status(), init_start()). I also agree that systemd_passwd_agent_domtrans() is the cleaner long-term direction, but it definitely needs additional tuning.

>Otherwise good for simple use cases. Have you tried some more complex? It will also require some assessment which transitions to allow, e.g. passwd does not work now.

Yes, the functional behavior for more complex sandboxing works well, with the exception of interactive TTY prompts. Here is what worked in my tests:

Explicit context overrides: systemd-run -p SELinuxContext=... correctly forced transition into a custom sandbox domain.
Context overrides + UID drops: -p User=nobody -p SELinuxContext=... correctly applied both UID switch and SELinux transition.
Package-manager child workloads: passing a dnf -y update payload through run0 executed the child process in the rpm_t domain (verified via ps aZxl snapshots). This indicates run0 inherits expected reference-policy transition behavior.

The limitation: as you noted with passwd, the interactive password flow is currently blocked (run0 passwd -> systemd-tty-ask-password-agent: Permission denied).

so maybe adding systemd_passwd_agent_domtrans(systemd_run_t) could be a potential fix.
but then i get `avc:  denied  { execute } for  pid=20666 comm="(passwd)" name="passwd" dev="nvme0n1p2" ino=139418 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:passwd_exec_t:s0 tclass=file permissive=0`

PR #3151, opened 2026-04-13
Label /usr/lib/systemd/systemd-ssh-proxy with systemd_ssh_proxy_exec_t

zpytela:
Make systemd_ssh_proxy_exec_t a part of the exec_type attribute.


zpytela:
Options: just exec_type, a transition, use bin_t.

PR #3154, opened 2026-04-15
Allow systemd_homework_t to delete systemd_homed_record_t dirs

jsegitz:
When using commands like
homectl update foobar --member-of systemd-journal
and also in some part of the error handling of systemd_homeworkd this is necessary.

Solves:
avc:  denied  { rmdir } for
comm="systemd-homewor" name=".identity-blob"
scontext=system_u:system_r:systemd_homework_t:s0
tcontext=system_u:object_r:systemd_homed_record_t:s0 tclass=dir


zpytela:
Merging, thank you.

PR #3155, opened 2026-04-16
config: make /etc/systemd/user same as /usr/lib/systemd/user

bboozzoo:
/etc/systemd/user is mislabeled as etc_t, while the user unit under /usr/lib/systemd/user and HOME_DIR/.config/systemd/user are correctly identified as systemd_unit_file_t.


bboozzoo:
@zpytela gentle ping. We're observing selinux logs related to this in our CI on both Fedora and openSUSE TW.

zpytela:
Merging, thank you.

PR #3160, opened 2026-04-20
Dontaudit read access of procmail to postfix_etc_t (bsc#1261933)

ca-hu:
Procmail does not need the access and it occurs because postfix has a file descriptor leak.

Fixes:

type=AVC msg=audit(1775896746.188:188): avc:  denied  { read } for  pid=11835 comm="procmail" path="/etc/postfix.lmdbtest/relay.lmdb" dev="nvme0n1p2" ino=9306487 scontext=system_u:system_r:procmail_t:s0 tcontext=unconfined_u:object_r:postfix_etc_t:s0 tclass=file permissive=0

### DIFF
--- a/policy/modules/contrib/postfix.if
+++ b/policy/modules/contrib/postfix.if
@@ -140,6 +140,25 @@ interface(`postfix_read_config',`
 
 ########################################
 ## <summary>
+##	Dontaudit read to postfix configuration files.
+## </summary>
+## <param name="domain">
+##	<summary>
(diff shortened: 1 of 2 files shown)


zpytela:
Have you reported this for the postfix component?

We try to avoid dontaudit rules in the policy since they make troubleshooting harder.

ca-hu:
postfix maintainer fixed this in postfix, so closing

PR #3169, opened 2026-04-24
Networkmanager: Allow NM to manage files under /run

### LINKED ISSUE #3123: Allow `NetworkManager_dispatcher_console_t` to write in `/run/issue.d`
I am working on updating `console-login-helper-messages` to write generated issues to `/run/issues.d` instead of `/etc`. This make more sense as those are generated content. They were wrote to `/etc` because historically that's where `agetty` was sourcing the issues.

Now agetty learned to source `/run/issues.d` as well, so it make sense to use it.
In `clhm` we ship a [NetworkManager dispatcher script](https://github.com/coreos/console-login-helper-messages/blob/main/etc/NetworkManager/dispatcher.d/90-console-login-helper-messages-gensnippet_if) that write an issue with the IP address of all the interfaces. With this change it fails because selinux prevents it to write to `/run/issue.d` 

I believe this is because `NetworkManager_dispatcher_console_t` is confined to manging `/etc/`:

https://github.com/fedora-selinux/selinux-policy/blob/f37f1665a3cc800bbd7d409327727530f4fe7965/policy/modules/contrib/networkmanager.te#L619 

Here is related the AVC denial : 
```
type=AVC msg=audit(1774515716.955:634): avc:  denied  { add_name } for  pid=4135 comm="mv" name="22_clhm_ens4.issue" scontext=system_u:system_r:NetworkManager_dispatcher_console_t:s0 tcontext=system_u:object_r:systemd_ssh_issue_var_run_t:s0 tclass=dir permissive=1
```

The operation the dispatcher script tries to do is creating a temporary file in `/run/console-login-helper-message` then `mv` it to `/run/issue.d/`. 

That's where it happens : 
https://github.com/coreos/console-login-helper-messages/blob/4c66bbb9c69731d87f56db77fb6cc796df68570d/usr/lib/console-login-helper-messages/libutil.sh#L39-L40

jbtrystram:
In console-login-helper-messages we moved the generated issues from `/etc` to `/run` in v0.22.0.

Allow `systemd_ssh_issue_manage_pid_files(NetworkManager_dispatcher_console_t)` through a new interface `systemd_ssh_issue_manage_pid_files`.

Fixes https://github.com/fedora-selinux/selinux-policy/issues/3123

### DIFF
--- a/policy/modules/contrib/networkmanager.te
+++ b/policy/modules/contrib/networkmanager.te
@@ -711,6 +711,7 @@ optional_policy(`
 	systemd_status_systemd_services(NetworkManager_dispatcher_sendmail_t)
 	systemd_start_systemd_services(NetworkManager_dispatcher_nvme_t)
 	systemd_status_systemd_services(NetworkManager_dispatcher_nvme_t)
+	systemd_ssh_issue_manage_pid_files(NetworkManager_dispatcher_console_t)
 ')
 
 optional_policy(`
--- a/policy/modules/contrib/networkmanager.te
+++ b/policy/modules/contrib/networkmanager.te
@@ -616,8 +616,6 @@ dev_rw_wireless(NetworkManager_dispatcher_tlp_t)
 
 domain_read_all_domains_state(NetworkManager_dispatcher_dnssec_t)
 
-files_manage_etc_files(NetworkManager_dispatcher_console_t)
-
 fs_getattr_pidfs(NetworkManager_dispatcher_iscsid_t)
 fs_getattr_pidfs(NetworkManager_dispatcher_sendmail_t)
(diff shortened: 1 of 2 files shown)


travier:
Suggestion in https://github.com/coreos/console-login-helper-messages/issues/135#issuecomment-4345717872

travier:
The naming feels weird. Is it because it follows convention from similar functions?

jbtrystram:
> The naming feels weird. Is it because it follows convention from similar functions?

Agreed. That's what I was suggested by one of the maintainers. Also yeah, I think the previous function was named similarly

jbtrystram:
@zpytela can you take a look at this please ? This have been blocking us to do any release of CLHM in fedora for a while

zpytela:
Merging, sorry for the delay.

Issue #2960, opened 2025-11-24
Files in /var/opt no longer have expected var_t type


CendioOssman:
There was a recent change in #2917 that broke things for us. All of a sudden, things in `/var` changed from being the expected `var_t` type, to the classically read-only `usr_t` type. This naturally broke our policy and started breaking services.

Could this please be reverted, and something more specific being added for whatever needs image mode needs?

CendioOssman:
@zpytela

samuellauren:
This change affects Linux products developed by WithSecure Corporation as well. Our convention places immutable files under `/opt/PRODUCT` and mutable files under `/var/opt/PRODUCT`, with separate SELinux rules for each. The updated policy treats both paths identically, which removes our ability to differentiate them or assign distinct rules.

Could you share the motivation for this change, and clarify how we should structure our installations going forward? Is the intent for everything to be placed under `/opt/`?

link22Hoffstedt:
The change in #2917 affects Linux products developed by link22 AB as well. Our products store immutable files under `/opt/<company>/<product>` and mutable files under `/var/opt/<company>/<product>` as the commenters above. SELinux policies for the products set a custom security context type for `/var/opt/<company>/<product>` and implements rules for the custom type. 

After #2917 our SELinux policies fail silently at changing security contexts for `/var/opt/<company>/<product>`.

Can [@zpytela](https://github.com/zpytela) explain if this is intended behavior?

remicollet:
Affects all "Software Collections" which also use `/opt/vendor/project` and  `/var/opt/vendor/project` and `/etc/opt/vendor/project` (see #3213)

PR #3183, opened 2026-05-05
Revert "Define file equivalency for /var/opt"

madnicendio:
This reverts commit c852a6898d55bf49d53461cce74d51c578310642.

The alias /var/opt -> /opt breaks third-party software that uses the /var/opt namespace for writable data. See issue #2960

### DIFF
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -36,4 +36,3 @@
 /bin                 /usr/bin
 /usr/etc             /etc
 /usr/sbin            /usr/bin
-/var/opt             /opt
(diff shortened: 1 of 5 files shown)


madnicendio:
The issue was also noted in [ostreedev/ostree#1589 ](https://github.com/ostreedev/ostree/issues/1589).

madnicendio:
@zpytela

CendioOssman:
This bug has apparently been pushed to RHEL now and is now breaking production systems. Can we get a fix here soon, @zpytela?

zpytela:
This issue needs to be resolved together with the original requestee.

CendioOssman:
Is that publicly available somewhere? There were no references in the commit or PR.

PR #3188, opened 2026-05-11
Allow sys_resource on execution of generic executables conditionally

zpytela:
Allow the sys_resource capability to each domain which is allowed to execute generic programs in system bin directories without domain transition when the coreutils_bin_sys_resource tunable is on. If the tunable is off, which is the default state, the capability is dontaudited instead. The allow or dontaudit rule applies to the sys_resource_type attribute. Unability to set the name of the calling thread or kernel memory map descriptor fields is not critical.

coreutils-single is a collection of GNU core utilities, but packaged as a single multicall binary. Each program now executes prctl() with PR_SET_MM which requires the CAP_SYS_RESOURCE capability.

prctl(PR_SET_NAME, "mkdir")             = 0
prctl(PR_SET_MM, PR_SET_MM_ARG_START, 0x7fff20a92dc7, 0, 0) = 0

capabilities(7)
      CAP_SYS_RESOURCE
             •  employ the prctl(2) PR_SET_MM operation;

PR_SET_MM(2const)
DESCRIPTION
       Modify certain kernel memory map descriptor fields of the calling
process. Usually these fields are set by the kernel and dynamic loader (see ld.so(8) for more information) and a regular application should not use this feature. However, there are cases, such as self-modifying programs, where a program might find it useful to change its own memory map. The calling process must have the CAP_SYS_RESOURCE capability.

Resolves: RHEL-141858

### DIFF
--- a/policy/global_tunables
+++ b/policy/global_tunables
@@ -153,3 +153,10 @@ gen_tunable(use_virtualbox, false)
 ## </p>
 ## </desc>
 gen_tunable(deny_bluetooth,false)
+
+## <desc>
+## <p>
+## Allow the sys_resource capability to all domains allowed to execute bin_t
+## </p>
+## </desc>
+gen_tunable(corecmd_bin_sys_resource, false)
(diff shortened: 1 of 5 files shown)


zpytela:
Conflicts: https://github.com/fedora-selinux/selinux-policy/pull/3111

PR #3191, opened 2026-05-12
Allow virtnwfilterd_t r/w on packet_socket (bsc#1264273)

jsegitz:
Already has permissions for other types and it's needed for e.g. dhcp network filter.
Solves:
avc:  denied  { read write } for comm="dhcp-snoop" path="socket:[21972]" dev="sockfs" scontext=system_u:system_r:virtnwfilterd_t:s0
tcontext=system_u:system_r:virtnwfilterd_t:s0
tclass=packet_socket permissive=0


zpytela:
Merging, thank you.

PR #3194, opened 2026-05-15
Allow the kernel to execute also special files

WOnder93:
Update the files_mmap_exec_all_files() to allow also execute on special files with any file label, because such access vectors can be reached by calling access(2) on such files vi NFS (they happen on the server side as a side effect of checking the DAC permissions against overriden creds).

Note that execute on some special files can also be triggered via access(2) on a regular filesystem against the current task, so the NFS checks aren't entirely nonsensical.

Resolves: https://redhat.atlassian.net/browse/RHEL-72728


zpytela:
Merging, thank you.

PR #3210, opened 2026-06-03
Remove extra parameters from interface headers

vmojzis:
Also add missing ### separators so that the test script can properly separate individual interfaces.

There are no policy changes (just comments), so this should be safe to merge.

Resolves: RHEL-129394


vmojzis:
TBH I'm not sure what does `<rolecap/>` do, so no idea if it should be removed as well (the test does not complain about it though)

zpytela:
Merging, thank you.

PR #3211, opened 2026-06-03
packit: Stop notifying martinpitt for Cockpit test failures

jelly:
Martin left Red Hat and no is no longer involved in Cockpit development.


zpytela:
Regret to hear that, but thank you for the commit.

PR #3218, opened 2026-06-04
Add missing interface requirements

vmojzis:
Found and fixed mostly by claude code.

Note that aside from adding missing attributes and types in the gen_require there some different changes found by checking the interface requires:

policy/modules/admin/sudo.if - the type associated with the role is passed as the 3rd argument and should not be "created" as  $1_t (even though this does not cause any policy changes since the macro is consistently used with $3 = $1_t)

policy/modules/contrib/kismet.if - kistmet_tmpfs_t and kistmet_t are just typos (not found so far because the interface was never used)

policy/modules/contrib/portage.if - I added a new gen_require section for the optional_policy block. This should be handled by an interface, but the type is not defined in the policy any more (should we just remove this alltogether?)

Resolves: RHEL-135488

### DIFF
--- a/policy/modules/admin/sudo.if
+++ b/policy/modules/admin/sudo.if
@@ -59,7 +59,7 @@ template(`sudo_role_template',`
 	allow $1_sudo_t $3:file read_file_perms;;
 	allow $1_sudo_t $3:key search;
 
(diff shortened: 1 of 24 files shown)


zpytela:
Merging, thank you.

PR #3219, opened 2026-06-04
Add missing interface requirements

vmojzis:
Found and fixed mostly by claude code (but thoroughly reviewed by me).

Note that aside from adding missing attributes and types in the gen_require there some different changes found by checking the interface requires:

policy/modules/admin/sudo.if - the type associated with the role is passed as the 3rd argument and should not be "created" as  $1_t (even though this does not cause any policy changes since the macro is consistently used with $3 = $1_t)

policy/modules/contrib/kismet.if - kistmet_tmpfs_t and kistmet_t are just typos (not found so far because the interface was never used)

policy/modules/contrib/portage.if - I added a new gen_require section for the optional_policy block. This should be handled by an interface, but the type is not defined in the policy any more (should we just remove this alltogether?)

### DIFF
--- a/policy/modules/contrib/condor.if
+++ b/policy/modules/contrib/condor.if
@@ -84,7 +84,7 @@ interface(`condor_domtrans_master',`
 #
 interface(`condor_startd_ranged_domtrans_to',`
     gen_require(`
-        type sshd_t;
+        type condor_startd_t;
     ')
     condor_startd_domtrans_to($1, $2)
 
--- a/policy/modules/contrib/sssd.if
+++ b/policy/modules/contrib/sssd.if
@@ -486,7 +486,7 @@ interface(`sssd_run_stream_connect',`
 #
 interface(`sssd_dontaudit_run_stream_connect',`
 	gen_require(`
-		type sssd_t, sssd_var_lib_t;
+		type sssd_t, sssd_var_run_t;
 	')
 
 	dontaudit $1 sssd_t:unix_stream_socket connectto;
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -6328,7 +6328,7 @@ interface(`files_filetrans_system_conf_named_files',`
 #
 interface(`files_relabelto_system_conf_files',`
     gen_require(`
-        type usr_t;
+        type system_conf_t;
     ')
 
     relabelto_files_pattern($1, system_conf_t, system_conf_t)
@@ -6346,7 +6346,7 @@ interface(`files_relabelto_system_conf_files',`
 #
 interface(`files_relabelfrom_system_conf_files',`
     gen_require(`
-        type usr_t;
+        type system_conf_t;
     ')
 
     relabelfrom_files_pattern($1, system_conf_t, system_conf_t)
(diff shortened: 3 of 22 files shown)


policy/modules/contrib/condor.if
```diff
@@ -84,7 +84,7 @@ interface(`condor_domtrans_master',`
 #
 interface(`condor_startd_ranged_domtrans_to',`
     gen_require(`
-        type sshd_t;
+        type sshd_t, condor_startd_t;
```
zpytela: sshd_t seems actually not to be needed here

policy/modules/contrib/portage.if
```diff
@@ -197,6 +197,10 @@ interface(`portage_compile_domain',`
 	ifdef(`TODO',`
 	# some gui ebuilds want to interact with X server, like xawtv
 	optional_policy(`
+		gen_require(`
```
zpytela: This would be a candidate for a new interface, but the type seems not to be defined anywhere. Anyway out of scope of this commit.

policy/modules/contrib/sssd.if
```diff
@@ -486,7 +486,7 @@ interface(`sssd_run_stream_connect',`
 #
 interface(`sssd_dontaudit_run_stream_connect',`
 	gen_require(`
-		type sssd_t, sssd_var_lib_t;
+		type sssd_t, sssd_var_lib_t, sssd_var_run_t;
```
zpytela: sssd_var_lib_t is not needed

policy/modules/kernel/files.if
```diff
@@ -6328,7 +6328,7 @@ interface(`files_filetrans_system_conf_named_files',`
 #
 interface(`files_relabelto_system_conf_files',`
     gen_require(`
-        type usr_t;
+        type system_conf_t, usr_t;
```
zpytela: usr_t not needed

policy/modules/kernel/files.if
```diff
@@ -6346,7 +6346,7 @@ interface(`files_relabelto_system_conf_files',`
 #
 interface(`files_relabelfrom_system_conf_files',`
     gen_require(`
-        type usr_t;
+        type system_conf_t, usr_t;
```
zpytela: usr_t not needed

policy/modules/contrib/portage.if
vmojzis: So should I just ignore this issue? The gen_require seems more important here given that the type is not defined anywhere in the policy.

policy/modules/contrib/condor.if
vmojzis: Yep, claude says there is 4397 interfaces with a redundant type/attribute in the gen_require so I wanted to address that separately. But I guess it makes sense to fix the interfaces we're already touching properly.

policy/modules/contrib/portage.if
zpytela: yes, ignore atm

policy/modules/contrib/condor.if
zpytela: if it's tracked elsewhere, then np

vmojzis:
Fixed all the redundant requires you pointed out (+ claude found an additional one in logging_systemctl_syslogd) and split the change in portage.if into a separate PR (https://github.com/fedora-selinux/selinux-policy/pull/3226, but we may want to just ignore it given that the module is disabled).

zpytela:
Merging, thank you.

PR #3220, opened 2026-06-04
Remove extra parameters from interface headers

vmojzis:
Also add missing ### separators so that the test script can properly separate individual interfaces.

Resolves: RHEL-129394


zpytela:
Merging, thank you.

PR #3222, opened 2026-06-09
Dontaudit virt_driver_domain execmem

VeronikaSyncakova:
virtqemud, when the policy was split from virtd_t was granted ‘execmem’ via ‘virtqemud_use_execmem’, which is not necessary as well as it isn’t necessary for any of the virt daemons to have this capability.

The virtqemud_use_execmem tunable was already removed, but the dontaudit rule was not implemented.

Resolves: RHEL-44901


zpytela:
explicit dontaudit for virtstoraged_t can be removed

zpytela:
Merging, thank you.

PR #3226, opened 2026-06-11
portage_compile_domain: Require xdm_xserver_tmp_t type

vmojzis:
Add a new gen_require section for the optional_policy block. This should be handled by an interface, but the type is not defined in the policy any more (should we just remove this altogether). The interface is only used in the portage policy module, which is disabled.


zpytela:
LGTM

PR #3240, opened 2026-06-15
vfs_snapper support with samba_share_t (bsc#1265400) 

ca-hu:
Support snapper to create snapshots in samba shares with samba_share_t
label. When a user decides(*) to set up their samba share:
- with samba_share_t label (without the samba booleans)
- AND vfs_snapper
..snapperd_t can not create snapshots under it as it can not create files
with the samba_share_t label.

However, it needs to do it since those files need to be accessible
for samba:
https://www.samba.org/samba/docs/current/man-html/vfs_snapper.8.html

To allow that scenario, allow snapper to manage samba_share_t
directories and files to be able to create snapshots.

(*) In openSUSE update-samba-security-profile will automatically do
that since [bsc#1259050](https://bugzilla.suse.com/show_bug.cgi?id=1259050).

https://bugzilla.suse.com/show_bug.cgi?id=1265400

### DIFF
--- a/policy/modules/contrib/samba.if
+++ b/policy/modules/contrib/samba.if
@@ -458,6 +458,42 @@ interface(`samba_read_secrets',`
 
 ########################################
 ## <summary>
+##	Allow the specified domain to manage samba's share files
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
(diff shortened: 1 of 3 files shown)


JMarcosHP:
Hi, can this be merged?

This affects Fedora 44 and 45 too.

zpytela:
LGTM

PR #3244, opened 2026-06-17
systemd pcrlock policy

ca-hu:
Basically we use systemd pcrlock mostly for [sdbootutil](https://github.com/openSUSE/sdbootutil) for a while now, we just did not find the time to upstream the general parts. 

I cherry-picked the commits that relate to general pcrlock for this PR. If you would like the sdbootutil(+snapper) policy as well, let me know. I don't think fedora is using it ATM though and it is still in movement, so I did not include it.

I can also squash the commits if you like, just thought that you might want to check the history/development for review. 

https://www.freedesktop.org/software/systemd/man/latest/systemd-pcrlock.html

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -127,6 +127,8 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /var/lib/systemd/linger(/.*)?		gen_context(system_u:object_r:systemd_logind_var_lib_t,mls_systemhigh)
 /var/lib/systemd/sleep(/.*)?		gen_context(system_u:object_r:systemd_sleep_var_lib_t,s0)
 /var/lib/systemd/timesync(/.*)?		gen_context(system_u:object_r:systemd_timedated_var_lib_t,s0)
+/var/lib/systemd/pcrlock.*   --   gen_context(system_u:object_r:systemd_pcrlock_var_lib_t,s0)
+/var/lib/pcrlock\.d(/.*)?              gen_context(system_u:object_r:systemd_pcrlock_var_lib_t,s0)
 /var/lib/private/systemd/journal-upload(/.*)?		gen_context(system_u:object_r:systemd_journal_upload_var_lib_t,s0)
 /var/lib/private/systemd/timesync(/.*)?		gen_context(system_u:object_r:systemd_timedated_var_lib_t,s0)
 /usr/lib/systemd/resolv.*   --   gen_context(system_u:object_r:lib_t,s0)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -161,7 +161,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/systemd/machines(/.*)?	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/machines.lock	--	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/oom(/.*)?		gen_context(system_u:object_r:systemd_oomd_var_run_t,s0)
-/run/systemd/pcrlock.json	--	gen_context(system_u:object_r:systemd_pcrlock_var_lib_t,s0)
+/run/systemd/pcrlock.json	--	gen_context(system_u:object_r:systemd_pcrlock_var_run_t,s0)
 /run/systemd/resolve(/.*)?	gen_context(system_u:object_r:systemd_resolved_var_run_t,s0)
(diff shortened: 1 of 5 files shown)


policy/modules/system/systemd.te
```diff
@@ -332,6 +332,9 @@ files_pid_file(systemd_oomd_var_run_t)
 systemd_domain_template(systemd_pcrextend)
 systemd_domain_template(systemd_pcrlock)
 
+type systemd_pcrlock_var_lib_t;
+files_type(systemd_pcrlock_var_lib_t);
```
zpytela: It it correct that there are no fc entries?

policy/modules/system/systemd.if
```diff
@@ -978,6 +996,30 @@ interface(`systemd_passwd_agent_exec',`
...
+## </summary>
+## </param>
+#
+interface(`systemd_pcrlock_exec',`
+    gen_require(`
+        type systemd_pcrlock_exec_t, systemd_pcrlock_var_lib_t;
+    ')
+
+    can_exec($1,systemd_pcrlock_exec_t)
+    files_var_lib_filetrans($1, systemd_pcrlock_var_lib_t, { dir file }, "pcrlock.d")
```
zpytela: I don't quite understand why rules not related to exec are here. Usually it is sufficient to use domtrans to allow such rules.

It is not common, therefore not transparent when one looks for a problem. I'd rather use another interface to allow it unless there is only one use case.

policy/modules/system/systemd.if
```diff
@@ -2048,6 +2048,25 @@ interface(`systemd_filetrans_named_content',`
...
+##	<summary>
+##      Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_pcrlock_filetrans_named_content',`
+	gen_require(`
+		type systemd_pcrlock_var_lib_t;
+	')
+	# Needed when /var/lib/pcrlock.d gets unlinked and recreated
```
zpytela, line 2065: When it gets unlinked?

policy/modules/system/systemd.if
```diff
@@ -2048,6 +2048,25 @@ interface(`systemd_filetrans_named_content',`
...
+	files_var_lib_filetrans($1, systemd_pcrlock_var_lib_t, { dir file }, "pcrlock.d")
```
zpytela, line 2066: pcrlock.d can also be a file?

policy/modules/system/systemd.fc
```diff
@@ -161,6 +161,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/systemd/machines(/.*)?	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/machines.lock	--	gen_context(system_u:object_r:systemd_machined_var_run_t,s0)
 /run/systemd/oom(/.*)?		gen_context(system_u:object_r:systemd_oomd_var_run_t,s0)
+/run/systemd/pcrlock.json	--	gen_context(system_u:object_r:systemd_pcrlock_var_lib_t,s0)
```
zpytela, line 164: use \ to escape regex

policy/modules/system/systemd.fc
```diff
@@ -175,6 +175,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/varlink/registry/.+	-l	gen_context(system_u:object_r:systemd_varlink_registry_t,s0)
 
 /run/log/bootchart.*	--	gen_context(system_u:object_r:systemd_bootchart_var_run_t,s0)
+/run/log/systemd/tpm2-measure.log	--	gen_context(system_u:object_r:systemd_pcrlock_var_lib_t,s0)
```
zpytela, line 178: use \

policy/modules/system/systemd.if
```diff
@@ -1008,7 +1008,7 @@ interface(`systemd_passwd_agent_exec',`
 #
 interface(`systemd_pcrlock_exec',`
     gen_require(`
-        type systemd_pcrlock_exec_t, systemd_pcrlock_var_lib_t;
+        type systemd_pcrlock_exec_t;
     ')
 
     can_exec($1,systemd_pcrlock_exec_t)
```
zpytela, line 1014: separate with a space for legibility

policy/modules/system/systemd.te
```diff
@@ -2137,6 +2136,9 @@ systemd_manage_pcrlock_files(systemd_pcrlock_t)
 # still keep it permissive for now. Failure to run can prevent booting
 permissive systemd_pcrlock_t;
 
+systemd_pcrlock_manage_var_lib_files(systemd_pcrlock_t)
```
zpytela, line 2139: we usually don't use interfaces for own types unless it helps somehow (simplification, legibility, manageability when used also for other domains, etc.)

zpytela (changes requested):

PR #3248, opened 2026-06-18
Fix typo in the comment of build.conf

nyoxi:
Based on the documentation in `build.md` and the content of `Makefile` I am judging this is a typo and the comment refers to `MCS_CATS` and not to `MLS_CATS`.


zpytela:
Good catch, thank you.

PR #3270, opened 2026-07-03
Changes adapting to bind packages with suffixes

pemensik:
bind package has started providing binaries with suffixes. That includes main daemon named, rndc and other tools. Links to the package are provided via alternatives.

Add few bind9 directories to redhat packages. They are not yet provided, but will be soon. Also consider /etc/named/ directory and their contents equivalent to named.conf.

named-sdb and named-pkcs11 are supported with new suffixes too, although they are not compiled on any supported Fedora version anymore. First is replaced by possible loadable DLZ plugins, the second by pkcs11 support via openssl configuration.

Related: RHEL-191263


pemensik:
This should fix change in rawhide bind package after update https://bodhi.fedoraproject.org/updates/FEDORA-2026-822e2bfb73. Sadly I have not created separate bug on Fedora for it and referenced it from bind changes. Anyway, it needs modification of selinux-policy rules.

pemensik:
Related: https://redhat.atlassian.net/browse/RHEL-192093

zpytela:
@pemensik, thank you for the proactive approach.

In the variable part, will there always be just a number/alphanum?

pemensik:
not sure. There is also variant bind9-next in Fedora, where I thought about named-next suffix for example. But for current implementations,  using only version made it easier to various regexes or patterns. Primary place of conflict was named-checkconf vs named-(something) for daemon variants. But yes, I think alphanum with dots or dashes should be always used for any suffixes.

pemensik:
but there are couple of tools called ``named-*`` in the existing package. They do not need to be protected the same way named(-9.18) itself is. But I guess if we apply the same context as for named, it should not break anything. They are usually working just with local files and trasforming them one way or another.

zpytela:
Merging, thank you.

PR #3271, opened 2026-07-03
Allow systemd-coredump socket access in containers

VeronikaSyncakova:
Add file transitions for coredump socket creation in nspawn containers. /run/ in the running container is labeled with tmpfs_t (or user_tmpfs_t if started via systemd-nspawn directly), including the /run/systemd/coredump socket.

type=AVC msg=audit(1778852345.055:500): avc:  denied  { write } for  pid=8634 comm="(sd-coredump)" name="coredump" dev="tmpfs" ino=43 scontext=system_u:system_r:systemd_coredump_t:s0 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=sock_file permissive=0

Resolves: RHEL-139983


policy/modules/system/systemd.if
```diff
@@ -2743,6 +2743,24 @@ interface(`systemd_rw_coredump_tmpfs_files',`
...
+## <summary>
+##	Read and write to systemd-coredump temporary socket files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_coredump_tmpfs_sock_files',`
```
zpytela: the permission name is missing in the interface name, but I think the interface is rather not needed at all since it is not used outside this module

policy/modules/system/systemd.te
```diff
@@ -1395,6 +1397,15 @@ optional_policy(`
 	logging_send_syslog_msg(systemd_coredump_t)
 ')
 
+optional_policy(`
+    gen_require(`
+            type unconfined_t;
+    ')
+    filetrans_pattern(unconfined_t, user_tmp_t, systemd_coredump_tmpfs_t, sock_file, "coredump")
```
zpytela: types outside this module should not be used directly, in this case there are even 2 of them in each rule; it is usually resolved by a chain of interfaces or using an attribute

zpytela (commented):
I am sorry for the delay, it took me quite some time to make up my mind - I think the reported problem should actually be resolved in systemd, but this usually takes more time.

policy/modules/system/systemd.if
```diff
@@ -2743,6 +2743,37 @@ interface(`systemd_rw_coredump_tmpfs_files',`
...
+########################################
+## <summary>
+##	Read and write to systemd-coredump temporary socket files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
```
zpytela: This should not be here.

policy/modules/system/systemd.te
```diff
@@ -1359,9 +1361,14 @@ allow systemd_coredump_t self:process setcap;
 allow systemd_coredump_t self:unix_stream_socket connectto;
 allow systemd_coredump_t self:user_namespace create;
 
+allow systemd_coredump_t systemd_coredump_tmpfs_t:sock_file write;
+
 manage_files_pattern(systemd_coredump_t, systemd_coredump_tmpfs_t, systemd_coredump_tmpfs_t)
 fs_tmpfs_filetrans(systemd_coredump_t, systemd_coredump_tmpfs_t, file )
 
+fs_tmpfs_filetrans(systemd_coredump_socket_type, systemd_coredump_tmpfs_t, sock_file, "coredump")
+userdom_user_tmp_filetrans(systemd_coredump_socket_type, systemd_coredump_tmpfs_t, sock_file, "coredump")
```
zpytela, line 1370: The systemd_coredump_tmpfs_t type is common for tmpfs_t and user_tmp_t, was that a deliberate decision? Like not to deal with an overkill for such a small (and temporary) feature?

zpytela (changes requested):
otherwise lgtm

policy/modules/system/systemd.te
VeronikaSyncakova: I used systemd_coredump_tmpfs_t for both because it was already defined and both scenarios needed the same permissions, so it seemed simpler and less complex than creating a new  type.

policy/modules/system/unconfined.te
```diff
@@ -53,4 +53,6 @@ optional_policy(`
 
 optional_policy(`
 	systemd_coredump_socket_create(unconfined_service_t)
+')
+    container_runtime_nnp_domtrans(unconfined_service_t)
```
zpytela, line 57: Different optional block need to be used for interfaces from different modules (invalid syntax aside)

zpytela (changes requested):

PR #3278, opened 2026-07-07
Allow net_admin to the nfsd_t domain

scottmayhew:
nfs-utils 2.9.1 added a netlink interface to mountd/exportd to handle sunrpc cache upcalls from the kernel.  The netlink interface requires CAP_NET_ADMIN.

type=AVC msg=audit(1783461620.873:503): avc:  denied  { net_admin } for  pid=57981 comm="rpc.mountd" capability=12  scontext=system_u:system_r:nfsd_t:s0 tcontext=system_u:system_r:nfsd_t:s0 tclass=capability permissive=0

### DIFF
--- a/policy/modules/contrib/rpc.te
+++ b/policy/modules/contrib/rpc.te
@@ -237,7 +237,7 @@ optional_policy(`
 # NFSD local policy
 #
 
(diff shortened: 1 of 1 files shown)


pcmoore:
@pebenito probably need a similar fix in https://github.com/SELinuxProject/refpolicy.

zpytela:
Merging, thank you.

Issue #3281, opened 2026-07-12
Allow domain transition from init_t into postgresql_t with NoNewPrivileges set


varuniyer:
## Summary

The policy has no `nnp_transition` rule for `postgresql_t`. When postgresql.service runs with an explicit `NoNewPrivileges=yes`, the `init_t` to `postgresql_t` transition is denied. Because `init_t` may run direct init entrypoints with `execute_no_trans`, the server then executes in `init_t` instead of failing, and cannot read its own `postgresql_db_t` files, so startup fails on reading `postgresql.conf`.

This is the postgresql analog of #2663 (redis), and of the dirsrv request in rhbz#2457951.

## Environment

- Fedora Cloud 44 (GCE image fedora-cloud-44), x86_64
- selinux-policy-targeted-44.3-1.fc44
- Stock Fedora `postgresql-server`
- Also reproduced on a Fedora 44 bootc system (aarch64, selinux-policy-targeted-44.4-1.fc44) with a source-built PostgreSQL 17 labeled by the shipped contexts

## Reproduction

```
dnf install -y postgresql-server
postgresql-setup --initdb
mkdir -p /etc/systemd/system/postgresql.service.d
printf '[Service]\nNoNewPrivileges=yes\n' > /etc/systemd/system/postgresql.service.d/nnp.conf
systemctl daemon-reload
systemctl start postgresql
```

The unit fails. `journalctl -u postgresql` shows:

```
LOG:  could not open configuration file "/var/lib/pgsql/data/postgresql.conf": Permission denied
FATAL:  configuration file "/var/lib/pgsql/data/postgresql.conf" contains errors
```

`journalctl _TRANSPORT=audit` shows both denials:

```
AVC avc:  denied  { nnp_transition } for  pid=1205 comm="(postgres)" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:postgresql_t:s0 tclass=process2 permissive=0
AVC avc:  denied  { read } for  pid=1205 comm="postgres" name="postgresql.conf" dev="sda3" ino=2369 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:postgresql_db_t:s0 tclass=file permissive=0
```

Note the AVCs appear only in the journal, not in dmesg, which makes this hard to spot on headless systems that expose only the kernel console.

Removing the drop-in and restarting recovers, and postgres then runs confined:

```
ps -eo comm,label | grep postgres
[...]

Issue #2663, opened 2025-04-28
Allow domain transition from `init_t` into `redis_t` with `NoNewPrivileges` set


yhabteab:
We have modified the default Redis systemd unit file and utilised numerous systemd security options, including the [`NoNewPrivileges`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Security) feature. However, this particular option appears to result in an SELinux `nnp_transition` denial and consequently prevents the service from transitioning from the `init_t` to the `redis_t` domain. I searched for similar issues with other daemons in this repository and indeed found that some of the modules from `modules/contrib` are already permitted to transition, even with this systemd option set. Thus, I would like to ask whether the same can be done for the `redis_t` domain. Thanks!

```bash
[root@fd41 ~]# ausearch -m avc -ts recent
----
time->Mon Apr 28 07:09:53 2025
type=AVC msg=audit(1745824193.880:2207): avc:  denied  { nnp_transition } for  pid=10529 comm="(s-server)" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:redis_t:s0 tclass=process2 permissive=0
----
```

PR #3282, opened 2026-07-12
Allow init_t nnp domain transition to postgresql_t

varuniyer:
One-line fix for #3281, mirroring the redis fix from #2663.

Verified on Fedora Cloud 44: with this change compiled via selinux-policy-devel and loaded over the base module, postgresql.service starts under an explicit NoNewPrivileges=yes and runs confined as postgresql_t, where it previously fell back to init_t and failed reading its own configuration.

Drafted with AI assistance (Claude); all reproduction and verification steps were run and reviewed by the submitter.


zpytela:
Merging, thank you.

PR #3285, opened 2026-07-14
 Fix whitespace across all policy modules

vmojzis:
Use tabs for indentation, correct indentation depth based on M4
    block nesting, remove trailing whitespace and whitespace-only lines,
    fix ## doc comment indentation in .if files.
    
    Includes scripts/fix_whitespace.py which parses M4 block structure
    (backtick/singlequote quoting, interface/template/optional_policy/
    ifdef/gen_require/ifelse blocks, C-style if(){} blocks) and
    re-indents each line to the correct depth.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


policy/modules/contrib/virt.if
vmojzis, line 1: This might be wrong.

vmojzis: https://github.com/fedora-selinux/selinux-policy/pull/3285/changes#diff-877615f81a2a3e69f48c3c51b3e71c444d12c6c921b98c4f9cccfc1aff7130e1L1655

vmojzis: Removed.

vmojzis:
Rebased and tweaked the script to also locate midline double spaces (e.g. between two permissions). The tweak found an unexpected amount of midline double spaces, so I made a separate commit.

policy/modules/kernel/corenetwork.if.m4
```diff
@@ -22,16 +22,16 @@ define(`create_netif_interfaces',``
 ##	</summary>
 ## </param>
 ## <infoflow type="both" weight="10"/>
-#
-interface(`corenet_tcp_sendrecv_$1_if',`
-	gen_require(`
-		$3 $1_$2;
-	')
+		#
```
vmojzis: This is definitely not right.  The script got confused by the "``" on line 14.

vmojzis: Fixed by adding a special case for the "``". The change only impacts corenetwork.if.m4 (which now has a much smaller diff).

vmojzis:
@zpytela I reviewed the whole thing, added 2 minor tweaks to the script that fixed special cases (those are merged into the first commit), then added a new commit removing midline double spaces and the last commit deals with dead code (I guess that is the only one you should check).

policy/modules/kernel/corenetwork.te.m4
```diff
@@ -22,124 +22,124 @@ define(`range_start',`ifelse(-1,index(`$1', `-'),$1,substr($1,0,index(`$1', `-')
...
 define(`network_interface_controlled',`
-ifdef(`__network_enabled_declared__',`',`
+	ifdef(`__network_enabled_declared__',`',`
 ## <desc>
-## <p>
-## Enable network traffic on all controlled interfaces.
-## </p>
+##	<p>
+##	Enable network traffic on all controlled interfaces.
+##	</p>
```
vmojzis: This comment section should be indented.

vmojzis: https://github.com/fedora-selinux/selinux-policy/pull/3285/changes#diff-c2cf692136baeb0b2bdbd953f2abe36326e6ac5a45eb56fc2542aa63e8b63387R57

vmojzis: Fixed by the latest commit.

vmojzis:
Ok, all anomalies I found are resolved. The only commit that does not solely deal with whitespaces is https://github.com/fedora-selinux/selinux-policy/pull/3285/changes/8d807a504ca22033746eb136283a0726abc6b3a3 

Please let me know once you are happy with the changes and I'll squash all the changes into a single commit (except for the github action).

zpytela (approved):
I am fine with it as is. My only comment is about
```
-	#	refpolicywarn(`$0($*) has been deprecated, please use mls_process_read_all_levels() instead.')
```
which I think should not have been commented out in commit f8233ab7b01 which brought the interface name change. Anyway commit is from 2007 and I feel no need to change anything other than removing the deprecated interfaces.

vmojzis:
Obsoleted by https://github.com/fedora-selinux/selinux-policy/pull/3413. Leaving this open just for the un-squashed commits.

vmojzis:
> which I think should not have been commented out in commit [f8233ab](https://github.com/fedora-selinux/selinux-policy/commit/f8233ab7b0154f836ecc81367bf00e0ff976af65) which brought the interface name change. Anyway commit is from 2007 and I feel no need to change anything other than removing the deprecated interfaces.

Right, mls_process_read_up and mls_process_write_down are both in use even though technically obsoleted in https://github.com/fedora-selinux/selinux-policy/commit/2d0c9cecaf9ec6377b0a22633ae95f6b3b55542d

@zpytela  Do you mean that we should replace all uses of both interfaces with mls_proces_*_all_levels alternatives and remove the interface definitions? Or just replace the uses?

zpytela:
> > which I think should not have been commented out in commit [f8233ab](https://github.com/fedora-selinux/selinux-policy/commit/f8233ab7b0154f836ecc81367bf00e0ff976af65) which brought the interface name change. Anyway commit is from 2007 and I feel no need to change anything other than removing the deprecated interfaces.
> 
> Right, mls_process_read_up and mls_process_write_down are both in use even though technically obsoleted in [2d0c9ce](https://github.com/fedora-selinux/selinux-policy/commit/2d0c9cecaf9ec6377b0a22633ae95f6b3b55542d)
> 
> @zpytela Do you mean that we should replace all uses of both interfaces with mls_proces_*_all_levels alternatives and remove the interface definitions? Or just replace the uses?

Replace, in some other commit. There are alltogether 254 occurences including those from other modules. Remove all deprecated interfaces a year or so later.

vmojzis:
Closing as obsole (https://github.com/fedora-selinux/selinux-policy/pull/3413)

PR #3288, opened 2026-07-16
Allow all domains to use inherited sshd-session pipes

### LINKED ISSUE #3286: sshd_session_t pipes cannot be passed over D-Bus: systemd-run / systemctl --machine fail over SSH without a TTY (regression from the sshd-session split)
> [!NOTE]
> **Disclaimer:** I am not an SELinux policy expert — apologies in advance if terminology or conventions are off. This investigation was carried out with the help of an AI assistant (Claude — Fable 5), and every reproducer, AVC, and fix in this report and its companion PR was verified on a real Fedora 44 system (enforcing, re-checked with `semodule -DB`). I've done my best to make this contribution as valuable as possible — corrections are very welcome.

On Fedora 44 (stock targeted policy, enforcing), any `systemd-run` or `systemctl --machine=…` invocation started from an SSH session **without a TTY** — the default transport of Ansible, pyinfra, and config-management tools generally — fails with a misleading "Connection reset by peer".

The command's stdio fds are pipes created by `sshd-session`, labeled `sshd_session_t`, and `systemd-run --pipe` forwards them inside the `StartTransientUnit` D-Bus message. Neither receiving domain may use them:

- **hop 1, client → dbus-broker:** `system_dbusd_t` has no allow rule at all for `sshd_session_t:fifo_file`;
- **hop 2, dbus-broker → PID 1:** `init_t` has none either. This second denial only becomes visible once hop 1 is allowed — measured below.

The kernel silently strips LSM-denied `SCM_RIGHTS` fds and sets `MSG_CTRUNC`, which sd-bus ≤ v259 / dbus-broker treat as fatal for the connection — hence the bare connection reset with no client-visible SELinux error. systemd upstream reviewed exactly this fd-stripping mechanism and adjudicated it as kernel/LSM behavior (systemd/systemd#34688, closed at the merge of the v260 mitigation systemd/systemd#40089, which keeps connections alive but leaves the fds stripped), so the functional fix can only be in policy.

### This is a regression from the OpenSSH sshd-session split

Pre-split, the stdio pipes of a no-TTY SSH session were created by sshd in `sshd_t`, and policy carries a domain-wide grant for exactly that — `ssh_rw_pipes(domain)` in `policy/modules/kernel/domain.te`, which expands to:

```
allow domain sshd_t:fifo_file rw_inherited_fifo_file_perms;
# rw_inherited_fifo_file_perms = { getattr read write append ioctl lock }  (obj_perm_sets.spt)
```

The sshd-session split moved these pipes to `sshd_session_t`, and the grant was never mirrored: the only caller of `ssh_session_rw_pipes()` is `passwd_t` (`policy/modules/admin/usermanage.te`).

A sibling of this failure was reported in 2022 as #1106: the same fd-strip mechanism on `StartTransientUnit`, but with shell/systemctl-created fds — most likely `unconfined_t` pipes or socketpairs; the denied object was never captured — and it reproduced from local consoles too, so it was not SSH session stdio. The pipe side of that family was addressed by [rhbz#2036829](https://bugzilla.redhat.com/show_bug.cgi?id=2036829) (commit 95d703493, `userdom_rw_inherited_user_pipes(init_t)`; dbus-broker was already covered through `application_domain_type`), yet #1106 was never closed: its reporter could not capture AVCs, most likely because the `unconfined_t` fifo denials were dontaudit'd domain-wide at the time. The present report is the SSH-stdio member of the same family, at the `sshd_session_t` label the split introduced — and here the AVCs land in the journal out of the box, supplying the kind of evidence #1106 has been missing.

### Reproducer

```
sudo useradd -r -m immich
sudo loginctl enable-linger immich
```

(`immich` is an arbitrary service user; any user with a running user manager works.) Then, from any client, over SSH **without** a TTY:

```
$ ssh <host> 'sudo systemctl --user --machine=immich@.host daemon-reload; echo rc=$?'
Failed to start transient service unit: Connection reset by peer
Reload daemon failed: Transport endpoint is not connected
rc=1

$ ssh <host> 'sudo systemd-run --pipe --wait --collect true; echo rc=$?'
Failed to start transient service unit: Connection reset by peer
rc=1
```

The same commands over `ssh -tt` succeed (terminal fds are `user_devpts_t`, which policy allows).

### AVCs (not dontaudited; `semodule -DB` re-check showed no further denials related to this operation)

Captured 2026-07-16, fully updated Fedora 44 (selinux-policy-targeted-44.4-1.fc44, systemd-259.7-1.fc44, dbus-broker-37-8.fc44, kernel 7.1.3-201.fc44).

First reproducer command (`systemctl --machine`) — the write-direction inherited stderr pipe:

```
type=AVC msg=audit(1784191722.492:1221): avc:  denied  { write } for  pid=580 comm="dbus-broker" path="pipe:[24234]" dev="pipefs" ino=24234 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tclass=fifo_file permissive=0
```

Second reproducer command (`systemd-run --pipe`) — the read-direction stdin pipe:

```
type=AVC msg=audit(1784191722.720:1303): avc:  denied  { read } for  pid=580 comm="dbus-broker" path="pipe:[26714]" dev="pipefs" ino=26714 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tclass=fifo_file permissive=0
```

With a local module allowing only `system_dbusd_t`, the same reproducer advances to hop 2 and PID 1 itself is denied — which additionally **tears down PID 1's API-bus connection** ("Got disconnect on API bus"; host-wide `systemctl` outage — outage/recovery timing measured in the companion issue; `systemctl daemon-reexec` recovers it immediately):

```
type=AVC msg=audit(1784191727.777:1443): avc:  denied  { read } for  pid=1 comm="systemd" path="pipe:[27458]" dev="pipefs" ino=27458 scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tclass=fifo_file permissive=0
```

So the fix must cover both `system_dbusd_t` and `init_t` — and since the transient unit's payload domain can inherit the same pipes as its stdio, the companion PR restores the pre-split domain-wide grant rather than granting the two domains individually.

### Fix

`ssh_session_rw_pipes(domain)` next to the existing `ssh_rw_pipes(domain)` in domain.te — the same breadth as the `sshd_t` rule it mirrors, no `open` granted (inherited fds only). Verified on this box; the permission rationale and verification details are in the companion PR: #3288.

Verified still present at upstream rawhide HEAD (c905bb439, by inspection of the interface's callers in the policy source) and on the `f44` branch; since Fedora 44 stable is where this bites, a cherry-pick to `f44` after merge would be appreciated.

Related: #1106, systemd/systemd#34688, systemd/systemd#40089, [rhbz#2036829](https://bugzilla.redhat.com/show_bug.cgi?id=2036829). Companion issue for the same fd-stripping mechanism hitting regular-file redirect targets (`user_tmp_t`/`admin_home_t`): #3287.

BenjaminVouillaume:
> [!NOTE]
> **Disclaimer:** I am not an SELinux policy expert — apologies in advance if terminology or conventions are off. This investigation was carried out with the help of an AI assistant (Claude — Fable 5), and every reproducer, AVC, and fix in this PR and its companion issue was verified on a real Fedora 44 system (enforcing, re-checked with `semodule -DB`). I've done my best to make this contribution as valuable as possible — corrections are very welcome.

When OpenSSH split out the sshd-session binary and the policy introduced `sshd_session_t`, the long-standing domain-wide grant for inherited sshd pipes — `ssh_rw_pipes(domain)` in domain.te — was not mirrored for the new type: `ssh_session_rw_pipes()` is only called by `passwd_t`.

**Mechanism.** Stdio pipes inherited from a non-interactive SSH session (no TTY — the default transport of ansible, pyinfra and similar tools) can no longer be passed over D-Bus in `StartTransientUnit` messages: neither dbus-broker nor PID 1 may use `sshd_session_t:fifo_file`, so the kernel strips the `SCM_RIGHTS` fds and sets `MSG_CTRUNC`, which the receiving peer treats as fatal for the connection (sd-bus ≤ v259 / dbus-broker; systemd/systemd#34688, adjudicated as kernel/LSM behavior and closed at the merge of the v260 mitigation systemd/systemd#40089, which keeps the connection alive but cannot restore the stripped fds).

**Symptom.** Every `systemd-run --pipe` and `systemctl --user --machine=<user>@.host` over ssh without a TTY fails with "Connection reset by peer". Once dbus-broker alone is allowed, the same denial reappears inside PID 1 and additionally tears down PID 1's API-bus connection ("Got disconnect on API bus", host-wide `systemctl` outage until the manager's eventual bus reconnect or an immediate `systemctl daemon-reexec` — timing measured in the companion issues).

**Scope of the grant.** The grant is made domain-wide, exactly matching the `sshd_t` rule one line above. (Adjacent precedent: [rhbz#2036829](https://bugzilla.redhat.com/show_bug.cgi?id=2036829) / commit 95d703493 similarly allowed init_t to use inherited userdomain pipes for the shell-pipe sibling of this failure family.) This makes the existing `ssh_session_rw_pipes(passwd_t)` (`usermanage.te:425`) and `allow sshd_net_t sshd_session_t:fifo_file write;` (`ssh.te:94`) redundant — happy to drop the passwd_t call here or in a follow-up, maintainer's preference. Beyond the two measured receivers (`system_dbusd_t`, `init_t`), the transient unit's own payload domain can inherit the same pipes as its stdio, so a confined service doing stdio there would hit the same denial one hop further (not observed in our captures, where payloads ran unconfined). `ssh_session_rw_pipes()` expands to `rw_inherited_fifo_file_perms` = `{ getattr read write append ioctl lock }` — the `read`/`write` denials are the ones observed on the box; the remaining permissions are set parity with the `sshd_t` rule (with the ioctl/isatty follow-on documented for the userdomain-pipe case in commit 95d703493). No `open` is included: inherited fds only.

Raw AVCs are in the commit message (Fedora 44, selinux-policy-targeted-44.4-1.fc44, enforcing, captured 2026-07-16; not dontaudited — `semodule -DB` re-check showed no further denials related to this operation).

Verified on a fully-updated fedora/44 box: with the equivalent raw rules as a local module, both reproducers pass with zero AVCs (also under `semodule -DB`); removing the module restores the failure. `make policy && make validate` pass on the patched tree. Reproduced on aarch64; the policy change is arch-independent.

Fixes #3286
Related: #1106 (sibling failure in the same fd-strip family, open since 2022; its denied object predates the split and was never captured, so closing it is left to the maintainer after verifying)

Companion PR for the same fd-stripping mechanism hitting regular-file redirect targets: #3289.

### DIFF
--- a/policy/modules/kernel/domain.te
+++ b/policy/modules/kernel/domain.te
@@ -699,6 +699,7 @@ optional_policy(`
 
 optional_policy(`
 	ssh_rw_pipes(domain)
+	ssh_session_rw_pipes(domain)
 ')
 
 optional_policy(`


zpytela:
Merging, thank you.

PR #3291, opened 2026-07-20
Add missing rule for automotive policy

mwperina:
Existing rule is defined only for minimum, mls and targeted policies, so
we need to extend it also for automotive policy.

Resolves: RHEL-216823
Signed-off-by: Martin Perina <mperina@redhat.com>


mwperina:
@zpytela Could you please take a look?

zpytela:
Sorry for the delay, we just needed a ticket. The content is ok.

zpytela:
Merging, thank you.

PR #3300, opened 2026-07-23
rasdaemon: optionally permit access to ipmitool

jcpunk:
rasdaemon can run `ipmitool` to collect extra information. A boolean is added to optionally grant this extra access.

This is an attempt to port over https://github.com/SELinuxProject/refpolicy/pull/1184


zpytela:
The current policy does not confine ipmitool so reasonably more work would be required.

PR #3311, opened 2026-07-31
Dontaudit tlp_t dac_override (bsc#1272935)

ca-hu:
Fixes:
avc:  denied  { dac_override } for  pid=6910 comm="ps" capability=1  scontext=system_u:system_r:tlp_t:s0 tcontext=system_u:system_r:tlp_t:s0 tclass=capability permissive=0


zpytela:
@ca-hu Any idea why the capability request was raised? As well as for dac_read_search in commit a73e5ce0a1. I can't see any PATH record in neither of the bugs.

ca-hu:
its ps accessing /proc, which is not needed, see https://github.com/fedora-selinux/selinux-policy/pull/3070

zpytela:
i see, thanks
let's have tlp-pd confined then https://github.com/fedora-selinux/selinux-policy/pull/3323

PR #3327, opened 2026-08-05
Allow initrc_t and unconfined_service_t to transition to install_t

gursewak1997:
When systemd runs a wrapper service rather than an `install_exec_t` binary directly, the process ends up in `initrc_t` or `unconfined_service_t`. Add domtrans rules so executing `install_exec_t` binaries (bootc, ostree, rpm-ostree) from these domains transitions to install_t.

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -530,6 +530,7 @@ optional_policy(`
 
 optional_policy(`
 	anaconda_domtrans_install(init_t)
+	anaconda_domtrans_install(initrc_t)
 ')
 
 optional_policy(`
--- a/policy/modules/system/unconfined.te
+++ b/policy/modules/system/unconfined.te
@@ -48,5 +48,9 @@ optional_policy(`
 ')
 
 optional_policy(`
+	anaconda_domtrans_install(unconfined_service_t)
+')
+
+optional_policy(`
     gpg_manage_admin_home_content(unconfined_service_t)
 ')


cgwalters (commented):
This looks really simple and straightforward to me!

gursewak1997:
@zpytela ^

HastD:
Fixes #3349.

cgwalters:
Friendly re-ping on this one, we have (multiple) downstream customers that need to carry ugly workarounds without this

policy/modules/system/init.te
```diff
@@ -530,6 +530,7 @@ optional_policy(`
 
 optional_policy(`
 	anaconda_domtrans_install(init_t)
+	anaconda_domtrans_install(initrc_t)
```
zpytela: please move it into a separate optional block around L1435

policy/modules/system/unconfined.te
```diff
@@ -47,6 +47,10 @@ optional_policy(`
 	container_runtime_domtrans(unconfined_service_t)
 ')
 
+optional_policy(`
+	anaconda_domtrans_install(unconfined_service_t)
+')
```
zpytela: please move it up before chrony, interfaces seem not to be properly alphabetically sorted, but still

zpytela:
I am sorry for the delay. I have some room now so we can finish this issue.

Can you just explain why there is any wrapper in place?

HastD:
@zpytela For example, this will happen if you have a systemd service like this:
```
[Service]
ExecStart=/bin/bash /usr/libexec/foo/some-shell-script.sh
# or:
ExecStart=/usr/bin/python3 /usr/libexec/foo/some-python-script.py
```
And then the script it invokes will run in `initrc_t` or `unconfined_service_t`, which means if the script executes something in `install_exec_t` (like rpm-ostree or bootc), the intended domain transition will not occur and it will not run in `install_t`.

zpytela:
I see, but my question is why. What's the content of the scripts? Would labeling them with e.g. install_exec_t do a better job? Or is there a mixed content?

HastD:
There could potentially be mixed content, sure. The point is, given that both `unconfined_t` and `init_t` transition to `install_t` when executing `install_exec_t`, why shouldn't `initrc_t` and `unconfined_service_t` do the same? This seems like a simple omission that I assumed was unintentional, not an unusual request that would require careful justification or case-by-case workarounds.

zpytela:
unconfined_service_t should be considered more like a workaround or a temporary solution, the proper way usually is labeling the executable and creating new policy

initrc_t was developed for sysvinit scripts

so again: What's the content of the scripts and why there is any wrapper in place?

cgwalters:
> Would labeling them with e.g. install_exec_t do a better job? 

The problem is no one realizes they need to do this. It's such a subtle, non-obvious requirement.

What makes it truly, incredibly painful is that things will *only fail when doing an update that includes labels (xattrs) unknown to the current policy*. Seriously...it will work when you're doing "minor" updates and then just explode when you're going between RHEL minors, or randomly one day if you're tracking CentOS Stream or another rolling OS.

---

This problem is *so bad* that I have considered in the past changing bootc so that in the places where we call `setxattr(security.selinux)` we do that in wholly separate systemd service like `bootc-backend.service` but *that* has vast implications on the whole system architecture (which doesn't involve a daemon right now by default).

PR #3331, opened 2026-08-11
Allow unconfined_service_t nnp_transition to container_runtime_t

JJanasek:
Use the container_runtime_nnp_domtrans() interface added in container-selinux PR #473 [1] to allow podman running as a systemd service under unconfined_service_t with NoNewPrivileges=yes to transition to container_runtime_t.

Resolves: rhbz#2417297
[1] https://github.com/containers/container-selinux/pull/473


zpytela:
Merging, thank you.

PR #3339, opened 2026-08-12
CI: add MCS/deny and automotive policy build matrix

rhakkohli:
## Summary
Extend the GitHub Actions `build` workflow with a compile matrix so policy is validated under additional build options used by downstream variants, not only the default targeted configuration with `UNK_PERMS=allow`.
The existing single-job behavior is preserved as `targeted-mcs-allow` (including `container.pp`).
## Motivation
Today CI runs:
- `make policy`
- `make validate`
- `make container.pp`
against `dist/targeted/modules.conf` with default `build.conf` settings (`UNK_PERMS=allow`).
Some policy variants (including automotive on the `c10s` branch) are built with:
- `TYPE=mcs`
- `UNK_PERMS=deny`
- a filtered module set and/or `NAME=automotive`
Regressions in those option combinations may not show up in the default job until packaging or integration testing. Running them in CI catches compile/link/expand failures earlier.
## Changes
Update `.github/workflows/build.yml` to use a `strategy.matrix` with `fail-fast: false`:
| Job | Module set | Build options | Notes |
|-----|------------|---------------|-------|
| `targeted-mcs-allow` | `dist/targeted` | `TYPE=mcs`, `UNK_PERMS=allow` | Same as today + `container.pp` |
| `targeted-mcs-deny` | `dist/targeted` | `TYPE=mcs`, `UNK_PERMS=deny` | Full targeted modules |
| `minimum-mcs-deny` | `dist/minimum` | `TYPE=mcs`, `UNK_PERMS=deny`, `NAME=minimum` | Filtered module set |
| `automotive-mcs-deny` | `dist/automotive` | `TYPE=mcs`, `UNK_PERMS=deny`, `NAME=automotive` | Installs `booleans.conf` and `users` from dist |

Each job runs `make policy` and `make validate` (modular link + expand via existing Makefile rules).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,39 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Default upstream path (UNK_PERMS=allow from build.conf).
+          - name: targeted-mcs-allow
+            modules_conf: targeted
+            unk_perms: allow
+            policy_name: ""
+            install_dist_conf: false
+            build_container_pp: true
+          # MCS with deny unknown permissions on full targeted modules.
+          - name: targeted-mcs-deny
+            modules_conf: targeted
+            unk_perms: deny
+            policy_name: ""
+            install_dist_conf: false
+            build_container_pp: false
+          # Filtered module set with MCS + deny.
+          - name: minimum-mcs-deny
+            modules_conf: minimum
+            unk_perms: deny
+            policy_name: minimum
+            install_dist_conf: false
+            build_container_pp: false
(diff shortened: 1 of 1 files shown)


rhakkohli:
@vmojzis Please provide your feedback here.. thanks!

vmojzis:
Hi, I don't see a reason for the additional builds. The "deny_unknown" option does not influence the compilation of the policy modules in any way (it basically just sets a variable in the binary policy file), so "targeted-mcs-deny" and "minimum-mcs-deny" will always have the same compilation result as "targeted-mcs-allow" and "minimum-mcs-allow" respectively.
Given that the automotive package is a subset of modules from targeted, it is very unlikely that it will fail to compile when targeted succeeds. And in such a case it will be caught by distgit build. Running it here would just needlessly prolong the test runs.

PR #3340, opened 2026-08-12
rhsmcertd: allow bootc/ostree transient package persistence detection

jmarrero:
On image-mode (bootc/ostree) systems, subscription-manager's rhsm-package-profile-uploader needs to determine which packages are persistent vs transient.  This requires:

rpm.fc: Label /usr/share/rpm as rpm_var_lib_t instead of usr_t.
  On image-mode systems, /usr/lib/sysimage/rpm is a symlink to
  ../../share/rpm, so the rpmdb actually lives at /usr/share/rpm.
  Without the correct label, overlay WAL/SHM files inherit usr_t
  and rhsmcertd_t is denied write access.  On traditional
  package-mode systems this path does not exist, so the rule is
  harmless.

rhsmcertd.te:
  - fs_getattr_xattr_fs: allow statfs() on root filesystem, needed by libostree when opening the sysroot.
  - can_exec(install_exec_t): allow executing /usr/bin/rpm-ostree to query base commit packages via 'rpm-ostree status --json' and 'rpm-ostree db list'.
  - systemd_exec_systemctl: allow rpm-ostree to call systemctl (it runs 'systemctl is-active rpm-ostreed' internally).
  - files_read_generic_pids: allow reading ostree deployment state at /run/ostree/deployment-state/.../unlocked-transient.

Resolves: RHEL-152111

Assisted-by: AI

Draft while iterating...


policy/modules/contrib/rhsmcertd.te
```diff
@@ -174,6 +176,24 @@ optional_policy(`
...
+# rpm-ostree to determine which packages are persistent vs transient.
+# This requires executing /usr/bin/rpm-ostree (install_exec_t),
+# rpm-ostree calling systemctl (systemd_systemctl_exec_t), and reading
+# /run/ostree/deployment-state/.../unlocked-transient (var_run_t).
+optional_policy(`
+	gen_require(`
+		type install_exec_t;
+	')
+	can_exec(rhsmcertd_t, install_exec_t)
+')
```
zpytela, line 185: Types from other modules cannot be used directly but through interfaces. I'd suggest to use anaconda_domtrans_install(rhsmcertd_t) first, remove all other allow rules and check the current system state.

zpytela (changes requested):

jmarrero:
I let AI force push and messed this up... new PR on: https://github.com/fedora-selinux/selinux-policy/pull/3342

policy/modules/contrib/rhsmcertd.te
jmarrero: sorry new PR https://github.com/fedora-selinux/selinux-policy/pull/3342

PR #3342, opened 2026-08-14
rhsmcertd: allow bootc/ostree transient package persistence detection

jmarrero:
rhsmcertd: allow bootc/ostree transient package persistence detection
On image-mode (bootc/ostree) systems, subscription-manager's
rhsm-package-profile-uploader needs to determine which packages are
persistent vs transient.  This requires:

rpm.fc: Label /usr/share/rpm as rpm_var_lib_t instead of usr_t.
  On image-mode systems, /usr/lib/sysimage/rpm is a symlink to
  ../../share/rpm, so the rpmdb actually lives at /usr/share/rpm.
  Without the correct label, overlay WAL/SHM files inherit usr_t
  and rhsmcertd_t is denied write access.  On traditional
  package-mode systems this path does not exist, so the rule is
  harmless.

rhsmcertd.te:
  - anaconda_domtrans_install: transition to install_t when
    executing rpm-ostree (install_exec_t), so the child process
    has the permissions it needs (systemctl, reading deployment
    state, etc).
  - fs_getattr_xattr_fs: allow statfs() on root filesystem,
    needed by libostree when opening the sysroot.
  - manage_lnk_files_pattern + lnk_file: allow rhsmcertd to
    manage symlinks in tmp (rpm-ostree creates temporary rpmdb
    symlinks).

Resolves: RHEL-152111

Assisted-by: AI

### DIFF
--- a/policy/modules/contrib/rhsmcertd.te
+++ b/policy/modules/contrib/rhsmcertd.te
@@ -61,7 +61,8 @@ files_lock_filetrans(rhsmcertd_t, rhsmcertd_lock_t, file)
 
 manage_dirs_pattern(rhsmcertd_t, rhsmcertd_tmp_t, rhsmcertd_tmp_t)
 manage_files_pattern(rhsmcertd_t, rhsmcertd_tmp_t, rhsmcertd_tmp_t)
-files_tmp_filetrans(rhsmcertd_t, rhsmcertd_tmp_t, { dir file })
+manage_lnk_files_pattern(rhsmcertd_t, rhsmcertd_tmp_t, rhsmcertd_tmp_t)
+files_tmp_filetrans(rhsmcertd_t, rhsmcertd_tmp_t, { dir file lnk_file })
 
 manage_files_pattern(rhsmcertd_t, rhsmcertd_tmpfs_t, rhsmcertd_tmpfs_t)
 fs_tmpfs_filetrans(rhsmcertd_t, rhsmcertd_tmpfs_t, file)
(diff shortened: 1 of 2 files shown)


zpytela:
LGTM now, can you just update the commit message?

jmarrero:
Looks like this still gives us:
```
type=AVC msg=audit(1786724853.588:167): avc:  denied  { getattr } for  pid=1592 comm="rhsm-package-pr" path="/run/ostree/deployment-state/db7166b7d77945b3e0e31abfc0a93901facb26957b7f2aa0b6f2cf089b216814.0/unlocked-transient" dev="tmpfs" ino=1778 scontext=system_u:system_r:rhsmcertd_t:s0 tcontext=unconfined_u:object_r:var_run_t:s0 tclass=file permissive=0
```

jmarrero:
Pushed yet another change let see how testing goes.

policy/modules/contrib/rhsmcertd.te
```diff
@@ -174,6 +176,18 @@ optional_policy(`
...
+# executes rpm-ostree (install_exec_t) to determine which packages are
+# persistent vs transient.  Transition to install_t so the child process
+# has the permissions it needs (systemctl, reading deployment state, etc).
+optional_policy(`
+	anaconda_domtrans_install(rhsmcertd_t)
+')
+# rhsm-package-profile-uploader directly reads /run/ostree/deployment-state/
+# to detect transient overlays (this happens in the rhsmcertd_t domain itself,
+# not in the rpm-ostree child process).
+files_read_generic_pids(rhsmcertd_t)
```
zpytela: please move the line up to other files_...

zpytela:
Merging, thank you.

PR #3350, opened 2026-08-17
Allow unconfined_service_t to execute install_exec_t in install_t

### LINKED ISSUE #3349: unconfined_service_t missing type transition to install_t
Both `init_t` and `unconfined_t` have type transition rules so that executing a file of type `install_exec_t` results in a process running in `install_t` (e.g. for running rpm-ostree or bootc). However, `unconfined_service_t` has no such rule. This results in programs like bootc running in the wrong domain (and therefore failing in various ways) when run from `unconfined_service_t`. This can happen, for example, if it's launched *indirectly* in a systemd service, that is, not as the main `ExecStart` but as a step in a script launched by a systemd service.

HastD:
Fixes https://github.com/fedora-selinux/selinux-policy/issues/3349

### DIFF
--- a/policy/modules/system/unconfined.te
+++ b/policy/modules/system/unconfined.te
@@ -24,6 +24,10 @@ init_use_notify(unconfined_service_t)
 usermanage_run_passwd(unconfined_service_t, system_r)
 
 optional_policy(`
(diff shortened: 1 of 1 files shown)


HastD:
Accidental duplicate of #3327.

PR #3397, opened 2026-09-05
Adjustments for systemd 262

keszybz:
Three patches so silence new AVCs reported when booting Fedora rawhide with systemd-262~rc1.

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -138,6 +138,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/systemd/.+\.conf\.d/.+\.conf	--	gen_context(system_u:object_r:systemd_conf_t,s0)
 /run/systemd/.+\.conf\.d/.+\.conf	-l	gen_context(system_u:object_r:systemd_conf_t,s0)
 /run/.*nologin.*		gen_context(system_u:object_r:systemd_logind_var_run_t,s0)
+/run/systemd/coredumpd(/.*)?	gen_context(system_u:object_r:systemd_coredump_var_run_t,s0)
 /run/systemd/default-hostname	--	gen_context(system_u:object_r:hostname_etc_t,s0)
 
 /run/systemd/generator(/.*)?	gen_context(system_u:object_r:systemd_generator_unit_file_t,s0)
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1329,6 +1329,8 @@ kernel_read_security_state(systemd_sysctl_t)
 kernel_write_security_state(systemd_sysctl_t)
 
 files_read_system_conf_files(systemd_sysctl_t)
+# create /run/sysctl.d and config files in it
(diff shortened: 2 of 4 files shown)


policy/modules/system/systemd.fc
```diff
@@ -138,6 +138,7 @@ HOME_DIR/\.config/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_unit
 /run/systemd/.+\.conf\.d/.+\.conf	--	gen_context(system_u:object_r:systemd_conf_t,s0)
 /run/systemd/.+\.conf\.d/.+\.conf	-l	gen_context(system_u:object_r:systemd_conf_t,s0)
 /run/.*nologin.*		gen_context(system_u:object_r:systemd_logind_var_run_t,s0)
+/run/systemd/coredumpd(/.*)?	-s	gen_context(system_u:object_r:systemd_coredump_var_run_t,s0)
```
zpytela: I am okay with the current state, just wonder why not label also the parent dir?

/run/systemd/coredumpd(/.*)?		gen_context(system_u:object_r:systemd_coredump_var_run_t,s0)

and I guess the /run/systemd/coredump socket in lower releases also would use the new type

zpytela (commented):
I'd like to create F45 and F46 builds today.

policy/modules/system/systemd.fc
keszybz: Yeah, the `-s` was a leftover from an earlier version. I edited the filepath pattern but forgot to remove that.

policy/modules/system/systemd.te
```diff
@@ -1367,6 +1370,10 @@ manage_files_pattern(systemd_coredump_t, systemd_coredump_var_lib_t, systemd_cor
 mmap_files_pattern(systemd_coredump_t, systemd_coredump_var_lib_t, systemd_coredump_var_lib_t)
 init_var_lib_filetrans(systemd_coredump_t, systemd_coredump_var_lib_t, dir, "coredump")
 
+# /run/systemd/coredumpd/kernel socket
+allow systemd_coredump_t systemd_coredump_var_run_t:sock_file manage_sock_file_perms;
+init_named_pid_filetrans(systemd_coredump_t, systemd_coredump_var_run_t, sock_file, "kernel")
```
zpytela: this is now either not needed if systemd labels the dir itself, or should be

init_named_pid_filetrans(systemd_coredump_t, systemd_coredump_var_run_t, dir, "coredumpd")

zpytela:
The build fails because the latest package version uses a quick patch with the same content as it is now in the policy.

+ /usr/bin/rpmuncompress /builddir/build/SOURCES/0001-Support-console-version-of-initial-setup.patch

policy/modules/system/systemd.te
keszybz: The directory is (if I'm interpreting all the tea leaves correctly ;) ) by PID1 via `RuntimeDirectory=systemd/coredumpd`. So I think it should be labelled automatically. I'll drop this line.

zpytela:
Merging, thank you.

PR #3399, opened 2026-09-08
Allow rsync to read var_t (bsc#1279565)

jsegitz:
Currently rsyncd fails to start unless rsync_export_all_ro is set

Solves:
avc:  denied  { read } for  comm="rsync" name="var" dev="vda2" scontext=system_u:system_r:rsync_t:s0
tcontext=system_u:object_r:var_t:s0
tclass=dir

### DIFF
--- a/policy/modules/contrib/rsync.te
+++ b/policy/modules/contrib/rsync.te
@@ -122,6 +122,7 @@ dev_read_urand(rsync_t)
 fs_getattr_xattr_fs(rsync_t)
 fs_search_auto_mountpoints(rsync_t)
 
+files_list_var(rsync_t)
 files_search_home(rsync_t)
 
 auth_use_nsswitch(rsync_t)


zpytela:
Why does it want to read /var, is this a new behaviour? Is there any change in the system setup?

jsegitz:
I don't know if it's new, but it happens on Fedora and openSUSE when running rsyncd. It tries to create a lock file 
`Sep 08 09:06:20 localhost.localdomain rsyncd[1437]: rsync: [Receiver] failed to open log-file /var/log/rsyncd.log: Permission denied (13)`

zpytela:
It does not in my Fedora, therefore asking. Moreover reading /var is not needed and never was. So any configuration change, chain of filesystems, symlinks? Can you e.g. strace it?

```
# rpm -q rsync-daemon
rsync-daemon-3.5.0-1.fc46.noarch
# getsebool rsync_export_all_ro 
rsync_export_all_ro --> off
```

jsegitz:
yes, you're correct. I apparently had an existing non-default rsyncd config on this machine (which is one of my template machines, so not a good idea).

log file = /var/log/rsyncd.log
This causes this AVC to happen. I think this is a reasonable config and the permission should be granted

zpytela:
Seems to be a feature of rsync-3.5.0:
```
openat(AT_FDCWD</root>, "/", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = 3</>
...
openat(3</>, "var", O_RDONLY|O_NOFOLLOW|O_CLOEXEC|O_DIRECTORY) = 4</var>
...
openat(4</var>, "log", O_RDONLY|O_NOFOLLOW|O_CLOEXEC|O_DIRECTORY) = 3</var/log> 
...                            
openat(3</var/log>, "rsync.log", O_WRONLY|O_CREAT|O_APPEND|O_NOFOLLOW, 0644) = 4</var/log/rsync.log>
...
write(4</var/log/rsync.log>, "2026/09/09 11:10:03 [1975] rsyncd version 3.5.0-g41cb6b71 starting, listening on port 873\n", 90) = 90
```
while in rsync-3.4.4 there is
```
openat(AT_FDCWD</>, "/var/log/rsync.log", O_WRONLY|O_CREAT|O_APPEND, 0666) = 3</var/log/rsync.log>
...
write(3</var/log/rsync.log>, "2026/09/09 11:15:53 [13122] rsyncd version 3.4.4 starting, listening on port 873\n", 81) = 81
```

zpytela:
Probably brought by 3dddbde5ffd, rsync now uses open_no_attacker_symlinks() in logfile_open() instead of fopen() and walks through the path to prevent from planted symlinks attack. Can you add something like this to the commit message?

jsegitz:
changed the commit message, thanks for the analysis

zpytela:
Merging, thank you.

PR #3400, opened 2026-09-08
Allow rsync to getattr pipes and sockes if rsync_export_all_ro is set

jsegitz:
Introduces two new interfaces
- files_getattr_non_auth_sockets
- files_getattr_non_auth_pipes interface
and uses it in the rsync_export_all_ro boolean to allow access


zpytela:
I wonder what is the use case for rsync to handle these files. Would dontaudit effect in the same state on the target system?

jsegitz:
I don't think so, because e.g. permissions on a FIFO would not get synced

zpytela:
Merging, thank you.

